PR #2, opened 2018-04-30
Code review

philippjfr:
DO NOT MERGE (actually it doesn't make any difference)


holoplot/__init__.py
```diff
@@ -0,0 +1,385 @@
+import holoviews
```
jbednar: I strongly prefer all .py files to have a module docstring that explains what this module is about, but I see that even holoviews doesn't have one for its `__init__.py`, so you must disagree! :-)

holoplot/__init__.py
```diff
@@ -0,0 +1,385 @@
...
+from holoviews import (
+    extension as _extension, Dimensioned as _Dimensioned, HoloMap as _HoloMap
+)
+
```
jbednar: I also don't much like parentheses in import statements, but again you must disagree!

holoplot/__init__.py
```diff
@@ -0,0 +1,385 @@
...
+        DataSource.plot = property(_patch_plot)
+    if 'xarray' in library:
+        try:
+            from xarray import DataArray
+        except ImportError:
+            raise ImportError('Could not patch plotting API onto xarray. '
+                              'Xarray could not be imported.')
+        DataArray.plot = property(_patch_plot)
+    if extension:
+        _extension(extension)
```
jbednar: Does this mean that no extension will be loaded when doing `from holoplot import xarray`?  I can understand why one wouldn't, to let the extension be selectable, but I strongly feel like an extension should be loaded so that the import is all that's required before using .plot.  Just getting some plot out of this system should be as easy as we can possibly make it, right?

Maybe `holoplot.xarray` could have something like `patch('xarray','auto')`, where `auto` tries loading the extensions according to a fixed importing order (`bokeh`,`matplotlib`, presumably?)?

holoplot/__init__.py
```diff
@@ -0,0 +1,385 @@
...
+class HoloPlot(object):
+
+    def __init__(self, data):
+        self._data = data
+
+    def __call__(self, x=None, y=None, kind=None, backlog=1000,
+             width=700, height=300, title=None, grid=False,
+             legend=True, logx=False, logy=False, loglog=False,
+             xticks=None, yticks=None, xlim=None, ylim=None, rot=None,
+             fontsize=None, colormap=None, hover=True, **kwds):
```
jbednar: Is there a specific reason these are all explicit keyword arguments, instead of Parameters or part of `kwds` so that they can be checked against a lookup table?  Is it for tab-completion?

holoplot/__init__.py
```diff
@@ -0,0 +1,385 @@
...
+        Parameters
+        ----------
+        x, y : label or position, optional
+            Coordinates for each point.
+        **kwds : optional
+            Keyword arguments to pass on to
+            :py:meth:`intake.source.base.DataSource.plot`.
+        Returns
+        -------
+        Element : Element or NdOverlay of Elements
```
jbednar: Given that holoplot is for returning an HV object from something that is not an HV object, should it convey that the underlying library is switching by saying `hv.Element`, etc. here?  Otherwise people may have no idea where to look for Element.

setup.py
```diff
@@ -0,0 +1,44 @@
...
+        "License :: OSI Approved :: BSD License",
+        "Development Status :: 5 - Production/Stable",
+        "Programming Language :: Python :: 2",
+        "Programming Language :: Python :: 2.6",
+        "Programming Language :: Python :: 2.7",
+        "Programming Language :: Python :: 3.2",
+        "Programming Language :: Python :: 3.3",
+        "Programming Language :: Python :: 3.4",
+        "Programming Language :: Python :: 3.5",
+        "Programming Language :: Python :: 3.6",
```
jbednar: Not sure if you've tried anything on py37, but it's apparently about a month from release, so it might be worth listing here.

environment.yml
```diff
@@ -0,0 +1,51 @@
+name: holoplot
```
jbednar: It might be good to clarify what this environment is for; is it just for running through the examples or for building the docs?

conda.recipe/meta.yaml
```diff
@@ -0,0 +1,76 @@
...
+    - python-snappy     # Required for reading snappy-compressed Parquet files
+    - pytest >=3.2      # Required by? Pinned to avoid installation problems on Windows with 3.1?
+    - cffi              # Unstated dependency of fastparquet?
+    - tornado 4.5.3     # Required for Jupyter notebook and bokeh server
+    # Intake data packages used in documentation #
+    - airline_flights
+    - us_crime
+    - intake
+    - intake-parquet
+
```
jbednar: Why are xarray, intake, dask, and pandas in the run requirements (i.e., dependencies)?  I thought the point of this package was that it could be installed alongside any of those packages, without requiring any of the others.  Shouldn't those be in the test requirements instead?

doc/getting_started/index.rst
```diff
@@ -0,0 +1,55 @@
...
+Installation
+------------
+
+HoloPlot supports Python 2.7-3.7 on Linux, Windows, or Mac.
+
+Installable ``holoplot`` packages will be created at some point, but
+at present, the recommended way to install HoloPlot is based on conda
+`conda <http://conda.pydata.org/docs>`_ environments and 
+`git <https://git-scm.com>`_:
+
```
jbednar: Presumably this needs updating as soon as you release a package.

doc/getting_started/index.rst
```diff
@@ -0,0 +1,55 @@
...
+run the examples shown on the website using
+`Jupyter <http://jupyter.org>`_::
+
+    cd examples
+    jupyter notebook --NotebookApp.iopub_data_rate_limit=1e8
+
+(Increasing the rate limit in this way is `required for the 5.0 Jupyter version
+<http://holoviews.org/user_guide/Installing_and_Configuring.html>`_,
+but should not be needed in earlier or later Jupyter releases.)
+
```
jbednar: Should the package pin to jupyter>5.0, at least as an optional dependency, to avoid having to say this?

doc/getting_started/index.rst
```diff
@@ -0,0 +1,55 @@
...
+To work with JupyterLab you will also need the HoloViews JupyterLab
+extension::
+
+    conda install -c conda-forge jupyterlab
+    jupyter labextension install @pyviz/jupyterlab_holoviews
+
+Once you have installed JupyterLab and the extension launch it with::
+
+    jupyter-lab
```
jbednar: Isn't that meant to be 'jupyter lab'?  Or do both work?

doc/user_guide/Streaming_Plots.rst
```diff
@@ -0,0 +1,6 @@
+Streaming Plots
```
jbednar: Not sure why latest_news.html is an empty file?

doc/user_guide/Streaming_Plots.rst
```diff
@@ -0,0 +1,6 @@
...
+_______________
+
+.. notebook:: holoplot ../../examples/user_guide/Streaming_Plots.ipynb
+    :skip_execute: True
+
```
jbednar: Is there some reason to commit .rst files like this to the repo?  Shouldn't they just be auto-generated?

holoplot/__init__.py
jbednar: (With 'bokeh' first in the list because I think most people will be interested in HoloPlot because of its ability to return an interactive plot, though of course matplotlib-based HoloViews plots also have additional capabilities beyond what the native .plot() interface will support.)

holoplot/__init__.py
philippjfr: I don't disagree, HoloViews is horribly underdocumented and I don't want to repeat that here.

holoplot/__init__.py
philippjfr: I do disagree, in this case it doesn't seem all that necessary but as soon as it doesn't fit on one line I prefer it.

holoplot/__init__.py
philippjfr: >Does this mean that no extension will be loaded when doing from holoplot import xarray? I can understand why one wouldn't, to let the extension be selectable, but I strongly feel like an extension should be loaded so that the import is all that's required before using .plot. Just getting some plot out of this system should be as easy as we can possibly make it, right?

Originally I did make it load the extension automatically, however loading the extension interferes with usage outside the notebook, because it's what holoplot uses to tell whether you're in a notebook or not. In other words the functionality described and shown in the Commandline & Scripts section [here](https://pyviz.github.io/holoplot/user_guide/Display_and_Saving.html) wouldn't work.

doc/getting_started/index.rst
philippjfr: Yes, I'll update it shortly.

setup.py
philippjfr: True.

environment.yml
philippjfr: Building docs and testing.

conda.recipe/meta.yaml
philippjfr: Agreed, I'll have to update this. I started by copying the pyviz meta.yaml but need to pare it down now.

doc/getting_started/index.rst
philippjfr: Yes, I think so.

doc/getting_started/index.rst
philippjfr: Both work.

doc/user_guide/Streaming_Plots.rst
philippjfr: Yes, it's the reason I filed https://github.com/pyviz/nbsite/issues/43, this notebook should not be run and therefore this file declares ``:skip_execute: True``.

holoplot/__init__.py
philippjfr: I'll have to update these, they aren't really accurate anymore, it could be any one of: Element/NdOverlay/HoloMap/DynamicMap/NdLayout/GridSpace

holoplot/__init__.py
jbednar: Ah.  I prefer just having multiple imports in that case, with each fitting onto one line, because it's visually easier for me to scan and dismiss a large block of imports that way; when there are parentheses it starts looking like actual code. But as you wish...

holoplot/__init__.py
jbednar: > Worth noting that I have made no effort at all to support matplotlib yet. It's likely totally broken because options like width and height are always set.

The options used in holoplot make a good candidate set of things that we would like to work with the same syntax across backends, so that would be something to try to do separately from holoplot (respect those plot options in matplotlib). I do realize that the sizing options are the most difficult to respect given the libraries' different sizing models, but they are also the most important...

> Originally I did make it load the extension automatically, however loading the extension interferes with usage outside the notebook, because it's what holoplot uses to tell whether you're in a notebook or not.

I don't get it; I thought we've for many months been able to load hv.extension() in or outside of a notebook, and have it do the right thing in both cases.  I think that's crucial for having a smooth user experience; any single case where we have to tell people to do something different in different circumstances cuts down our audience dramatically.  Can't we have hv.extension() leave some marker that indicates whether it's in a notebook or not?  This is a very important issue; the instructions really have to look simple and clean for this interface to be compelling to its intended audience.

holoplot/__init__.py
jbednar: Maybe just "Returns _ HoloViews Object (any one of ...)", then.

conda.recipe/meta.yaml
jbednar: Ok, good.

holoplot/converter.py
```diff
@@ -0,0 +1,652 @@
...
+try:
+    from intake.source.base import DataSource
+except ImportError:
+    DataSource = None
+
+try:
+    import dask.dataframe as dd
+except ImportError:
+    dd = None
+
```
jbednar: I'm alarmed by seeing all these conditional imports here, because it means that if I have any of these packages in my environment, anything I use with holoplot now depends on them to be working without errors.  Plus this approach will slow down everyone on every usage, and more so when more libraries are added.

Given that this is a package very likely to be used with only one of the packages from this list by any given user at any given time, shouldn't the imports be postponed until later and only either when the functionality is actually used or else swallowing all exceptions, not just ImportError for any library not used?

holoplot/converter.py
```diff
@@ -0,0 +1,652 @@
...
+                 xlim=None, ylim=None, xticks=None, yticks=None,
+                 fontsize=None, colormap=None, stacked=False,
+                 logx=False, logy=False, loglog=False, hover=True,
+                 style_opts={}, plot_opts={}, use_index=True,
+                 value_label='value', group_label='Group',
+                 colorbar=False, streaming=False, backlog=1000,
+                 timeout=1000, persist=False, use_dask=False,
+                 datashade=False, subplots=False, label=None,
+                 groupby=None, dynamic=True, index=None, show=False,
+                 crs=None, fields={}, **kwds):
```
jbednar: It would be nice if this list of supported options could be shared with the main `__init__` file; otherwise seems like it will be unlikely to stay matching over time.

holoplot/converter.py
philippjfr: Yes, that's one of my to-do items but atm it's somewhat pointless because HoloViews will import all these things anyway (except for streamz maybe). The solution I come up with here can then presumably should also be applied to HoloViews ideally.

jbednar: Ok, good luck!

jbednar: If you need any extra motivation, this issue was theoretical when I wrote the above, but now that I actually tried it, it's immediately a problem:

image

I.e. I presumably have an older streamz library installed, and even though this example doesn't use streamz at all, it's failing due to streamz anyway.

holoplot/__init__.py
philippjfr: >I don't get it; I thought we've for many months been able to load hv.extension() in or outside of a notebook, and have it do the right thing in both cases

That's because we don't offer any function that aims to display an object in or outside the notebook. For ease of use I've introduced a ``holoplot.show`` function that will display a plot in the notebook if the extension is enabled, or open a browser window with a static or server backed plot otherwise. I'd be okay making ``show`` always open a new window by default though.

How about making the imports load the extension and leaving patch as it is? Most users will just need the import.

jbednar: If you'll be making a PR for that, I'll change the notebooks to describe this way of doing things, as I already have extensive (but local and tiny) edits to them.

examples/user_guide/Plotting_API.ipynb
```diff
@@ -0,0 +1,643 @@
...
+   "source": [
+    "flights.plot(x='distance', y='depdelay', by='carrier', kind='scatter', alpha=0.2).select(carrier=[b'OH', b'F9'])"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This approach differs slightly from the usual pandas plotting API, which usually assumes that the dataframe has a meaningful index and the columns represent similar observations. We can reproduce this functionality by enabling the ``use_index`` option and declaring the columns to plot (by default it will use all columns). The ``use_index`` can either be set to ``True`` in which case it will use the actual index or reference a column in which case it will be treated like the index. In this approach the y-axis will receive a default label named 'value' however we can already override the label using the ``value_label`` option.\n",
+    "\n",
```
jbednar: Shouldn't use_index be True by default, so that people who use Pandas can use Pandas's mechanisms to declare which dimensions are the index(es) (key dimensions)?

philippjfr: I think it is True now actually.

jbednar:
It looks like all the review comments have been dealt with, so this PR can now be closed.

PR #5, opened 2018-05-02
Cleaned up notebook stories and examples

jbednar:
Did a complete pass over the notebooks:

* Simplified code for loading extension and patching .plot
* Added pure-Pandas initial examples
* Split out homepage into a series of self-contained examples
* Stripped notebook metadata
* Moved most contents of doc/index.rst into homepage.ipynb
* Simplified notebook names

Remaining for Philipp to do, as separate PRs:
* Remove logo for call to extension, by default.
* The Streams notebook should presumably be cut down significantly, just to focus on the parts specific to streamz
* The remaining images in the Streams notebook should then presumably all be animated.
* Should intake have its own page?
* Should enable colorbar=True automatically when appropriate


philippjfr:
Going to go ahead an merge, since only a small thing errored which I've now fixed.

PR #10, opened 2018-05-08
Packaging

ceball:


ceball:
@philippjfr, could you take your best guess at the dependencies and enter them into setup.py in this PR (or as a comment here)? Doesn't matter for now if they are not quite right/complete. 

I.e. `install_requires` and then whatever categories you think are necessary in `extras_require`. I think GeoViews ended up with `recommended` (I think lets you run most of the examples), and `examples_extra` (awkward dependencies but required to run all the examples), plus 'standard' ones (doc, tests, build).

philippjfr:
Aren't you meant to be on holiday? 😄 Anyway thanks for this, I'll work on adding the dependencies shortly.

philippjfr:
Thanks for getting this started, all seems to be working now. Still needs some polish but that can happen in a later PR.

PR #11, opened 2018-05-19
Add improved geographic support

philippjfr:
- [x] Adds support for parsing CRS as EPSG code or proj.4 string
- [x] Adds support for plotting Points
- [ ] Adds documentation section

### DIFF
--- a/holoplot/converter.py
+++ b/holoplot/converter.py
@@ -115,7 +115,7 @@ class HoloViewsConverter(param.Parameterized):
                        'contours']
 
     def __init__(self, data, x, y, kind=None, by=None, use_index=True,
-                 group_label='Group', value_label='value',
+                 group_label='Variable', value_label='value',
                  backlog=1000, persist=False, use_dask=False,
                  crs=None, fields={}, groupby=None, dynamic=True,
                  width=700, height=300, shared_axes=True,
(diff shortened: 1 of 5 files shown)


holoplot/converter.py
```diff
@@ -546,6 +455,14 @@ def __call__(self, kind, x, y):
             eltype = 'Image'
             if 'cmap' in self._style_opts:
                 style['cmap'] = self._style_opts['cmap']
+
+        if self.crs:
+            # Apply projection before rasterizing
+            import cartopy.crs as ccrs
+            from geoviews import project
+            projection = self._plot_opts.get('projection', ccrs.GOOGLE_MERCATOR)
+            obj = project(obj, projection=projection)
```
jbednar: Web Mercator is the default output projection, not Plate Carree?

jbednar: I guess that way the projection won't change if someone overlays it with a tile source, and I suppose it's arbitrary anyway....

philippjfr: >Web Mercator is the default output projection, not Plate Carree?

For bokeh yes, for matplotlib no, but holoplot is focused on bokeh output for now.

>I guess that way the projection won't change if someone overlays it with a tile source

That's right, we picked GOOGLE_MERCATOR as the default for bokeh as that was the only way tile sources would work, and now we're somewhat restricted by backward compatibility.

philippjfr:
Will go ahead and merge. Documentation can come later, when there are new releases for holoviews and geoviews to fix some issues.

PR #80, opened 2018-10-04
Attempt to switch tests to pytest

philippjfr:


philippjfr:
Would require work in HoloViews first so closing for now.

PR #85, opened 2018-10-05
Add crs from ds if possible

### LINKED ISSUE #82: `geo=True` should look for `crs` on `xarray.dataset` if none provided
This came out of a conversation with @philippjfr. Potentially `crs` could also accept a string which would refer to an `attr` or I guess a `coord` containing the `crs` info. This would be a little trickier to implement since `crs` already accepts a proj4 string. But I think it'd be nicely parallel to how x, y, and z are set.

jsignell:
Closes #82 

I'm not sure where to document this behavior but the goal was to only have defaults be a little more helpful so it should just do what is expected with whatever is given be it a proj4string, a cartopy.crs or a ds.attr name.

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -145,7 +145,7 @@ class HoloViewsConverter(param.Parameterized):
         self.group_label = group_label
         self.dynamic = dynamic
         self.geo = geo or crs or global_extent or projection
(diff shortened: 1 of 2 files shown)


jsignell:
**NOTE** that the new tests don't run because test env does not have needed packages.

PR #94, opened 2018-10-18
Only check for coordinates that have dimensions in common with the value array

ahuang11:
https://github.com/pyviz/hvplot/issues/74

```
import xarray as xr
import hvplot.xarray
url = 'http://thredds.ucar.edu/thredds/dodsC/grib/FNMOC/WW3/Global_1p0deg/Best'
ds = xr.open_dataset(url)

ds['sig_wav_ht_surface'].hvplot(groupby='time')
```

image


ahuang11:
Reopened clean commit branch: https://github.com/pyviz/hvplot/pull/97

Issue #74, opened 2018-09-06
Improve 2D checking to generate best default plot with xarray


rsignell-usgs:
I've successfully used hvplot to make super-quick maps of some global wave model data with time slider as described in the [gridded data user guide](https://hvplot.pyviz.org/user_guide/Gridded_Data.html):
![2018-09-05_23-06-46](https://user-images.githubusercontent.com/1872600/45132850-84ee8100-b160-11e8-9822-56029c66527b.png)

But for other global wave model data, I still get the histogram using the same approach:
![2018-09-05_23-11-56](https://user-images.githubusercontent.com/1872600/45133005-24ac0f00-b161-11e8-99c2-54534038068e.png)

The notebook is here:
https://nbviewer.jupyter.org/gist/rsignell-usgs/3ff85b150bfd118f2b2ed768f98116a1

Any idea why and how I can work around this?

ahuang11:
I think if you manually input lon and lat like this: 

`ds['sig_wav_hg_surface'].hvplot('lon', 'lat', groupby='time2')`

it should work

rsignell-usgs:
@ahuang11 , yes!  That gives me what I was looking for:
![2018-09-06_6-26-28](https://user-images.githubusercontent.com/1872600/45152551-52b04400-b19f-11e8-9983-54613a2de19e.png)

Do you know why `hvplot` didn't do this by default in this case?

philippjfr:
I suspect it got thrown off by the ``reftime`` coordinate and the logic to check whether data is 2D is faulty and missing check that the coordinate is actually indexing anything.

rsignell-usgs:
@philippjfr, yep, seems like that's it, because this works:
```
ds = xr.open_dataset(url, drop_variables=['reftime2'])
ds['sig_wav_ht_surface'].hvplot(groupby='time2', clim=(0,5))
```

philippjfr:
I'll keep this open because we should only be checking for actual coordinates, i.e. coordinates that have dimensions in common with the value array.

PR #97, opened 2018-10-19
Only check for core dims

ahuang11:
#74 
```
import xarray as xr
import hvplot.xarray
url = 'http://thredds.ucar.edu/thredds/dodsC/grib/FNMOC/WW3/Global_1p0deg/Best'
ds = xr.open_dataset(url)

ds['sig_wav_ht_surface'].hvplot(groupby='time')
```
image


philippjfr:
This looks like it would be problematic for quadmeshes with curvilinear coordinates.

philippjfr:
This was superseded by #246.

PR #110, opened 2018-11-09
Updating to use nbsite 0.5.0

jsignell:
After 0.5.0 of nbsite, evaluated notebooks live in `/doc` next to their dynamically generated `.rst` files. Lines are added to `.gitignore` to prevent accidental checking in of these files. Note that it is sometimes desirable to check in `doc/*.rst` files but it is preferable to generate them from notebooks. The workflow for re-evaluating a notebook and regenerating it's html is:

```bash
$ rm builtdocs/user_guide/Plotting.html
$ rm doc/user_guide/Plotting.ipynb
$ nbsite build --what=html --output=builtdocs
```

Upload the entire contents of `./builtdocs` to your server, or `gh-pages` branch to refresh your website.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,3 +99,12 @@ ENV/
 
 # mypy
 .mypy_cache/
+
+# nbsite
+# these files normally shouldn't be checked in as they should be
+# dynamically built from notebooks
+doc/**/*.rst
+doc/**/*.ipynb
+doc/**/*.json
+# this dir contains the whole website and should not be checked in on master
+builtdocs/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 # We deliberately don't use travis's language=python option because
-# we install miniconda and use conda to get python. Additionally, 
+# we install miniconda and use conda to get python. Additionally,
 # Travis's auto-install of python doesn't work on osx images (see
 # https://github.com/travis-ci/travis-ci/issues/4729).
 language: generic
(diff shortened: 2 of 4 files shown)


philippjfr:
Any changes needed for the automatic website builds on travis?

jsignell:
I'm not exactly sure what you mean.

philippjfr:
I'll merge once tests pass.

jsignell:
Ok I tagged nbsite 0.5.1 it is still building, but should be ready in ~30min. So you can restart the travis builds when that is ready.

philippjfr:
After a few fixups this worked great, just to record what I had to change in case it applies to other projects:

* Latest nbsite seems to skip a notebook if it is in the same folder as the .rst. This means `doc/index.rst` includes a notebook it will not be built. I fixed this by moving the homepage.ipynb in hvplot to examples/homepage.ipynb
* When building holoviews notebooks into a website the ``HV_DOC_HTML=true`` environment variable should be set, this means that holoviews plots are exported as raw HTML, which is much more efficient than the default which is geared towards notebook usage.
* hvPlot notebooks don't generally have a title so I disabled the ``--offset``

philippjfr:
There's still an issue, it appears json files generated by HoloMaps in the notebooks are not copied, which means the widgets don't update the plot, e.g. see the xarray plot here: https://hvplot.pyviz.org/

jbednar:
I'm not sure what the specific problem is. E.g. hvplot with dynamic=False works fine here: http://pyviz.org/tutorial/01_Workflow_Introduction.html

philippjfr:
I guess it's only copying json files in certain directories.

PR #115, opened 2018-11-24
Fix image hover for additional cols

ahuang11:
image
https://github.com/ioam/holoviews/pull/3193


philippjfr:
Sorry for not merging this immediately, this change has a number of implications so I'm going to have to check this out in more detail.

PR #126, opened 2018-12-07
Fixing up some dimension handling in hvplot.util

### LINKED ISSUE #125: Getting an error on to_dataframe from a basic xarray.hvplot
xarray on latest master: 0.11.0+11.g3ae93ac

```python
import xarray as xr
import hvplot.xarray

ds.hvplot()
```
```python-traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-15-95a80b7187c4> in <module>()
----> 1 ds.hvplot()

/Users/jsignell/hvplot/hvplot/__init__.py in __call__(self, x, y, kind, **kwds)
     98         HoloViews object: Object representing the requested visualization
     99         """
--> 100         return self._get_converter(x, y, kind, **kwds)(kind, x, y)
    101 
    102     def _get_converter(self, x=None, y=None, kind=None, **kwds):

/Users/jsignell/hvplot/hvplot/__init__.py in _get_converter(self, x, y, kind, **kwds)
    106         kind = kind or params.pop('kind', None)
    107         return HoloViewsConverter(
--> 108             self._data, x, y, kind=kind, **params
    109         )
    110 

/Users/jsignell/hvplot/hvplot/converter.py in __init__(self, data, x, y, kind, by, use_index, group_label, value_label, backlog, persist, use_dask, crs, fields, groupby, dynamic, width, height, shared_axes, grid, legend, rot, title, xlim, ylim, clim, xticks, yticks, logx, logy, loglog, hover, subplots, label, invert, stacked, colorbar, fontsize, colormap, datashade, rasterize, row, col, figsize, debug, xaxis, yaxis, framewise, aggregator, projection, global_extent, geo, precompute, flip_xaxis, flip_yaxis, dynspread, hover_cols, x_sampling, y_sampling, project, xlabel, ylabel, xformatter, yformatter, tools, **kwds)
    150         self._process_data(kind, data, x, y, by, groupby, row, col,
    151                            use_dask, persist, backlog, label, value_label,
--> 152                            hover_cols, kwds)
    153         self.use_index = use_index
    154         self.value_label = value_label

/Users/jsignell/hvplot/hvplot/converter.py in _process_data(self, kind, data, x, y, by, groupby, row, col, use_dask, persist, backlog, label, value_label, hover_cols, kwds)
    372             data, x, y, by_new, groupby_new = process_xarray(data, x, y, by, groupby,
    373                                                              use_dask, persist, gridded,
--> 374                                                              label, value_label)
    375 
    376             if kind not in self._stats_types:

/Users/jsignell/hvplot/hvplot/util.py in process_xarray(data, x, y, by, groupby, use_dask, persist, gridded, label, value_label)
    270             data = data.persist() if persist else data
    271         else:
--> 272             data = dataset.to_dataframe(name=name)
    273             if len(data.index.names) > 1:
    274                 data = data.reset_index()

TypeError: to_dataframe() got an unexpected keyword argument 'name'
```

jsignell:
Closes #125

### DIFF
--- a/hvplot/util.py
+++ b/hvplot/util.py
@@ -249,8 +249,9 @@ def process_xarray(data, x, y, by, groupby, use_dask, persist, gridded, label, v
     index_dims = [d for d in dims if d in dataset.indexes]
 
     if gridded:
-        print('GRIDDED')
         data = dataset
+        if len(dims) < 2:
+            dims += [dim for dim in list(data.dims)[::-1] if dim not in dims]
         if not (x or y):
             x, y = index_dims[:2] if len(index_dims) > 1 else dims[:2]
         elif x and not y:
@@ -268,6 +269,8 @@ def process_xarray(data, x, y, by, groupby, use_dask, persist, gridded, label, v
             data = dataset.to_dataframe()
             if len(data.index.names) > 1:
                 data = data.reset_index()
+        if len(dims) == 0:
+            dims = ['index']
         if x and not y:
             y = dims[0] if x in data_vars else data_vars
         elif y and not x:
--- a/hvplot/util.py
+++ b/hvplot/util.py
@@ -237,18 +237,20 @@ def is_geopandas(data):
 
 def process_xarray(data, x, y, by, groupby, use_dask, persist, gridded, label, value_label):
     import xarray as xr
-    dataset = data
-    data_vars = list(dataset.data_vars) if isinstance(data, xr.Dataset) else [data.name]
+    if isinstance(data, xr.Dataset):
+        dataset = data
+    else:
+        name = data.name or label or value_label
+        dataset = data.to_dataset(name=name)
+
+    data_vars = list(dataset.data_vars)
     ignore = (by or []) + (groupby or [])
-    dims = [c for c in data.coords if data[c].shape != () and c not in ignore][::-1]
-    index_dims = [d for d in dims if d in data.indexes]
+    dims = [c for c in dataset.coords if dataset[c].shape != () and c not in ignore][::-1]
+    index_dims = [d for d in dims if d in dataset.indexes]
 
     if gridded:
+        print('GRIDDED')
         data = dataset
-        if data_vars == [None]:
-            label = label or value_label
-            data = data.to_dataset(name=label)
-            data_vars = [label]
         if not (x or y):
             x, y = index_dims[:2] if len(index_dims) > 1 else dims[:2]
         elif x and not y:
@@ -259,17 +261,11 @@ def process_xarray(data, x, y, by, groupby, use_dask, persist, gridded, label, v
             dims = list(data.coords[x].dims) + list(data.coords[y].dims)
             groupby = [d for d in index_dims if d not in (x, y) and d not in dims]
     else:
-        name = None
-        if not isinstance(dataset, xr.Dataset):
-            name = dataset.name or label or value_label
-            data_vars = [name]
         if use_dask:
-            if not isinstance(dataset, xr.Dataset):
-                dataset = dataset.to_dataset(name=name)
             data = dataset.to_dask_dataframe()
             data = data.persist() if persist else data
         else:
-            data = dataset.to_dataframe(name=name)
+            data = dataset.to_dataframe()
             if len(data.index.names) > 1:
                 data = data.reset_index()
         if x and not y:
--- a/hvplot/tests/testutil.py
+++ b/hvplot/tests/testutil.py
@@ -1,0 +1,149 @@
+"""
+Tests  utilities to convert data and projections
+"""
+import numpy as np
+
+from unittest import TestCase, SkipTest
+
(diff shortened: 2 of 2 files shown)


hvplot/util.py
```diff
@@ -237,18 +237,21 @@ def is_geopandas(data):
...
+    index_dims = [d for d in dims if d in dataset.indexes]
 
     if gridded:
         data = dataset
-        if data_vars == [None]:
-            label = label or value_label
-            data = data.to_dataset(name=label)
-            data_vars = [label]
+        if len(dims) < 2:
+            dims += [dim for dim in list(data.dims)[::-1] if dim not in dims]
```
philippjfr: Could you explain this?

jsignell: yeah I was trying to take whatever is already in `dims` and extend it with any `data.dims` that aren't included. Also I wanted to flip the order in which I add the `data.dims` so that image works out right for instance.

jsignell: the `dims` are set above from coords.

philippjfr: >Also I wanted to flip the order in which I add the data.dims so that image works out right for instance.

This worries me a little bit, but I guess it's not an issue because of the conditional (`if len(dims) < 2:`).

jsignell: Right that is why it doesn't break pre-existing behavior. I know it it strange but I think it is very nice to do the right thing with an image with no coords.

philippjfr: Okay, sounds fine. My only question is about the case where ``len(dims) == 1``, I think that should be unchanged. If there are coordinates defined it should use those and if there aren't it should fall back to the dims, but I wouldn't mix the two cases.

jsignell: why not mix? I tested that scenario in the tests and it works fine.

philippjfr: It seems to me that if coordinates were declared there was probably a reason for declaring coordinates for certain dimensions but not others, so it seems to me like this is dropping information the user has provided.

philippjfr: I might not be thinking about this correctly though and ignoring the dimensions would actually result in an error or something.

jsignell: I think ignoring the dimensions would result in an error. The coordinates will always go in first and if the user has defined x and y then those will be used.

jsignell: Yeah I ran the relevant test on master:

```python
hvplot/tests/testutil.py::TestProcessXarray::test_process_2d_xarray_dataarray_with_coords_as_gridded FAILED [100%]

=================================================== FAILURES ====================================================
___________________ TestProcessXarray.test_process_2d_xarray_dataarray_with_coords_as_gridded ___________________

self = <hvplot.tests.testutil.TestProcessXarray testMethod=test_process_2d_xarray_dataarray_with_coords_as_gridded>

    def test_process_2d_xarray_dataarray_with_coords_as_gridded(self):
        import xarray as xr

        da = xr.DataArray(
            data=np.random.randn(4,5),
            coords={'y': [3, 4, 5, 6, 7]},
            dims=['x', 'y'])

        kwargs = self.default_kwargs
        kwargs.update(gridded=True)

>       data, x, y, by, groupby = process_xarray(data=da, **kwargs)

hvplot/tests/testutil.py:119:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

data = <xarray.Dataset>
Dimensions:  (x: 4, y: 5)
Coordinates:
  * y        (y) int64 3 4 5 6 7
Dimensions without coordinates: x
Data variables:
    value    (x, y) float64 -1.406 1.064 0.0174 1.719 ... 0.1767 0.4939 -1.138
x = None, y = None, by = None, groupby = None, use_dask = False, persist = False, gridded = True, label = 'value'
value_label = 'value'

    def process_xarray(data, x, y, by, groupby, use_dask, persist, gridded, label, value_label):
        import xarray as xr
        dataset = data
        data_vars = list(dataset.data_vars) if isinstance(data, xr.Dataset) else [data.name]
        ignore = (by or []) + (groupby or [])
        dims = [c for c in data.coords if data[c].shape != () and c not in ignore][::-1]
        index_dims = [d for d in dims if d in data.indexes]

        if gridded:
            data = dataset
            if data_vars == [None]:
                label = label or value_label
                data = data.to_dataset(name=label)
                data_vars = [label]
            if not (x or y):
>               x, y = index_dims[:2] if len(index_dims) > 1 else dims[:2]
E               ValueError: not enough values to unpack (expected 2, got 1)

hvplot/util.py:253: ValueError
==================================== 1 failed, 7 deselected in 2.23 seconds =====================================

jsignell: So I think the it is better to plot something and let the user say "hmm, that's not what I meant" than to plot nothing.

PR #160, opened 2019-02-07
New style readme [skip ci]

jsignell:
Not sure if you'll want the coverage row in there. 

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,12 +1,18 @@
-[![linux/mac build status](https://travis-ci.org/pyviz/hvplot.svg?branch=master)](https://travis-ci.org/pyviz/hvplot)
-
 <h1>
 hvPlot <img src="/doc/_static/hvplot-logo.png" width="50" height="50">
 </h1>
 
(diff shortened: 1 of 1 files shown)


jsignell:
Is this ok with you @philippjfr

philippjfr:
Yes, thanks.

jbednar:
Can we find a standard way to make a link to the overall status dashboard that shows all the PyViz projects?  It would be nice if each project had a link back both to pyviz.org and also to status.pyviz.org in the same place on each project.

jsignell:
Maybe there should be an "About Pyviz" section at the end of each readme

jbednar:
That would probably work:

# About PyViz

<project_name> is part of the PyViz initiative for making Python-based visualization tools work well together.  See [pyviz.org](http://pyviz.org) for related packages that you can use with <project_name> and [status.pyviz.org](http://status.pyviz.org) for the current status of each PyViz project.

PR #172, opened 2019-02-20
If line or scatter, warn that it will not project

ahuang11:
Because I spent half an hour being dumbfounded why coastlines weren't showing even though the lats and lons were within bounds!


philippjfr:
Thanks @ahuang11!

PR #193, opened 2019-03-24
Add min_count option for hexbin plots

ivirshup:
Pretty much just added this option in to `_kind_options`, made sure the converter would pass it if specified, and copied relevant docs from holoviews.

Now you can do this:

image


philippjfr:
Looks great, thanks for the PR!

PR #195, opened 2019-04-01
When persist is passed to process_intake, persist dask array

jsignell:
Persist was already a valid argument - but was just being used to return a dask array. This PR makes it return a persisted array.


jsignell:
@philippjfr ok with you if I merge this?

jsignell:
Ok Philipp explained to me that it already worked. It just happened in a different spot than we were looking, so this PR isn't necessary.

PR #206, opened 2019-05-09
Determine whether data is divergent and use a RdBu colormap / center clim

ahuang11:
https://github.com/pyviz/hvplot/issues/205

The centering is based off all time step / all dimensions.

```
import hvplot.xarray
import xarray as xr

da = ds['air']
da_anom = da - da.mean()
da_anom.hvplot('lon', 'lat')
```

image

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -488,7 +488,8 @@ class HoloViewsConverter(object):
             divergent = cmin < 0 and cmax > 0
             if divergent:
                 symmetric = True
-            # TODO: add cmap from hv.config
+            else:
+                symmetric = False
 
         return symmetric
 
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -465,7 +465,6 @@ class HoloViewsConverter(object):
         # Process dimensions and labels
         self.label = label
         self._relabel = {'label': label} if label else {}
-
         self._dim_ranges = {'c': clim or (None, None)}
 
         # High-level options
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -293,7 +293,6 @@ class HoloViewsConverter(object):
                  y_sampling=None, project=False, tools=[],
                  attr_labels=None, coastline=False, tiles=False,
                  sort_date=True, **kwds):
-
         # Process data and related options
         self._redim = fields
         self.use_index = use_index
@@ -448,12 +447,17 @@ class HoloViewsConverter(object):
         if title is not None:
             plot_opts['title_format'] = title
 
-        try:
-            if self.kind in self._colorbar_types and not use_dask:
-                plot_opts['symmetric'] = self._process_symmetric(
-                    clim, symmetric, kwds)
-        except TypeError:
-            pass
+        if (self.kind in self._colorbar_types or plot_opts.get('colorbar')) and not use_dask:
+            symmetric= self._process_symmetric(symmetric, clim)
+
+            cmap = kwds.pop('cmap', kwds.pop('colormap', None))
+            if cmap is None:
+                if symmetric:
+                    self._style_opts['cmap'] = self._default_cmaps['diverging']
+                else:
+                    self._style_opts['cmap'] = self._default_cmaps['linear']
+
+            plot_opts['symmetric'] = symmetric
 
         self._plot_opts = plot_opts
         self._overlay_opts = {k: v for k, v in self._plot_opts.items()
@@ -475,29 +479,26 @@ class HoloViewsConverter(object):
             param.main.warning('Plotting {kind} plot with parameters x: {x}, '
                                'y: {y}, by: {by}, groupby: {groupby}'.format(**kwds))
 
-    def _process_symmetric(self, clim, symmetric, kwds):
-        if clim is not None and symmetric is not None:
+    def _process_symmetric(self, symmetric, clim):
+        if symmetric is not None:
             return symmetric
+        elif clim is not None:
+            return False
 
         if is_xarray(self.data):
             # chunks mean it's lazily loaded; nanquantile will eagerly load
             if self.data.chunks:
                 return False
-            z = kwds.get('z', list(self.data.data_vars)[0])
-            data = self.data[z]
+            data = self.data[self.z]
+        elif self._color_dim:
+            data = self.data[self._color_dim]
         else:
-            data = self.data[kwds['C']]
+            return False
 
         cmin = np.nanquantile(data, 0.05)
         cmax = np.nanquantile(data, 0.95)
-        clim = (cmin, cmax)
-        divergent = cmin < 0 and cmax > 0
-        if divergent:
-            symmetric = True
-        else:
-            symmetric = False
 
-        return symmetric
+        return bool(cmin < 0 and cmax > 0)
 
     def _process_crs(self, data, crs):
         """Given crs as proj4 string, data.attr, or cartopy.crs return cartopy.crs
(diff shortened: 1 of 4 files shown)


hvplot/converter.py
```diff
@@ -387,6 +396,42 @@ def __init__(self, data, x, y, kind=None, by=None, use_index=True,
...
+        if center:
+            if auto:  # if user provided clim, don't change it
+                cmax = np.max([np.abs(cmin), np.abs(cmax)])
+                cmin = -cmax  # make top == bottom
+                clim = (cmin, cmax)
+
+            # if cmap isn't explicitly provided and the values center
+            # around 0 use the RdBu_r colormap instead of fire colormap
+            if not self._style_opts.get('cmap', None):
+                self._style_opts['cmap'] = 'RdBu_r'
```
jbednar: I don't think this colormap name should be hardcoded; for one thing it's not perceptually uniform.  The percetpually uniform `coolwarm` or `CET_D1A` palettes from [Colorcet](http://colorcet.pyviz.org/user_guide/Continuous.html) could be used, or `balance` from [cmocean](https://matplotlib.org/cmocean), but in any case I think it should be a parameter somewhere that the user can modify.

ahuang11: It can be modified using specifying a cmap, unless you mean adding this functionality to holoviews' `bokeh/__init__.py` and `matplotlib/__init__.py` store?

jbednar: This PR provides the ability to automatically switch to a divergent colormap, right?  I'm saying that the colormap that is switched to in such a case should be configurable.  `coolwarm` from Matplotlib isn't perceptually uniform, while `coolwarm` from colorcet isn't necessarily available, so people aren't going to agree on what should be switched to.  So users should be able to define something like `hvplot.default_diverging_cmap="CET_D1A"`, allowing them to choose the diverging colormap without having to lose the auto-switching.  @philippjfr would be able to say where such a default could be defined.

hvplot/converter.py
```diff
@@ -387,6 +396,42 @@ def __init__(self, data, x, y, kind=None, by=None, use_index=True,
             param.main.warning('Plotting {kind} plot with parameters x: {x}, '
                                'y: {y}, by: {by}, groupby: {groupby}'.format(**kwds))
 
+    def _process_clim(self, clim, center):
+        if clim is None:
+            if is_xarray(self.data):
+                data = self.data[list(self.data.data_vars)[0]]
+            else:
+                data = self.data
```
philippjfr: Confused how this would work for a dataframe.

hvplot/converter.py
```diff
@@ -215,7 +216,8 @@ def __init__(self, data, x, y, kind=None, by=None, use_index=True,
                  crs=None, fields={}, groupby=None, dynamic=True,
                  width=700, height=300, shared_axes=True,
                  grid=False, legend=True, rot=None, title=None,
-                 xlim=None, ylim=None, clim=None, xticks=None, yticks=None,
+                 xlim=None, ylim=None, clim=None, center=None,
```
philippjfr: I'd say `center` should be called `symmetric` which is the equivalent keyword in holoviews.

hvplot/converter.py
```diff
@@ -387,6 +396,42 @@ def __init__(self, data, x, y, kind=None, by=None, use_index=True,
...
+        if center is None:
+            divergent = cmin < 0 and cmax > 0
+            # if the order of magnitudes equate; we don't want to
+            # center the clim or use a divergent colormap
+            # if cmin == -1 and cmax == 20 since it's mostly positive
+            equal_oom = order_magn(cmin) == order_magn(cmax)
+            center = divergent and equal_oom
+
+        if center:
+            if auto:  # if user provided clim, don't change it
```
philippjfr: I'd simply set the symmetric plot option instead of overriding the clim.

hvplot/converter.py
philippjfr: I'd prefer if we defined the default colormaps for these three cases in holoviews on ``hv.config``:

* numerical
* categorical
* diverging

Cc: @jlstevens

jbednar: I'd call the first one `sequential`, but otherwise sounds good!

jlstevens: Seems reasonable to me. This just acknowledges the fact there are different types of colormaps so having multiple default makes sense.

ahuang11: I'm unfamiliar with using hv.config; is hv.config under holoviews/holoviews/core/util.py/Config (class)

If so, do I just add a param?

jbednar: @jlstevens , can you advise?

hvplot/converter.py
```diff
@@ -387,6 +396,22 @@ def __init__(self, data, x, y, kind=None, by=None, use_index=True,
...
+            else:
+                data = self.data[kwds['C']]
+
+            cmin = np.nanquantile(data, 0.05)
+            cmax = np.nanquantile(data, 0.95)
+            clim = (cmin, cmax)
+            divergent = cmin < 0 and cmax > 0
+            if divergent:
+                symmetric = True
+            # TODO: add cmap from hv.config
```
jsignell: Note that #258 adds a default diverging colormap so that would be the right thing to use here.

ahuang11: Thanks! Not sure if the way I implemented it is logical though. symmetric is under plot_opts, but the cmaps are decided under process_style which uses style_opts

hvplot/converter.py
```diff
@@ -348,9 +348,17 @@ def __init__(self, data, x, y, kind=None, by=None, use_index=True,
 
         # Process options
         self.stacked = stacked
-        self._style_opts, kwds = self._process_style(kwds)
 
-        plot_opts = {**self._default_plot_opts,
+        try:
+            if self.kind in self._colorbar_types and not use_dask:
```
philippjfr: `use_dask` is only effective for intake datasets, when operating directly on a dask dataframe or dask array backed xarray the data may still be out-of-core. I think that will need to be handled inside `_process_symmetric`.

hvplot/converter.py
```diff
@@ -451,6 +459,23 @@ def __init__(self, data, x, y, kind=None, by=None, use_index=True,
             param.main.warning('Plotting {kind} plot with parameters x: {x}, '
                                'y: {y}, by: {by}, groupby: {groupby}'.format(**kwds))
 
+    def _process_symmetric(self, clim, symmetric, kwds):
+        if clim is None and symmetric is None:
+            if is_xarray(self.data):
+                data = self.data[list(self.data.data_vars)[0]]
```
philippjfr: This will probably have to handle an explicit `z`.

hvplot/converter.py
```diff
@@ -451,6 +459,23 @@ def __init__(self, data, x, y, kind=None, by=None, use_index=True,
             param.main.warning('Plotting {kind} plot with parameters x: {x}, '
                                'y: {y}, by: {by}, groupby: {groupby}'.format(**kwds))
 
+    def _process_symmetric(self, clim, symmetric, kwds):
+        if clim is None and symmetric is None:
```
philippjfr: Should return early and then un-indent the rest: 

```python
if clim is not None or symmetric is not None:
    return symmetric
```

hvplot/converter.py
ahuang11: Does handling it mean skipping it?
```
            # chunks mean it's lazily loaded; nanquantile will eagerly load
            if self.data.chunks:
                return False
```

jsignell:
Ok I think this is reasonable now.

jsignell:
Well I guess it's not quite done. Still need some more tests around non-numeric datatypes I suppose. Or put back in the try/except.

jsignell:
@philippjfr good with you if I merge?

PR #230, opened 2019-07-01
add plot function for pandas plot api

joy-rosie:
Following from pandas pull request https://github.com/datapythonista/pandas/pull/9 from @datapythonista, added a plot function to hvplot/\_\_init\_\_.py to help with pandas plot directly rather than having to go through monkey patch.

For pandas side, we need to use https://github.com/datapythonista/pandas/tree/plot_api to test the plotting.

There is a notebook created to try and test the plot cases using the new pandas plot api; https://nbviewer.jupyter.org/github/joy-rosie/test_pandas_hvplot/blob/master/PlottingExamples.ipynb?flush_cache=true. 

Feel to to comment/give suggestions to this. There are a few warnings and errors which still need to be dealt with.


philippjfr:
Thanks for getting this started. I'll try to make a push for a 0.5.0 release in the coming weeks and hopefully fill in some of the gaps in the API. We will likely simply have to suppress some of the warnings or drop certain kwargs entirely in the plot function. @datapythonista do you know the rough timeline for merging the PR and getting it into a pandas release just so I have a rough idea about the timescales involved?

datapythonista:
https://github.com/pandas-dev/pandas/pull/27009 should be merged today, and we're planning to release pandas 0.25rc1 also today :)

.gitignore
```diff
@@ -1,3 +1,6 @@
+# Pycharm
+.idea/
```
datapythonista: In general it's a good practice to keep unrelated changes in separate PRs, even if they are small changes like this.

datapythonista:
I updated the pandas branch `plot_api`, now it only passes the `kwargs` specified by the user to backends (it still passes all them to the matplotlib one, since that is needed at this point).

philippjfr:
Thanks @datapythonista, very much appreciated. I'll try to spend some time next week to make a number of fixes and updates to hvplot and hopefully release 0.5.0 next week.

joy-rosie:
@datapythonista , I updated the notebook with some of the tests and I get a warning for 'reuse_plot' option as seen in https://nbviewer.jupyter.org/github/joy-rosie/test_pandas_hvplot/blob/master/PlottingExamples.ipynb?flush_cache=true

datapythonista:
Thanks @joy-rosie that makes sense. I think the best is that the backend receives this method from pandas, we should probably add it to the signature of `Series.plot()`. Does hvplot has an equivalent? If it doesn't have this option now, I'd say that you can capture it in the `plot` function signature, and ignore it for now.

datapythonista:
I've been having a look, and this looks to work quite well.

There are couple of small things, not about this PR but about `hvplot` that doesn't seem right to me, not sure if I'm missing something.

- The scatter plot doesn't show the index when hovering (I think that would be very useful, one of the main reasons I wanted to have pandas plots in bokeh/hvplot)
- The margins of the plots doesn't seem correct. They seem to be set at the exact minimums and maximums, and I think something like a 10% more should be left.

This is what it's generated by default:

hvplot_margins

This is what I would prefer:

hvplot_margins_ok

@philippjfr let me know if those make sense, and I'll create issues for them.

Btw, are you thinking on keeping `Series.hvplot` for backward compatibility in the next release, or are you planning to drop it?

philippjfr:
>The scatter plot doesn't show the index when hovering (I think that would be very useful, one of the main reasons I wanted to have pandas plots in bokeh/hvplot)

An issue about this would be appreciated.

>The margins of the plots doesn't seem correct. They seem to be set at the exact minimums and maximums, and I think something like a 10% more should be left.

Yes, in the next release I'll enable some minimum padding.

>Btw, are you thinking on keeping Series.hvplot for backward compatibility in the next release, or are you planning to drop it?

Is there any change in the behavior in pandas? Series.hvplot worked by promoting the Series to a DataFrame and then using the same codepath as DataFrame.hvplot in the past. Is that behavior what you would expect?

datapythonista:
I didn't mean specifically the `Series.hvplot` as opposed to `DataFrame.hvplot`, but keeping `(Series|DataFrame).hvplot` as opposed to the new `(Series|DataFrame).plot` with the `backend.plotting` option.

The transformations from `Series` to `DataFrame` are done here: https://github.com/pandas-dev/pandas/blob/master/pandas/plotting/_core.py#L717 and I guess they can affect you, for what I saw the `DataFrame` plots are usually sent to the backend as a `Series` plot with the `x` column as the index and the `y` column as the values. Not sure if you were doing this same conversion before.

I'll create an issue and a PR for adding the index to the hover, I want to show it in a lightning talk this weekend.

philippjfr:
>I didn't mean specifically the Series.hvplot as opposed to DataFrame.hvplot, but keeping (Series|DataFrame).hvplot as opposed to the new (Series|DataFrame).plot with the backend.plotting option.

Ah, yes I'll likely keep it for the next release. I was considering issuing a deprecation warning in the release after that and then finally remove it. That said I still have to consider what to do about the dask and xarray plot interfaces.

>I'll create an issue and a PR for adding the index to the hover, I want to show it in a lightning talk this weekend.

Sounds good, let me know if you need help. I wanted to do some refactoring and clean some stuff up this weekend, but maybe I'll wait until after your PR goes in.

joy-rosie:
Hello @philippjfr , I was wondering if I can help with anything on this PR?

jsignell:
This was superseded by #319. But thanks @joy-rosie for holding our feet to the fire on this :)

PR #234, opened 2019-07-07
Add convenient coastline kwd

ahuang11:
https://github.com/pyviz/hvplot/issues/232

Makes exploration of geographic datasets slightly faster and code a bit cleaner. I wonder though if returning an Overlay is fine and whether this should be under geoviews as an `.opts`, i.e. `plot.opts(coastline=True)`?

```
import hvplot.xarray
import xarray as xr
ds = xr.tutorial.open_dataset('air_temperature')['air']

ds.hvplot('lon', 'lat', coastline=True)
```

Without this feature:
```
import hvplot.xarray
import xarray as xr
import geoviews as gv
ds = xr.tutorial.open_dataset('air_temperature')['air']

ds.hvplot('lon', 'lat', geo=True) * gv.feature.coastline()
```
image

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -227,7 +227,7 @@ class HoloViewsConverter(object):
                  dynspread=False, hover_cols=[], x_sampling=None,
                  y_sampling=None, project=False, xlabel=None, ylabel=None,
                  clabel=None, xformatter=None, yformatter=None, tools=[],
-                 padding=None, **kwds):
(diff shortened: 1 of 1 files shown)


jsignell:
This will be really convenient! It just needs a rebase and some tests.

philippjfr:
I think I'll close and reopen in a new PR, the approach taken here isn't really extensible.

PR #248, opened 2019-07-17
Allow list types for xlim,ylim tuples

jbednar:
Addresses https://github.com/pyviz/datashader/issues/766 . @philippjfr , not sure if you'd want to allow *only* list types, or any sequence as here.  An alternative to this PR would be to change param.Tuple to coerce inputs to tuples in general, but that does cut down on detecting some genuine error conditions, so I made this more conservative PR instead.


philippjfr:
I'm +1 on this fix.

PR #249, opened 2019-07-19
Only include numeric columns by default if available in data

### LINKED ISSUE #241: Line plot should ignore non-numeric column
I got this example:
```python
df = pandas.DataFrame({'foo': [1, 2, 3, 4],
                       'bar': [3, 1, 4, 2],
                       'label': list('xyzh')},
                      index=list('abcd'))
df.hvplot(kind='line')
```

Which is generating the next plot:

hvplot_line

Not sure about other cases, but I think in this case it'd make sense to ignore the `label` column, instead of adding its values to the y axis. That's what the pandas matplotlib backend does.

jsignell:
Closes #241 

Not quite sure if this is possible with a dask dataframe.

Screen Shot 2019-07-19 at 3 09 50 PM

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -919,6 +919,12 @@ class HoloViewsConverter(object):
         y = y or self.y
         if not y:
             ys = [c for c in data.columns if c not in [x]+self.by+self.groupby]
+            if len(ys) > 1 and self.datatype == 'pandas':
+                # if columns have different dtypes, only include numeric columns
+                from pandas.api.types import is_numeric_dtype as isnum
+                num_ys = [col for col in data.columns if isnum(data[col])]
+                if len(num_ys) >= 1:
+                    ys = num_ys
             y = ys[0] if len(ys) == 1 else ys
         return data, x, y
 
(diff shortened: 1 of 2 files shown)


hvplot/converter.py
```diff
@@ -878,6 +878,12 @@ def _process_args(self, data, x, y):
         y = y or self.y
         if not y:
             ys = [c for c in data.columns if c not in [x]+self.by+self.groupby]
+            if len(ys) > 1 and self.datatype == 'pandas':
```
philippjfr: I think this should also work or be made to work for datatype='dask'/'streamz'.

jsignell:
@philippjfr do you think I need a streamz test?

philippjfr:
>@philippjfr do you think I need a streamz test?

Might be nice, but I'll leave it up to you.

jsignell:
I have tested it locally and things seem to work properly but it is tricky to test because the data aren't pulled until render, so in the tests where there is no render, it is hard to figure out what is going on. I am inclined to merge as is and think more about streamz testing in a different PR

philippjfr:
If you ever want to test the rendering, you have get a renderer with:

```python
renderer = hv.renderer('bokeh')
renderer.get_plot(hv_obj)
```

Issue #178, opened 2019-03-01
Default color maps makes points invisible


ivirshup:
Points with the maximum value are invisible by default for scalar valued color maps. I think this is bad default behavior. Here's an example:

```python
import pandas as pd
import numpy as np
import hvplot.pandas
from itertools import repeat

d = pd.DataFrame({"x": np.arange(10), "y": np.arange(10), "c": np.ones(10)})
d.loc[5, "c"] = 0

d.hvplot.scatter("x", "y", c="c")
```
<img width="788" alt="image" src="https://user-images.githubusercontent.com/8238804/53612029-c1d59500-3c24-11e9-8ca8-3ff66a9d2580.png">

philippjfr:
Definitely agreed on that, related discussion in https://github.com/pyviz/holoviews/issues/3500

philippjfr:
I think we should change the default colormap in hvPlot independently from HoloViews.

jbednar:
See discussion in https://github.com/pyviz/holoviews/issues/3500 before deciding...

jsignell:
Just to note that matplotlib has switched the colormap direction so that dark colors indicate high values by default, so I think it would be relatively safe for us to move to `kbc_r`.

```python
ax = d.plot.scatter("x", "y", c="c")
d.hvplot.scatter("x", "y", c="c", cmap='kbc_r', width=400)
```
<img width="959" alt="Screen Shot 2019-07-24 at 1 48 22 PM" src="https://user-images.githubusercontent.com/4806877/61816235-f4716b80-ae19-11e9-8d6f-ec765405513d.png">

jbednar:
Oh, that's great news!  I think that's really helpful for the future of science, which appears to be occurring on a white background. :-) Can you link to the PR where that change was made (here and in https://github.com/pyviz/holoviews/issues/3500), so that we have the surrounding discussion for context?

jsignell:
We could even check for if `bgcolor` is set and flip the cmap if it's `black`.

jbednar:
Sure; that's what my old code for running simulations did (circa 1995...)

jsignell:
I can't quite find the PR or even a blog post announcing the change.

jsignell:
closed in #258

PR #258, opened 2019-07-23
Use default opts if set, changing default numeric cmap

### LINKED ISSUE #231: Setting global default plot options
I would like to set default options globally for the plots generated by hvplots.

Normally in holoviews I would do this to set defaults for every plot:

    import holoviews as hv
    from holoviews import opts
    hv.extension('bokeh')

    # Set default options here
    opts.defaults(opts.Curve( height=400, width=900 ,show_grid=True))

Now that I'm using hvplots the import has become the following for xarray based plotting

    import xarray as xr
    import hvplot.xarray

If I have an xarray dataset and I want to plot one of the dimensions then I use this

    ds.myDims.hvplot()

However this gives a plot with a default width and no grid lines. To add this I need to specify the options with the plot.

    ds.myDims.hvplot(height=400, width=900,grid=True)

Ideally I'd like to set the height, width, grid etc, colours etc globally, like I can in holoviews, so I don't have to specify it for every plot.

jsignell:
Closes #231, #178

I added some tests for the plot_opts and cmap/colormap. Also made it possible to pass `cmap='cyclic'` so leaving open the option to do diverging by default or something.


jsignell:
In trying to write tests I found what I suspect is a shortcoming in holoviews. I don't think it is possible to overwrite `opts.defaults(xaxis=None)` in a particular opts call because in order to turn the axis back on, you need to unset it from the dict. 

If you try to set it to True you get this error: 

```python-traceback
/conda/envs/hvplot-dev/lib/python3.7/site-packages/holoviews/plotting/bokeh/element.py in _init_axes(self, plot)
    562         if self.xaxis is None:
    563             plot.xaxis.visible = False
--> 564         elif 'top' in self.xaxis:
    565             plot.above = plot.below
    566             plot.below = []
```

I am leaving in a failing test to demonstrate.

philippjfr:
>If you try to set it to True you get this error:

Hmm, yeah that's a bug, it's listed in the ObjectSelector but not actually handled.

jsignell:
@philippjfr I just noticed the discussion in https://github.com/pyviz/hvplot/pull/206#discussion_r297351822. Is this still something that we will want the user to be able to configure? Do you think what I've implemented here is a good first step before migrating the functionality up to holoviews?

hvplot/converter.py
```diff
@@ -118,7 +118,7 @@ class HoloViewsConverter(object):
     rot: number
         Rotates the axis ticks along the x-axis by the specified
         number of degrees.
-    shared_axes (default=False): boolean
+    shared_axes (default=True): boolean
```
jbednar: While I agree with this change, doesn't it need discussion?  Seems like a change in behavior that should be called out.

philippjfr:
I'm happy to have this be configurable in hvplot for now.

hvplot/converter.py
```diff
@@ -252,26 +252,33 @@ class HoloViewsConverter(object):
...
 
+    _default_plot_opts = {
+        'logx': False, 'logy': False, 'show_legend': True, 'legend_position': 'right',
+        'show_grid': False, 'responsive': False, 'shared_axes': True}
+
+    _default_cmaps = {
+        'linear': 'kbc_r',
+        'categorical': 'Category10',
+        'cyclic': 'colorwheel',
+        'diverging': 'coolwarm'
```
jbednar: Which 'coolwarm' is this?  It looks like the linear and cyclic defaults are taken from colorcet, so I assume it's colorcet, but it's ambiguous as matplotlib also has a coolwarm.  Colorcet's one was meant to be prefixed with cet_ to avoid confusion, but I think the prefix isn't used in some cases (?) in holoviews, so I don't know what it means here.

hvplot/converter.py
```diff
@@ -252,26 +252,33 @@ class HoloViewsConverter(object):
...
-                 precompute=False, flip_xaxis=False, flip_yaxis=False,
+                 framewise=True, aggregator=None,
+                 projection=None, global_extent=None, geo=False,
+                 precompute=False, flip_xaxis=None, flip_yaxis=None,
                  dynspread=False, hover_cols=[], x_sampling=None,
-                 y_sampling=None, project=False, xlabel=None, ylabel=None,
-                 clabel=None, xformatter=None, yformatter=None, tools=[],
-                 padding=None, responsive=False, min_width=None,
-                 min_height=None, max_height=None, max_width=None,
+                 y_sampling=None, project=False, tools=[],
```
jbednar: What is the implication of losing the various constructor arguments (colormap, clabel, etc.)?

hvplot/converter.py
jsignell: There was a discrepancy between the docstring and the implementation. It `shared_axes` **was** and **is** `True` by default

hvplot/converter.py
jsignell: I had assumed colorcet. Since matplotlib isn't supported as a backend yet we probably don't have to think about it too hard. Also I am just setting up this dict as a stop-gap until holoviews sets up a config option.

hvplot/converter.py
jsignell: I don't think there is an implication. The same defaults are still being fed in. It is easier to add `plot_opts` now since you'd just add to the `plotwds` list below. These were never exposed via tab completion, so I dont' think they were really serving anyone.

hvplot/converter.py
jbednar: I don't think the lack of an mpl backend will affect whether it's mpl's coolwarm; I think the way HoloViews handles colormaps depends not on the selected backend but just on what's installed and importable on the system.  I don't really know the details of what hv does here, though.

PR #259, opened 2019-07-24
Sort by date by default when using pandas objects

### LINKED ISSUE #121: hvPlot doesnot sort date automatically like pandas.plot()
Simple setup 
```
import numpy as np
import pandas as pd
index = pd.date_range('1/1/2000', periods=1000)
index = [d.strftime("%a, %d %b %Y %H:%M:%S +0000") for d in index]
df = pd.DataFrame(np.random.randn(1000, 4), columns=list('ABCD')).cumsum()
shuffle(index)
df['date'] = index
df['date'] = df["date"].apply(pd.to_datetime)
df.head()
```

With pandas plot()
```
%matplotlib inline
df.plot(x='date')
```
image

```
import hvplot.pandas
df.hvplot(x='date')
```
image

The graph is sorted if no type is applied to **date** column (just comment the line to the the result).

jsignell:
Closes #121 

Also includes a kwarg to turn off this sorting in case it is undesirable. For now I am specifically only doing this for pandas objects because sorting for streamz is a very bad idea, and dask could also be strange. I think starting with pandas will at least be an easier change to default behavior, then if we want to expand it, that will be easy enough.

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -141,6 +141,8 @@ class HoloViewsConverter(object):
         The width and height of the plot in pixels
     attr_labels (default=True): bool
         Whether to use an xarray object's attributes as labels
+    sort_date (default=True): bool
+        Whether to sort the x-axis by date before plotting
 
     Datashader options
     ------------------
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks!

PR #262, opened 2019-07-25
Fix label for xr.Dataset

### LINKED ISSUE #261: Fix warning from #173
In #173 a new warning came in that is very loud and assumes that things are dataarrays. 

```python
ds.hvplot(groupby='time')
ds.mean(dim=['lat', 'lon']).hvplot()
```

Screen Shot 2019-07-25 at 4 01 57 PM

Screen Shot 2019-07-25 at 3 56 16 PM

jsignell:
Closes #261 

The last commit in this PR makes the warning only show if the user explicitly sets `attr_labels` to `True`. Not sure whether or not that change is acceptable. But I kept the commit separate so that it is easily revertable if people don't like that idea.

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -638,7 +638,10 @@ class HoloViewsConverter(object):
         if da is not None and attr_labels:
             try:
                 var_tuples = [(var, da[var].attrs) for var in da.coords]
-                var_tuples.append((da.name, da.attrs))
+                if isinstance(da, xr.Dataset):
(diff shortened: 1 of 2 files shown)


philippjfr:
Hmm, don't think this is sufficient, will need to work out a better fix for Dataset.

jsignell:
> Hmm, don't think this is sufficient, will need to work out a better fix for Dataset.

What were you seeing in particular? 

Also do you agree with failing silently unless the user explicitly asks for this functionality?

philippjfr:
Nevermind missed your other commit, looks good!

PR #173, opened 2019-02-20
Label from attributes

ahuang11:
https://github.com/pyviz/hvplot/issues/155
examples https://anaconda.org/ahuang11/test_attr_labels

image

I tried mimicking new line if label too long, but I don't think bokeh supports '\n'
image

I'm happy where it is now I think.

PR #264, opened 2019-07-26
Using attrs to set default x and y for gridded outputs

jsignell:
After #173 went in, I realized that there is more info in the coords attrs that can be used to set better defaults. 

This PR:
Screen Shot 2019-07-26 at 1 11 43 PM

For reference, this is the output of the same call on master:
Screen Shot 2019-07-26 at 1 19 57 PM


philippjfr:
Great! Feel free to merge if it's ready.

Issue #192, opened 2019-03-22
geo=True should force aspect to be 1:1


jsignell:
Now that layouts has landed in bokeh - it'd be nice to do something smarter if geo=True in hvplot. Can we force aspect to be 1:1 by default?

philippjfr:
It would be nice if we could, and we should be able to do so safely as long as ``responsive=False``. Bokeh does not yet provide a way to guarantee a fixed data aspect if ``responsive=True``.

philippjfr:
Is this true for all projections? I assume the answer is yes, but I'm not 100% sure.

jsignell:
I am not entirely sure, but I think it'd be a better default anyways. Can always raise another issue later if it's not...

Issue #75, opened 2018-09-07
Is there a way to control aspect ratio for hvplot?


rsignell-usgs:
I thought from https://github.com/ioam/holoviews/issues/30#issuecomment-104613952 that this might work:
```python
import xarray as xr
import hvplot.xarray
url = 'https://rsignell.s3.amazonaws.com/NLCD/nlcd_2011_landcover_2011_edition_2014_10_10.tif'
da = xr.open_rasterio(url, chunks={'x':2056, 'y':2056})

%%opts Image [width=700 height=500 colorbar=True aspect='equal'] (cmap='magma') 
da[0, 50000:51000, 80000:81000].hvplot()
```
but I guess `aspect` is not a thing.

Is it possible to use `Raster` with hvplot?

I couldn't figure that out either.   😿

philippjfr:
Currently bokeh does not afford us sufficient control over the inner plot dimensions to be able to offer aspect options. There is however a current PR in progress in bokeh that will allow us to take full control over the inner dimensions of a plot and therefore implement reasonable aspect control. I'd suggest checking back once bokeh 1.0 has been released and we've had a chance to expose the appropriate options in holoviews.

rsignell-usgs:
Okay, @philippjfr , thanks.   Is there a way to use `Raster` with hvplot?

rsignell-usgs:
@philippjfr Is this the bokeh PR? https://github.com/bokeh/bokeh/pull/8085#issuecomment-427104927

philippjfr:
That's the one, it's on hold until bokeh 1.1 unfortunately as it's a huge chunk of work and didn't quite make the cut-off for 1.0, which should happen very soon.

tomchor:
@philippjfr I was also trying to set the aspect of a plot and came across this issue. It seems to me that that problem was [fixed on Bokeh's end](https://github.com/bokeh/bokeh/pull/8085) and that it's available now in version 1.1. So I'm checking back to see if this is implemented on `hvplot`'s end.

philippjfr:
It isn't yet but I was planning on spending next weekend on making various updates to hvplot.

PR #268, opened 2019-08-01
Allow aspect and data_aspect in opts and set data_aspect to 1 when geo=True

jsignell:
Still working on tests.

Screen Shot 2019-08-01 at 4 08 05 PM

Closes: #192, #75


jsignell:
Note that this doesn't work with `datashade` and `rasterize` since those require that a height and width be known ahead of time.

philippjfr:
>Note that this doesn't work with datashade and rasterize since those require that a height and width be known ahead of time.

That's not really too important the only difference that makes is that the initial render has roughly the correct resolution, but once displayed it triggers a re-render anyway. So I'd just fall back to the default width/height or come up with your own default for this case.

PR #273, opened 2019-08-05
If datashade=True then hover should be false by default

### LINKED ISSUE #204: hover broken for datashade
image

```
import pandas as pd
import hvplot.pandas

tst = pd.DataFrame([0, 1, 2])

tst.hvplot('index', 0, datashade=True)
```

jsignell:
Closes #204 

This is kind of a half-solution. It just turns off the hover when datashade is true if the user doesn't specifically ask for it.


philippjfr:
Needs to be rebased.

Issue #226, opened 2019-06-24
c argument in scatter does not work with datashader


joshua-gould:
The following 2 plots are identical, even though in the second plot c='yr' is specified.
```
import holoviews as hv
import hvplot.pandas
from bokeh.sampledata.autompg import autompg
hv.extension('bokeh')
autompg.hvplot.scatter(x='mpg', y='yr',  datashade=True, dynspread=True)
autompg.hvplot.scatter(x='mpg', y='yr', c='yr', datashade=True, dynspread=True)
```

jsignell:
What would you expect the outcome to be? Should the aggregation method change from counts to mean of `c` values within bin when `c` is set?

joshua-gould:
I was expecting the plots to look nearly identical which would necessitate the default aggregation method to be the mean of c values within a bin. Thanks.

jsignell:
Hmm I think the issue is actually that we expect `x` `y` and `c` to be different columns.

For example this works as expected:

```python
autompg.hvplot.scatter(x='mpg', y='yr', c='cyl', datashade=True, dynspread=True)
```

<img width="822" alt="Screen Shot 2019-08-05 at 10 57 39 AM" src="https://user-images.githubusercontent.com/4806877/62473954-dd167480-b76f-11e9-98b0-8f0535251024.png">

philippjfr:
I've partially fixed this already, that's why you're only seeing an error if the columns are the same. Before https://github.com/pyviz/hvplot/pull/257 this wasn't working at all.

PR #274, opened 2019-08-05
Allow datashade/rasterize even when c is same as x or y

jsignell:
Closes: #226 

Note that this PR also fixes some instances of hvplot altering the input data.

```python
import holoviews as hv
import hvplot.pandas
from bokeh.sampledata.autompg import autompg
hv.extension('bokeh')
autompg.hvplot.scatter(x='mpg', y='yr', c='yr', datashade=True, dynspread=True)
autompg.hvplot.scatter(x='mpg', y='yr', c='yr', rasterize=True, y_sampling=1, x_sampling=1)
```

Screen Shot 2019-08-05 at 1 42 35 PM

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -644,8 +644,9 @@ class HoloViewsConverter(object):
         self.use_dask = use_dask
         self.indexes = indexes
         if isinstance(by, (np.ndarray, pd.Series)):
-            self.data['by'] = by
-            self.by = ['by']
+            self.data = self.data.assign(_by=by)
+            self.by = ['_by']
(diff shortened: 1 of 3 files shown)


philippjfr:
LGTM

philippjfr:
Oh, it's still WIP, nvm.

jsignell:
Yeah now it's ready. I was still cleaning up logic and adding tests. The only dangling bit now is that the _color field is included in the hover. Not sure how to turn that off.

jsignell:
Ok if I merge?

PR #279, opened 2019-08-07
Made hvplot API dynamic by accepting parameter instances

philippjfr:
This PR makes it possible to build GUIs by passing parameters instances directly into the hvPlot methods. Combined with panel this makes it easy to build complex GUIs very easily.

```python
import param
import panel as pn
import hvplot.pandas

from bokeh.sampledata.iris import flowers

cols = list(flowers.columns[:-1])
x = pn.widgets.Select(name='x', value='sepal_length', options=cols)
y = pn.widgets.Select(name='y', value='sepal_width', options=cols)
kind = pn.widgets.Select(name='kind', value='scatter', options=['bivariate', 'scatter'])
by_species = pn.widgets.Checkbox(name='By species')
color = pn.widgets.ColorPicker(value='#ff0000')

@pn.depends(by_species.param.value, color.param.value)
def by_species_fn(by_species, color):
    return 'species' if by_species else color

plot = flowers.hvplot(x=x.param.value, y=y.param.value, kind=kind.param.value, c=by_species_fn)

pn.Row(pn.WidgetBox(x, y, kind, color, by_species), plot)
```

iris_explorer

This examples demonstrates how both parameters and functions which depend on some parameter(s) can be used to dynamically update a plot. In this case the ``by_species_fn`` returns either the 'species' column or the `ColorPicker` depending on whether the checkbox is ticked, making it to specify even complex behaviors.

- [ ] Clean up code
- [ ] Add docs
- [ ] Add tests

### DIFF
--- a/hvplot/__init__.py
+++ b/hvplot/__init__.py
@@ -7,6 +7,7 @@ import textwrap
 import param
 import numpy as _np
 import pandas as _pd
+import panel as _pn
 import holoviews as _hv
 
 from holoviews import Store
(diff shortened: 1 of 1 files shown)


jsignell:
Hmm might be easier to open a new PR for this given the reorg.

Issue #177, opened 2019-02-26
datashader canvas does not use width and ylim options


marcbernot:
When plotting this (both in notebook and lab),
`y = np.random.randn(100000)
pd.DataFrame(y).reset_index().hvplot('index','0',datashade=True,width=3000,ylim=(-1,1))
`
I would expect to obtain the same result as
`datashade(hv.Curve(y),y_range=(-1,1),width=3000).options(width=3000)`
though it renders like 
`datashade(hv.Curve(y)).options(width=3000,ylim=(-1,1))`
i.e. a stretched and cropped version of the rasterization obtained on the smaller default width.

rtmatx:
I tried this on my system and I can get it working with the first option where you mentioned `width` inside the `hvplot` command.

marcbernot:
Ok, that is strange. Just to be sure I updated everything so I have holoviews=1.11.3 datashader=0.6.9 and hvplot=0.4.0 from conda-forge and I don't get the expected rendering with hvplot. I have the bug on windows7 with chrome and osX with safari. What is your os and browser and versions of the libraries?

rtmatx:
I am using chrome on osx

marcbernot:
Well with width=900 you'll won't notice the difference, try with width=3000

rtmatx:
I actually did, since I couldn't capture the entire photo, I didnt use that for showing here

marcbernot:
Well I have this on chrome osX
<img width="731" alt="hvplot" src="https://user-images.githubusercontent.com/8888261/53524435-a90f9700-3adf-11e9-8f47-b7cb7919c5dd.png">
What could be different? I use this set of imports then the three lines, that's it:
```
import numpy as np
import pandas as pd
import hvplot.pandas
import holoviews as hv
hv.extension('bokeh')
from holoviews.operation.datashader import datashade
```

jsignell:
I can't reproduce:

<img width="1028" alt="Screen Shot 2019-08-05 at 3 44 43 PM" src="https://user-images.githubusercontent.com/4806877/62490696-f92e0c00-b797-11e9-8c4d-89450f4263ef.png">

Please reopen if you still see this bug on latest versions.

marcbernot:
Well, your screenshot shows that you reproduced the bug. The two images should be the same and they are not. 
Notice also that if you slightly zoom in and out of the upper plot, it will rerasterize and produce the expected figure.

jsignell:
[...]

PR #280, opened 2019-08-07
Fixing lim should be used as range when datashading

jsignell:
The axis zoom was happening after datashading when it should have been happening before.

Closes: #177 

```python
import numpy as np
import pandas as pd
import hvplot.pandas

pd.DataFrame(y).hvplot(datashade=True, xlim=(0, 20000))
```

Before this PR:

Screen Shot 2019-08-07 at 9 49 14 AM

After this PR:

Screen Shot 2019-08-07 at 9 48 45 AM

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -921,6 +921,10 @@ class HoloViewsConverter(object):
             opts['x_sampling'] = self.x_sampling
         if self.y_sampling:
             opts['y_sampling'] = self.y_sampling
+        if self._plot_opts.get('xlim') is not None:
+            opts['x_range'] = self._plot_opts['xlim']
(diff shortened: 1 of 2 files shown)


jsignell:
@philippjfr not quite sure how to test this.

jbednar:
Good catch!

philippjfr:
>@philippjfr not quite sure how to test this.

Use dynamic=False, and then check the range matches I guess:

```
img = pd.DataFrame(np.random.randn(100).cumsum()).hvplot(datashade=True, xlim=(0, 20000), dynamic=False)
assert img.range(0) == (0, 20000)
```

philippjfr:
This is having some bad results for quadmesh aggregations, but it seems like this is a bug in datashader and/or holoviews.

PR #281, opened 2019-08-07
Run notebooks on travis

### LINKED ISSUE #283: Incorrect param warning on bar plot + width
There is a new param warning that feels incorrect:

`WARNING:param.BarPlot02810: BarPlot width option is deprecated use 'bar_width' instead.`

Screen Shot 2019-08-08 at 9 57 27 AM

jsignell:
- fixes kde
 - adds notebook tests on CI
 - gets rid of dead code
 - adds work-around for #282
 - closes #283


hvplot/converter.py
```diff
@@ -1172,6 +1172,10 @@ def _category_plot(self, element, x, y, data):
                 .relabel(**self._relabel).opts(**opts))
 
     def bar(self, x, y, data=None):
+        # As of bokeh 1.3.0 and holoviews 1.12.4, width is valid on style but throws
+        # a param warning. Use bar_width to control width of bars.
+        self._style_opts.pop('width', None)
```
philippjfr: Why is width in the style opts?

jsignell: It is a valid bokeh style, so when the kwds get passed through the validator it is added to style_opts.

philippjfr: But surely it should get captured and dropped during plot option handling before the style opts pick it up?

jsignell: Hmm I think with the opts, maybe the plot_opts should be popped before passing all the kwds to style validator.

philippjfr: Right, only unprocessed kwds should ever making through to the style validator.

Issue #223, opened 2019-06-19
hvplot.networkx.draw() coloring does not respect original node order


stuarteberg:
When using `nx.draw()`, you can customize the color of each node by passing a list of `colors`, in the same order as `G.nodes`.  But the equivalent `hvplot` function doesn't preserve the order correctly.

For example, here's a simple graph using randomly ordered nodes, and an alternating colortable.

```python
import numpy as np
import networkx as nx

# Create nodes (1-10) in random order
nodes = np.random.choice(np.arange(1,11), 10, False)
edges = [*zip(nodes[:-1], nodes[1:])]

g = nx.Graph()
g.add_nodes_from(nodes)
g.add_edges_from(edges)

# Alternating colortable
colors = 5*['pink', 'skyblue']
pos = nx.kamada_kawai_layout(g)

print(f"nodes: {g.nodes}")
print(f"edges: {g.edges}")
print(f"colors: { {n:c for n,c in zip(g.nodes, colors)} }")
```

```
nodes: [3, 5, 2, 10, 8, 4, 9, 7, 1, 6]
edges: [(3, 5), (5, 2), (2, 10), (10, 8), (8, 4), (4, 9), (9, 7), (7, 1), (1, 6)]
colors: {3: 'pink', 5: 'skyblue', 2: 'pink', 10: 'skyblue', 8: 'pink', 4: 'skyblue', 9: 'pink', 7: 'skyblue', 1: 'pink', 6: 'skyblue'}
```

Plain `nx.draw()` displays the colors faithfully:

```python
# Display with matplotlib
nx.draw(g, node_color=colors, pos=pos, with_labels=True)
```

![image](https://user-images.githubusercontent.com/1655821/59793805-76afc280-92a5-11e9-92f6-21a06b7285e5.png)

But `hvnx.draw()` mixes up the colortable order.  The colors in this graph do not alternate correctly.  Instead, the colortable was applied to the nodes in *sorted* order, not in the order of `g.nodes`.  (Notice that all even nodes below are colored blue, and odd nodes are pink.)

```python
# Display with hvplot
import hvplot.networkx as hvnx
hvnx.draw(g, node_color=colors, pos=pos, with_labels=True)
```

![image](https://user-images.githubusercontent.com/1655821/59793832-8b8c5600-92a5-11e9-9b36-280ea0d7e41c.png)

(For now, as a workaround, one can pre-sort the nodes before inserting them into the graph, which fixes the issue.)

<details>

[...]

PR #284, opened 2019-08-08
Fixing color order should match node order

jsignell:
Closes: #223 

Using example from issue:

Screen Shot 2019-08-08 at 11 30 06 AM

Not sure how to test.

### DIFF
--- a/hvplot/networkx.py
+++ b/hvplot/networkx.py
@@ -75,7 +75,7 @@ def _from_networkx(G, positions, nodes=None, cls=Graph, **kwargs):
     node_columns = defaultdict(list)
 
     # Unpack node positions
-    for idx, pos in sorted(positions.items()):
+    for idx, pos in positions.items():
         node = G.nodes.get(idx)
         if node is None:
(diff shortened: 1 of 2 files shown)


hvplot/networkx.py
```diff
@@ -75,7 +75,7 @@ def _from_networkx(G, positions, nodes=None, cls=Graph, **kwargs):
     node_columns = defaultdict(list)
 
     # Unpack node positions
-    for idx, pos in sorted(positions.items()):
+    for idx, pos in positions.items():
```
philippjfr: Wonder why I sorted here in the first place.

jsignell: Hard to tell why since there aren't any tests :(

philippjfr:
>Not sure how to test.

Check the order of this maybe?

```
plot.nodes.dimension_values(2)
```

hvplot/networkx.py
philippjfr: 😞

Issue #278, opened 2019-08-06
Derived datetime for tabular formats


jsignell:
This would look a lot like the xarray equivalent implemented in #263.

PR #286, opened 2019-08-08
Derived datetime for tabular data

jsignell:
Closes: #278

I was experimenting with this simulated dataset:

```python
import numpy as np
import pandas as pd
import hvplot.pandas

time_df = pd.DataFrame({
    'time': pd.date_range('1/1/2000', periods=5*24, freq='1H', tz='UTC'),
    'temp': np.sin(np.linspace(0, 5*2*np.pi, 5*24)).cumsum()})

time_df.hvplot(x='time.hour', by='time.dayofweek')
time_df.hvplot.heatmap(x='time.hour', y='time.day', C='temp')
```

Screen Shot 2019-08-08 at 3 15 00 PM

Screen Shot 2019-08-08 at 4 25 56 PM

### DIFF
--- a/hvplot/__init__.py
+++ b/hvplot/__init__.py
@@ -710,7 +710,7 @@ class hvPlot(object):
         """
         return self(x, y, c=c, kind='polygons', **kwds)
 
-    def paths(self, **kwds):
+    def paths(self, x=None, y=None, c=None, **kwds):
         """
         LineString and LineRing plot for geopandas dataframes.
 
@@ -724,7 +724,7 @@ class hvPlot(object):
         obj : HoloViews object
             The HoloViews representation of the plot.
         """
-        return self(x=None, y=None, kind='paths', **kwds)
+        return self(x, y, c=c, kind='paths', **kwds)
 
     def labels(self, x=None, y=None, text=None, **kwds):
         """
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -1108,10 +1108,10 @@ class HoloViewsConverter(object):
             data = data.reset_index()
 
         # calculate any derived time
-        dims = []
-        for dim in [x, y, self.by, self.hover_cols]:
-            if dim is not None:
-                dims.extend(dim if isinstance(dim, list) else [dim])
+        dimensions = []
+        for dimension in [x, y, self.by, self.hover_cols]:
+            if col is not None:
+                dimension.extend(dimension if isinstance(col, list) else [dimension])
 
         not_found = [dim for dim in dims if dim not in self.variables]
         _, data = process_derived_datetime_pandas(data, not_found, self.indexes)
@@ -1444,12 +1444,12 @@ class HoloViewsConverter(object):
                                       c not in data.columns):
                 data = data.reset_index()
             # calculate any derived time
-            dims = []
-            for dim in [x, y, self.by, self.hover_cols]:
-                if dim is not None:
-                    dims.extend(dim if isinstance(dim, list) else [dim])
+            dimensions = []
+            for dimension in [x, y, self.by, self.hover_cols]:
+                if dimension is not None:
+                    dimensions.extend(dimension if isinstance(dimension, list) else [dimension])
 
-            not_found = [dim for dim in dims if dim not in self.variables]
+            not_found = [dim for dim in dimensions if dim not in self.variables]
             _, data = process_derived_datetime_pandas(data, not_found, self.indexes)
 
         return data, x, y, z
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -1109,11 +1109,11 @@ class HoloViewsConverter(object):
 
         # calculate any derived time
         dimensions = []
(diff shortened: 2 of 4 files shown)


hvplot/__init__.py
```diff
@@ -710,7 +710,7 @@ def polygons(self, x=None, y=None, c=None, **kwds):
         """
         return self(x, y, c=c, kind='polygons', **kwds)
 
-    def paths(self, **kwds):
+    def paths(self, x=None, y=None, c=None, **kwds):
```
jsignell: This was making it impossible to use paths with non-geopandas

hvplot/converter.py
```diff
@@ -1347,26 +1362,28 @@ def dataset(self, x=None, y=None, data=None):
...
         hmap = HeatMap(data, [x, y], z, **self._relabel)
         if 'reduce_function' in self.kwds:
             hmap = hmap.aggregate(function=self.kwds['reduce_function'])
         return hmap.redim(**redim).opts(**opts)
 
     def hexbin(self, x, y, data=None):
-        data = self.data if data is None else data
-        if not x: x = data.columns[0]
-        if not y: y = data.columns[1]
+        self.use_index = False
```
philippjfr: Should we handle these in some more general way?

philippjfr:
One comment which is more of a question, otherwise looks good.

hvplot/converter.py
jsignell: I actually think that we should let them use index, but I didn't want to change prior behavior.

jsignell: I'd be happy to change it though if you don't think it'll be too painful

philippjfr: Can you explain the consequences?

jsignell: instead of x defaulting to the first column, it'd default to the index , and then y instead of defaulting to the second columns would default to the first.

philippjfr: Let's leave that for another time, I'm not totally sure we should do that for 2D plots.

jbednar: Doesn't Pandas .plot() do it?

philippjfr: Pandas plot does not have these plot types.

jbednar: I mean, doesn't Pandas .plot() use the index in that way, in general?  I think hvPlot should follow what Pandas does, even for new plot types, to be predictable and unsurprising.

jsignell:
I just wrote a little comparer for matplotlib and hvplot

```python
import pandas as pd
import hvplot.pandas
import matplotlib.pyplot as plt
%matplotlib inline

df = pd.DataFrame([[1,2],[3,4], [5,6],[7,8]] , columns=['A', 'B'])
method_list = [func for func in dir(df.plot) if not func.startswith('_')]

def compare_mpl_hvplot(kind):
    try:
        fig = plt.Figure(figsize=(6, 5))
        ax = fig.subplots()
        df.plot(kind=kind, ax=ax);
    except Exception as err:
        fig = pn.Column(pn.Spacer(height=30), pn.Pane(err.__repr__(), width=400))
    
    try:
        hv_fig = df.hvplot(kind=kind, height=300, width=400, legend='top_left')
    except Exception as err:
        hv_fig = pn.Pane(err.__repr__(), width=400)
    
    return pn.Column(
        f'### {kind}',
        pn.Row(fig, pn.Column(pn.Spacer(height=30), hv_fig))
    )

pn.interact(compare_mpl_hvplot, kind=method_list)
```

hvplot/converter.py
jsignell: These types of plots require explicit x and y in pandas.plot

jsignell:
Similarly:

```python 
import panel as pn
import xarray as xr

import hvplot.xarray
import matplotlib.pyplot as plt

%matplotlib inline

ds = xr.tutorial.open_dataset('air_temperature').load()
da = ds.air.isel(time=500) - 273

method_list = [func for func in dir(da.plot) if not func.startswith('_')]
method_list += [func for func in dir(da.hvplot) if not func.startswith('_') and func not in method_list]

def compare_mpl_hvplot(kind):
    try:
        if kind == 'quadmesh':
            kind = 'pcolormesh'
        if kind == 'image':
            kind = 'imshow'
        fig = plt.Figure(figsize=(6, 5))
        ax = fig.subplots()
        getattr(da.plot, kind)(ax=ax);
    except Exception as err:
        fig = pn.Column(pn.Spacer(height=30), pn.Pane(err.__repr__(), width=400))
    
    try:
        if kind == 'pcolormesh':
            kind = 'quadmesh'
        if kind == 'imshow':
            kind = 'image'
        hv_fig = da.hvplot(kind=kind, height=300, width=400, legend='top_left')
    except Exception as err:
        hv_fig = pn.Pane(err.__repr__(), width=400)
    
    return pn.Column(
        f'### {kind}',
        pn.Row(fig, pn.Column(pn.Spacer(height=30), hv_fig))
    )

pn.interact(compare_mpl_hvplot, kind=method_list)
```

jsignell:
I'm not quite sure what's going on with that networkx notebook.

PR #291, opened 2019-08-21
Load Intake Dataset partition or Dask.sample, resolves #290

### LINKED ISSUE #290: Proposal to load subset of data for Intake Datasets
### Issue:
hvPlot by default loads all partitions into memory, see [process_intake()](https://github.com/pyviz/hvplot/blob/c8d88238e282c108705edaed6be73bb4a716efd7/hvplot/util.py#L245).  This is an issue for very large datasets.

### Proposed solution:
Two extra parameters (in addition to `use_dask`) sent to `hvPlot.process_intake()`: read_partition, dask_sample.

* dask_sample: value would be between 0 and 1.
* read_partition: would indicate which partition to read.

The contents of process_intake can be updated to:
```python
    # todo: argument validation
    if dask_sample:
        data = data.to_dask().sample(frac=float(dask_sample), replace=True) 
    elif use_dask:
        data = data.to_dask()
    elif read_partition:
        data = data.read_partition(read_partition)
    else:
        data = data.read()
    return data
```

### Related
* Parent issue from Intake https://github.com/intake/intake/issues/409
* intake select a subset  https://github.com/pyviz/hvplot/issues/72

talebzeghmi:


maximlt:
Hi @talebzeghmi, I'm so sorry no one ever replied to you on this PR!

A few things have changed since then:
- Intake's integration isn't as much of a priority as it used to be
- Requirements to add new arguments to hvPlot are stricter (it has too many already)

Because of this, I do not favor merging this PR and will close it.

PR #306, opened 2019-09-13
Geopandas should set geo=True, also clabel/zlabel

jsignell:
Geopandas requires that geoviews be in the env, so it should set `geo=True` I think. Also added some tests around clabel and aliased to zlabel.

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -137,8 +137,8 @@ class HoloViewsConverter(object):
         e.g. '%.3f', and bokeh TickFormatter)
     xlabel/ylabel/clabel (default=None): str
         Axis labels for the x-axis, y-axis, and colorbar
-    xlim/ylim (default=None): tuple or list
-        Plot limits of the x- and y-axis
+    xlim/ylim/clim (default=None): tuple or list
+        Plot limits of the x- and y-axis, and colorbar
     xticks/yticks (default=None): int or list
         Ticks along x- and y-axis specified as an integer, list of
         ticks positions, or list of tuples of the tick positions and labels
@@ -280,7 +280,7 @@ class HoloViewsConverter(object):
                  backlog=1000, persist=False, use_dask=False,
                  crs=None, fields={}, groupby=None, dynamic=True,
                  grid=None, legend=None, rot=None, title=None,
-                 xlim=None, ylim=None, clim=None,
+                 xlim=None, ylim=None, clim=None, clabel=None,
                  logx=None, logy=None, loglog=None, hover=None,
                  subplots=False, label=None, invert=False,
                  stacked=False, colorbar=None, fontsize=None,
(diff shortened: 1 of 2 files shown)


philippjfr:
>Geopandas requires that geoviews be in the env, so it should set geo=True I think

I'm still a -1 on this.

jbednar:
I'm not sure. I think defaults should generally follow the principle of least surprise, but it's not clear how to do so in this case. On the one hand, someone using Geopandas may reasonably expect that values given in lat,lon are automatically taken as geo coordinates, and people using non lat/lon geo coordinates would need to declare the crs in most cases anyway, so it seems like defaulting to `geo=True` would make things "just work" in many cases that they otherwise wouldn't. On the other hand, using GeoPandas just means that there is a shape column defined, not that it's actually being used in this plot, right?  Would having `geo` default to true cause problems for people not actually plotting lat,lon in that particular plot?  I'm not sure of all that happens when  you set `geo=True` right now.

jsignell:
FWIW this is what surprised me:

Screen Shot 2019-09-12 at 1 45 20 PM

I am not sure what would happen if crs were not lat lon.

jbednar:
Shouldn't `tiles=True` also make `geo=True` by default, even with regular Pandas?  Seems independent of what's being discussed here.  I don't see any docs for `tiles=True`, but I assume it underlays some default map tile set?

jsignell:
> Shouldn't tiles=True also make geo=True 

But you don't need geoviews anymore for tiles, so this isn't as clear.

jbednar:
Ah, good point.

philippjfr:
>But you don't need geoviews anymore for tiles, so this isn't as clear.

I tentatively vote for enabling geo=True for both but suppressing the warning when tiles=True but geoviews is not available.

philippjfr:
>I thought it was a declaration that the x and y values are in geo coordinates, which is information that can be used by GeoViews but is also meaningful without it.

Hmm, I guess but not really. The datashader lat/lon conversion works for points and that's about it, so it's really a **very** partial solution.

jsignell:
> I tentatively vote for enabling geo=True for both but suppressing the warning when tiles=True but geoviews is not available.

I think this would work reasonably well. We should probably be thinking about this from the user's perspective and the user is probably not super aware of whether they have geoviews installed. So I think anything we can do to make it work as expected more often is a good idea.

jbednar:
> Overlaying lat/lon data on top of tiles wouldn't work.

I assume you mean, it wouldn't work without adding a bit of code to hvplot to use the Datashader approach?  Using lnglat_to_meters seems like it would work for Points and probably for shapes encoded with NaN separators. My guess is that would cover a lot of ground (no pun intended) without having to pull in GeoViews.  But I'll leave that to your judgment.

jsignell:
Ok what should be done here? Should I just revert the `geo=True` for geopandas? Should I make `geo=True` have more semantic meaning (don't raise error for missing geoviews, just try to handle whatever is passed with the libraries available)?

jsignell:
Closing this since it's pretty stale at this point

PR #307, opened 2019-09-13
Get rid of methods that are invalid for certain objects

jsignell:
The goal is to only show plot types that are actually valid for particular objects.


jsignell:
This is not going to work because it deletes on the class.

PR #308, opened 2019-09-14
Allow the source Dataset to flow through to Histogram elements

jonmmease:
When used with HoloViews https://github.com/pyviz/holoviews/pull/3919, this PR ensures that a resulting `Histogram` element has a `.dataset` property containing all of the dimensions in the original dataset.  No change in behavior for prior HoloViews versions.


jsignell:
LGTM. Thanks!

PR #316, opened 2019-09-24
Make glasbey_category10 the default categorical colormap

### LINKED ISSUE #311: default cmap for categorical should change depending on # of categories
The default cmap for categorical is Category10 so when more than 10 categories are in a dataset colors are recycled. It would be nice if the cmap would change to 'Category20' for example, when 15 categories are shown.

jsignell:
Closes #311


jsignell:
I'm pretty confident the failure is with datashader rather than this change.

PR #318, opened 2019-09-24
Fix stats plots ordering

### LINKED ISSUE #317: Sorting non-categorical axis by
Not sure whether this is an an hvplot issue or a holoviews issue, but there doesn't seem to be any sorting for non-cat `by` at least for violin plots.

```python
import xarray as xr
import hvplot.xarrray

ds.hvplot.violin('air', by='lat', c='lat', cmap='fire')
```

Screen Shot 2019-09-24 at 3 31 33 PM

Compared to the [docs](https://hvplot.pyviz.org/user_guide/Gridded_Data.html): 

Screen Shot 2019-09-24 at 3 33 36 PM

jsignell:
Closes #317 

Screen Shot 2019-09-24 at 5 10 14 PM

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -1107,29 +1107,23 @@ class HoloViewsConverter(object):
         return chart.redim(**self._redim).opts(opts)
 
     def _process_chart_x(self, data, x):
+        """This should happen before _process_chart_y"""
         if x is False:
             return None
+
         x = x or self.x
         if x is None:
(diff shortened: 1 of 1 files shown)


philippjfr:
Seems like an okay fix, but really we should avoid sorting on columns that never end up being used.

philippjfr:
Nevermind, I realize that's what it's doing now.

PR #335, opened 2019-10-12
Added absolute_import for py2 compat

philippjfr:


jsignell:
Yikes. Thanks @philippjfr

PR #343, opened 2019-10-17
Fix multi z

### LINKED ISSUE #336: Missing behavior for zs
As reported in https://stackoverflow.com/questions/58350586/generate-xarray-dataset-image-subplots, passing a list of `zs` appears not to be fully implemented in current hvplot master:

```
import xarray as xr, numpy as np, hvplot.xarray
ds = xr.Dataset({
    'temp': (('lon', 'lat'), 15 + 8 * np.random.randn(2, 2)),
    'precip': (('lon', 'lat'), 10 * np.random.rand(2, 2))},
    coords={'lon': [-99.83, -99.32],'lat': [42.25, 42.21]})
ds.hvplot(x="lat", y="lon", z=['temp', 'precip'])
```

```
---------------------------------------------------------------------------
UnboundLocalError                         Traceback (most recent call last)
<ipython-input-18-ca3236b8574f> in <module>
      4     'precip': (('lon', 'lat'), 10 * np.random.rand(2, 2))},
      5     coords={'lon': [-99.83, -99.32],'lat': [42.25, 42.21]})
----> 6 ds.hvplot(x="lat", y="lon", z=['temp', 'precip'])

~/hvplot/hvplot/plotting/core.py in __call__(self, x, y, kind, **kwds)
     72                 return pn.panel(plot, **panel_dict)
     73 
---> 74         return self._get_converter(x, y, kind, **kwds)(kind, x, y)
     75 
     76     def _get_converter(self, x=None, y=None, kind=None, **kwds):

~/hvplot/hvplot/converter.py in __call__(self, kind, x, y)
    938 
    939         if not (self.datashade or self.rasterize):
--> 940             return self._apply_layers(obj)
    941 
    942         try:

UnboundLocalError: local variable 'obj' referenced before assignment
```

Plotting works fine if a single item (precip or temp here) is passed for `z`, but a multi-item list of zs fails. It _appears_ to have been intended to be implemented, given the handling for `len(zs) > 1` in 
[`hvplot.converter.HoloViewsConverter.__call__()`](https://github.com/pyviz/hvplot/blob/master/hvplot/converter.py#L913), but that code defines `dataset` without defining `obj`, causing a problem when all the various branches eventually lead to something processing `obj`.  A single-item list of zs also fails, for a different reason, but that might not have been expected to be implemented.  The original poster was hoping to use multiple `zs` as subplots, which seems reasonable given how multiple `ys` work, though without subplots it's not clear how the multiple zs should be handled.  (With `ys`, the default is  to overlay them, but overlaying multiple colormaps isn't particularly meaningful, so a default of getting a widget seems more useful in that case.)  In any case, this case appears to be _partially_ implemented, and it seems like it should either be fully implemented or explicitly made into an error.

jsignell:
closes #336

Screen Shot 2019-10-17 at 1 03 33 PM

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -923,12 +923,13 @@ class HoloViewsConverter(object):
                 else:
                     obj = dataset.map(lambda ds: method(x, y, data=ds.data), Dataset)
             elif len(zs) > 1:
+                dimensions = [Dimension(self.group_label, values=zs)]
                 if self.dynamic:
                     obj = DynamicMap(lambda z: method(x, y, z, data=dataset.data),
-                                     kdims=[Dimension(self.group_label, values=zs)])
+                                     kdims=dimensions)
                 else:
                     obj = HoloMap({z: method(x, y, z, data=dataset.data) for z in zs},
-                                   kdims=[self.group_label])
+                                  kdims=dimensions)
             else:
                 obj = method(x, y, data=dataset.data)
             if self.grid:
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -30,7 +30,7 @@ from holoviews.util.transform import dim
 from pandas import DatetimeIndex, MultiIndex
 
 from .util import (
-    is_tabular, is_series, is_dask, is_intake, is_streamz, is_xarray,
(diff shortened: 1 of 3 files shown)


jbednar:
Thanks!

PR #345, opened 2019-10-17
Allow tiles to be objects as well as strings

jsignell:
I was surprised that this wasn't already allowed.


jsignell:
Which doesn't necessarily mean it's a good idea.

philippjfr:
I think it's a good idea, oversight on my part. I'd definitely expect this to work.

PR #347, opened 2019-10-17
Exposing plot at top level

jsignell:
```python
import numpy as np
import pandas as pd

data = np.random.normal(size=[50, 2])
df = pd.DataFrame(data, columns=['x', 'y'])

df.plot(x='x', y='y')
```

### DIFF
--- a/hvplot/__init__.py
+++ b/hvplot/__init__.py
@@ -14,7 +14,7 @@ from .util import get_ipy
 from .utilities import save, show # noqa
 from .plotting import (hvPlot, hvPlotTabular,  # noqa
                        andrews_curves, lag_plot,
-                       parallel_coordinates, scatter_matrix)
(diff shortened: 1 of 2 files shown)


jbednar:
How does that example depend on having plot at the top level?

jsignell:
This is from the gitter convo

jsignell:
There are two ways to implement backends for `pd.options.plotting.backend `: entrypoints or exposing a top level plot function in the module.

jbednar:
To summarize the conversation from gitter for posterity, the issue is that `pd.options.plotting.backend = 'holoviews'` currently works, via the entry point mechanism, but `pd.options.plotting.backend = 'hvplot'` does not.  With this PR, `pd.options.plotting.backend = 'hvplot'` now works, via the top-level plot function method.

Issue #363, opened 2019-11-06
Typing a string to input as crs/projection instead of a ccrs object similar to cmap/colors


ahuang11:
Instead of having to import `cartopy.crs as ccrs` and typing `projection=ccrs.Robinson()`, what if it could simply be `projection='Robinson'`, just like how you can enter `cmap='jet'` instead of `cmap=plt.cm.jet`?  Another example is the datetime strings and hover tool. Also, if so, not sure if it'd be good to have in geoviews or hvplot.

Implementation I think is straightforward
```
if isinstance(projection, str):
    projection = getattr(ccrs, projection)()
```

philippjfr:
I wouldn't object to this, the logic should be added to `hvplot.util.process_crs`. Cc @jsignell @jbednar

jbednar:
I have no objection, though it would need to be a bit more robust than above, i.e. to try to import `cartopy.crs` only if a string is supplied, and even then failing gracefully with a nice message like "Referring to projections by string name requires installing cartopy", if cartopy.crs is not importable.

jsignell:
You can already pass a string to crs, but the string ins't an a cartopy class. It's a proj4 string - something like: 'EPSG:32618'. I don't think being able to pass the name of a cartopy class has much value since many require arguments an there would be no way to pass those.

jsignell:
So I guess I'm -1 on this suggestion

ahuang11:
I guess I am unfamiliar with proj4 strings. In most cases, I am simply projecting to GOOGLE_MERCATOR (for tiles) or Robinson (for global datasets) which require no additional arguments. If the user does want to provide additional arguments, it's similar to using `tools=[HoverTool()]`

jbednar:
Seems like it's down to the details of an implementation, so I'd have to see that before deciding whether it's reliable enough to do this. E.g. one would have to first check proj4, have an exception, then try a Cartopy projection class name. Not crazy, but needs to not be confusing in practice.

maximlt:
I'm also -1. GeoPandas uses `pyproj` under the hood to parse the value passed to the `crs` parameter when construction a GeoDataframe, this is what it currently accepts:

```
        PROJ string
        Dictionary of PROJ parameters
        PROJ keyword arguments for parameters
        JSON string with PROJ parameters
        CRS WKT string
        An authority string [i.e. ‘epsg:4326’]
        An EPSG integer code [i.e. 4326]
[...]

PR #372, opened 2019-11-13
Support string projection kwd

ahuang11:
As jsignell suggested, string crs probably isn't too useful since you need to pass arguments in most cases if it's not degree latitude/longitude crs so I didn't support that, but I think being able to use a simple string for `projection` is useful.
image

https://github.com/holoviz/hvplot/issues/363

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -337,6 +337,22 @@ class HoloViewsConverter(object):
                     "following plot types: %r" % (self.kind, self._geo_types))
             from cartopy import crs as ccrs
             from geoviews.util import project_extents
+
+            if isinstance(projection, basestring):
+                all_crs = [proj for proj in dir(ccrs) if
+                           callable(getattr(ccrs, proj)) and
+                           proj not in ['ABCMeta', 'CRS'] and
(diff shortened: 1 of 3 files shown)


hvplot/converter.py
```diff
@@ -337,6 +337,22 @@ def __init__(self, data, x, y, kind=None, by=None, use_index=True,
...
+            if isinstance(projection, basestring):
+                all_crs = [proj for proj in dir(ccrs) if
+                           callable(getattr(ccrs, proj)) and
+                           proj not in ['ABCMeta', 'CRS'] and
+                           proj[0].isupper() or
+                           proj == 'GOOGLE_MERCATOR']
+                if projection in all_crs and projection != 'GOOGLE_MERCATOR':
+                    projection = getattr(ccrs, projection)()
+                elif projection == 'GOOGLE_MERCATOR':
+                    projection = getattr(ccrs, projection)
```
jbednar: I don't quite know what's happening with GOOGLE_MERCATOR here.  Could it maybe be handled by checking if it's callable, and only calling callable things?

PR #382, opened 2019-11-25
More notebooks for the reference gallery

jsignell:
The goal is to get at least one examples of each type of plot before merging this.


jsignell:
If this passes I'll merge it. The only one missing is paths.

jsignell:
The only remaining notebook failure is from an example that is not changed in this PR.

PR #385, opened 2019-11-30
Add codecov.yml

ceball:
Mainly I'm triggering a commit to see if I've fixed codecov.io for hvplot.

Anyone know if we need this file? I copied it from panel.


ceball:
Same problem - repo is set up for coveralls, but badge links to codecov.io

Issue #395, opened 2020-01-09
Add keyword fontscale to hvplot


SandervandenOord:
Fontscale works nicely in holoviews:
[https://github.com/holoviz/holoviews/issues/4107]( https://github.com/holoviz/holoviews/issues/4107)

This scales all fonts (title, labels, x-axis and y-axis etc.) in the plot simultaneously and saves time.

Can this feature also be added to hvplot?

I guess syntax would look something like:
```python
df.hvplot.scatter(
    x='x', 
    y='y', 
    fontscale=1.4,
)
```

philippjfr:
Would you be willing to submit a PR?

SandervandenOord:
I'm going to give it a try :)

PR #397, opened 2020-01-10
add keyword fontscale (solving #395)

SandervandenOord:
Fontscale keyword was added so that hvplot can also handle this keyword. This should solve #395 . 

Also added documentation about fontscale in docstring of class HoloViewsConverter(object).

I didn't add fontscale to the __init__() of class HoloViewsConverter(object), but it falls now under the **kwds. If this still needs to be done, please let me know.

Tested some simple plots myself in Jupyter, that seemed to work. Didn't add any formal tests.

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -230,10 +230,9 @@ class HoloViewsConverter(object):
                      'yaxis', 'xformatter', 'yformatter', 'xlabel', 'ylabel',
                      'clabel', 'padding', 'responsive', 'max_height', 'max_width',
                      'min_height', 'min_width', 'frame_height', 'frame_width',
-                     'aspect', 'data_aspect']
+                     'aspect', 'data_aspect', 'fontscale']
 
-    _style_options = ['color', 'alpha', 'colormap', 'fontsize', 'c', 'cmap',
-                      'fontscale']
+    _style_options = ['color', 'alpha', 'colormap', 'fontsize', 'c', 'cmap']
 
     _op_options = ['datashade', 'rasterize', 'x_sampling', 'y_sampling',
                    'aggregator']
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -86,6 +86,9 @@ class HoloViewsConverter(object):
     fontscale: number
         Scales the size of all fonts by the same amount, e.g. fontscale=1.5
         enlarges all fonts (title, xticks, labels etc.) by 50%
+    fontsize: number or dict
+        Set title, label and legend text to the same fontsize. Finer control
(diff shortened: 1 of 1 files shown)


hvplot/converter.py
```diff
@@ -229,7 +232,8 @@ class HoloViewsConverter(object):
                      'min_height', 'min_width', 'frame_height', 'frame_width',
                      'aspect', 'data_aspect']
 
-    _style_options = ['color', 'alpha', 'colormap', 'fontsize', 'c', 'cmap']
+    _style_options = ['color', 'alpha', 'colormap', 'fontsize', 'c', 'cmap',
+                      'fontscale']
```
philippjfr: Note it's not a plot_option, I realize it's misnamed but it should be in `_axis_options`.

philippjfr:
Thanks! Not sure what's up with that test. Will have to look at it.

hvplot/converter.py
```diff
@@ -229,7 +232,8 @@ class HoloViewsConverter(object):
                      'min_height', 'min_width', 'frame_height', 'frame_width',
                      'aspect', 'data_aspect']
```
SandervandenOord: you mentioned plot_option in you note, but do you mean it shouldn't be in _style_options, but instead in _axis_options?

hvplot/converter.py
SandervandenOord: I have some questions still: you mention plot_option in your note, but do you mean it shouldn't be in _style_options, but instead in _axis_options?

hvplot/converter.py
```diff
@@ -398,7 +402,8 @@ def __init__(self, data, x, y, kind=None, by=None, use_index=True,
                    'padding', 'xformatter', 'yformatter',
                    'height', 'width', 'frame_height', 'frame_width',
                    'min_width', 'min_height', 'max_width', 'max_height',
-                   'fontsize', 'responsive', 'shared_axes', 'aspect', 'data_aspect']
+                   'fontsize', 'fontscale', 'responsive', 'shared_axes',
```
SandervandenOord: so i added fontscale here to variable plotwds. Is this correct? When should keywords be added to plotwds?

hvplot/converter.py
philippjfr: Sorry, yes, that's what I meant, edited.

hvplot/converter.py
philippjfr: There was some refactoring done here, which I unfortunately did not follow completely. I **think** only kwds that are not listed explicitly should be added and only kwds that need some further processing should be listed explicitly, so maybe drop fontsize from the `__init__` as part of this PR as well.

philippjfr:
Thanks @SandervandenOord!

Issue #405, opened 2020-01-18
Use __signature__ to generate signatures for hvplot accessors


philippjfr:
Currently we use a docstring trick to generate signatures for hvplot methods. This approach is no longer supported by IPython in recent versions and is also pretty hacky (and ugly to look at in a docstring since all the keywords appear there). Instead we should use the __signature__ object (see PEP362).

PR #411, opened 2020-01-27
Added support of tab completion via signature

jlstevens:
Work in progress PR to address #405.

### DIFF
--- a/hvplot/__init__.py
+++ b/hvplot/__init__.py
@@ -55,11 +55,25 @@ def _get_doc_and_signature(cls, kind, completions=False, docstring=True, generic
         defaults = argspec.defaults or [None]*len(args)
         for arg, dflt in zip(args, defaults):
             parameters.append((arg, dflt))
+        signature = None
     else:
         sig = inspect.signature(method)
         for name, p in list(sig.parameters.items())[1:]:
             if p.kind == 1:
                 parameters.append((name, p.default))
+
+        extra_kwargs = _hv.core.util.unique_iterator(valid_opts + kind_opts
+                                                     + converter._axis_options
+                                                     + converter._op_options)
+        filtered_signature = [p for p in sig.parameters.values()
+                              if p.kind != inspect.Parameter.VAR_KEYWORD]
+        extra_params = [inspect.Parameter(k, inspect.Parameter.KEYWORD_ONLY)
+                        for k in extra_kwargs
+                        if k not in [p.name for p in filtered_signature]]
+        all_params = (filtered_signature + extra_params
+                      + [inspect.Parameter('kwargs', inspect.Parameter.VAR_KEYWORD)])
+        signature = inspect.Signature(all_params)
+
     parameters += [(o, None) for o in
                    valid_opts+kind_opts+converter._axis_options+converter._op_options]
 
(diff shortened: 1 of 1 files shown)


jlstevens:
Tab completion based on ``__signature__`` seems to be working starting at 5c46947. Not quite sure if this PR is ready yet: I had to do a fair bit of filtering on the arguments due to name clashes. I'll see if the tests pass on travis before taking another look (and probably adding a unit test or two).

hvplot/__init__.py
```diff
@@ -55,21 +55,36 @@ def _get_doc(cls, kind, completions=False, docstring=True, generic=True, style=T
...
+        extra_kwargs = _hv.core.util.unique_iterator(valid_opts + kind_opts
+                                                     + converter._axis_options
+                                                     + converter._op_options)
+        filtered_signature = [p for p in sig.parameters.values()
+                              if p.kind != inspect.Parameter.VAR_KEYWORD]
+        extra_params = [inspect.Parameter(k, inspect.Parameter.KEYWORD_ONLY)
+                        for k in extra_kwargs
+                        if k not in [p.name for p in filtered_signature]]
+        all_params = (filtered_signature + extra_params
+                      + [inspect.Parameter('kwargs', inspect.Parameter.VAR_KEYWORD)])
```
jlstevens: Note that the ``**kwargs`` `VAR_KEYWORD`` at the end here is not strictly needed for tab-completion to work. If we don't except any other ``kwargs`` I am happy to remove it.

hvplot/__init__.py
```diff
@@ -55,21 +55,36 @@ def _get_doc(cls, kind, completions=False, docstring=True, generic=True, style=T
...
+        signature = inspect.Signature(all_params)
+
     parameters += [(o, None) for o in
                    valid_opts+kind_opts+converter._axis_options+converter._op_options]
```
philippjfr: Should the extra_kwargs processing and this not be merged somehow? Seems like they would be duplicated here as well.

hvplot/__init__.py
```diff
@@ -55,21 +55,36 @@ def _get_doc(cls, kind, completions=False, docstring=True, generic=True, style=T
...
+        signature = None
     else:
         sig = inspect.signature(method)
         for name, p in list(sig.parameters.items())[1:]:
             if p.kind == 1:
                 parameters.append((name, p.default))
+
+        extra_kwargs = _hv.core.util.unique_iterator(valid_opts + kind_opts
+                                                     + converter._axis_options
+                                                     + converter._op_options)
```
philippjfr: Don't have to change this now but I've adopted this indentation style (which conforms to blacks line wrapping rules) for Panel and hvPlot now:

```python        
extra_kwargs = _hv.core.util.unique_iterator(
    valid_opts + kind_opts + converter._axis_options + converter._op_options
)
```

hvplot/__init__.py
jlstevens: I assume you are talking about `parameters` in which case yes...there will be duplication in the docstring completion hint. The difference is that duplication in the docstring hint doesn't generate an error (and tab completion works fine) while the signature approach validates against duplication.

I can use `unique_iterator` here but I'm not sure it is worth doing anything much more complicated: we will be dropping Python 2 before long (so we won't be using the docstring hint anymore) and the only improvement would be a slight reduction in the verbosity of part of the docstring no one looks at anyway...

jlstevens: Addressed in bbf6f61.

hvplot/__init__.py
jlstevens: Addressed in bbf6f61.

philippjfr:
Thanks, looks good.

PR #442, opened 2020-04-15
Adding Google Analytics

kebowen730:
@jlstevens @philippjfr 

I added the Google Analytics UA to the conf.py file.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -61,7 +61,7 @@ html_context.update({
     # will work without this - for canonical (so can ignore when building locally or test deploying)
     'WEBSITE_SERVER': 'https://hvplot.pyviz.org',
     'VERSION': version,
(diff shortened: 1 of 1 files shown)


doc/conf.py
```diff
@@ -61,6 +61,7 @@
     # will work without this - for canonical (so can ignore when building locally or test deploying)
     'WEBSITE_SERVER': 'https://hvplot.pyviz.org',
     'VERSION': version,
+    'GOOGLE_ANALYTICS_UA': 'UA-154795830-1',
```
philippjfr, line 64:
```suggestion
    'GOOGLE_ANALYTICS_UA': 'UA-154795830-5',
```

kebowen730: I pulled that key from I believe Datashader, should I be incrementing that last number for each project I add it to?

PR #446, opened 2020-04-19
color_key alias for cmap

### LINKED ISSUE #444: allow color_key an alias for cmap
The ability to pass cmap dicts through to datashader color_key was solved in the fix for #267, but it may be good to also accept `color_key` as an argument. See https://discourse.holoviz.org/t/keyed-color-map-for-image/505.

BTW, @jsignell mentioned in the fix the need to think about the legend ... which would be great to have too.

itcarroll:
Closes #444 

This PR is WIP because:
1. I cannot get tox to run tests locally, hoping automation does it, but mostly
1. Use of `color_key` causes the faulty warning `WARNING:param.main: color_key option not found for image plot; similar options include: ['color_key', 'color']`

I suspect this has to do with [converter.py#L786](https://github.com/itcarroll/hvplot/blob/b43020cdc317f75f54a2395546f387e5c84afd43/hvplot/converter.py#L786), which looks a teensy bit like a dumpster fire and I'm not sure what it's supposed to do. Plz advise.

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -235,7 +235,8 @@ class HoloViewsConverter(object):
                      'min_height', 'min_width', 'frame_height', 'frame_width',
                      'aspect', 'data_aspect', 'fontscale']
 
-    _style_options = ['color', 'alpha', 'colormap', 'fontsize', 'c', 'cmap']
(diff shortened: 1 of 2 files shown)


philippjfr:
I think I'll merge and clean this up in another PR.

Issue #412, opened 2020-02-04
numpy 1.15.0 is required for hvplot 0.5.2


AlbertDeFusco:
#### ALL software version info (this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc)

```
conda create -n hv5 python=3.6 pandas=0.25 numpy=1.13.3 hvplot=0.5.2
```

#### Description of expected behavior and the observed behavior

Conda should have failed because NumPy 1.13.3 does not support `np.nanquantile`

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import pandas as pd
import hvplot.pandas
pd.DataFrame({'x':[0,0],'y':[1,1], 'c':['a','b']}).hvplot.scatter(x='x',y='y',color='c')

```

#### Stack traceback and/or browser JavaScript console output

```
Traceback (most recent call last):
  File "test.py", line 4, in <module>
    pd.DataFrame({'x':[0,0],'y':[1,1], 'c':['a','b']}).hvplot.scatter(x='x',y='y',color='c')
  File "/Users/adefusco/Applications/miniconda3/envs/hv5/lib/python3.6/site-packages/hvplot/plotting/core.py", line 191, in scatter
    return self(x, y, kind='scatter', **kwds)
  File "/Users/adefusco/Applications/miniconda3/envs/hv5/lib/python3.6/site-packages/hvplot/plotting/core.py", line 72, in __call__
    return self._get_converter(x, y, kind, **kwds)(kind, x, y)
  File "/Users/adefusco/Applications/miniconda3/envs/hv5/lib/python3.6/site-packages/hvplot/plotting/core.py", line 80, in _get_converter
    self._data, x, y, kind=kind, **params
  File "/Users/adefusco/Applications/miniconda3/envs/hv5/lib/python3.6/site-packages/hvplot/converter.py", line 460, in __init__
    symmetric = self._process_symmetric(symmetric, clim, check_symmetric_max)
  File "/Users/adefusco/Applications/miniconda3/envs/hv5/lib/python3.6/site-packages/hvplot/converter.py", line 513, in _process_symmetric
    cmin = np.nanquantile(data, 0.05)
AttributeError: module 'numpy' has no attribute 'nanquantile'
```

philippjfr:
Tbh, instead of requiring that let's just make our own wrapper around quantile which handles nans, e.g.:

```python
def nanquantile(data, q):
    return np.quantile(data[np.isfinite(data)], q)
```

[...]

PR #457, opened 2020-04-30
Do not use nanquantile for NumPy backward compatibility

philippjfr:
Fixes https://github.com/holoviz/hvplot/issues/412

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -519,8 +519,14 @@ class HoloViewsConverter(object):
         else:
             return
 
-        cmin = np.nanquantile(data, 0.05)
-        cmax = np.nanquantile(data, 0.95)
+        if hasattr(data, 'isna'):
+            finite = data[~data.isna()]
+        elif self.gridded_data:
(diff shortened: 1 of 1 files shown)


hvplot/converter.py
```diff
@@ -519,8 +519,14 @@ def _process_symmetric(self, symmetric, clim, check_symmetric_max):
...
-        cmin = np.nanquantile(data, 0.05)
-        cmax = np.nanquantile(data, 0.95)
+        if hasattr(data, 'isna'):
+            finite = data[~data.isna()]
+        elif self.gridded_data:
+            finite = data.flat[np.isfinite(data.flat)]
+        else:
+            finite = data[np.isfinite(data)]
+        cmin = np.quantile(finite, 0.05)
+        cmax = np.quantile(finite, 0.95)
```
jbednar, line 529: This seems like something to wrap up as a utility function.

philippjfr: I've given up on this, it's too tricky as there are too many potential edge cases.

PR #495, opened 2020-07-27
Adding datashade and rasterize options to scatter_matrix

### LINKED ISSUE #114: datashade kwarg for scatter_matrix
It would be nifty to have `datashade=True` work for the `scatter_matrix` function.  I'm happy to contribute this if someone can point me in the right direction.

ppwadhwa:
Fixes #114 

@tonyfast, @jlstevens and I worked through a solution to add a datashade option for a hvplot scatter_matrix.  @jlstevens pointed out datashade options (aggregator, rasterize, etc) may need to be considered when implementing this.  This PR, so far, only includes an argument to set `datashade = True` in the call to scatter_matrix, which will then use `grid.apply(datashade, per_element=True)`, as @philippjfr suggested as a solution in the issue.


hvplot/plotting/scatter_matrix.py
```diff
@@ -7,7 +7,7 @@
 
 
 @with_hv_extension
-def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **kwds):
+def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', datashade = False, alpha=0.5, **kwds):
```
philippjfr, line 10:
```suggestion
def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', datashade=False, alpha=0.5, **kwds):
```

jlstevens:
@philippjfr Looks like the travis builds are failing due to some python version issue. @ppwadhwa This means you will need to rebase this PR once this issue is fixed on master.

jlstevens:
I've just realized that although the name is `scatter_matrix` and it does have scatters, there is also the diagonal to think about:

image

Therefore, unless there are appropriate elements along the diagonal that can support datashader (I don't think histograms do, KDEs maybe?), only the non-diagonal scatters are likely to be datashaded right now, leaving the diagonals to be rendered with bokeh. Do you agree @philippjfr ?

philippjfr:
I do agree, yes.

jbednar:
The diagonals don't need Datashader, as they are already aggregated and thus independent of the dataset size.  The scatterplots do need Datashader, as they are dependent on the dataset size.

ppwadhwa:
I've made a few updates, so that I could include a few keyword arguments for datashade, dynspread, and rasterize.  I only have a couple of options, however,  I wanted to get some feedback on whether what I have is going in the right direction, or if there is a totally different way to approach this problem.

jlstevens:
>  I only have a couple of options, however, I wanted to get some feedback on whether what I have is going in the right direction, or if there is a totally different way to approach this problem.

@philippjfr may have a different opinion but what I see in the diff looks reasonable to me!

hvplot/plotting/scatter_matrix.py
```diff
@@ -4,10 +4,20 @@
...
+from holoviews.operation.datashader import shade as _shade 
+from holoviews.operation.datashader import dynspread as _dynspread 
+from holoviews.operation.datashader import rasterize as _rasterize 
 
 
 @with_hv_extension
-def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **kwds):
+def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, 
+    datashade=False, 
+    shade=False,
```
jbednar, line 16: I don't think there's a way that `shade=True` could work in a call to `scatter_matrix`; `shade()` expects an `Image` and converts it to an `RGB`, but here there are just points and histograms.  So I don't think this PR should include anything to do with `shade`, just `datashade` (the combination of rasterize and shade), `rasterize`, and `dynspread`. 

Maybe instead of `shade` you meant `spread`? Supporting `spread` would be useful, though you'd have to support at least some of its arguments the same way you support `max_px` for dynspread.

hvplot/plotting/scatter_matrix.py
```diff
@@ -38,19 +49,50 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **
...
                                           chart_type=chart,
                                           diagonal_type=diagonal)
-        grid = (grid * groups).map(lambda x: x.get(0) if isinstance(x.get(0), chart) else x.get(1),
+        if datashade == True:
+            grid = (grid * groups).map(lambda x: x.get(0).apply(_datashade, **ds_kwds) if isinstance(x.get(0), chart) else x.get(1),
+                                   _hv.Overlay)
+        elif dynspread == True:
+            grid = (grid * groups).map(lambda x: x.get(0)
+            .apply(_datashade, **ds_kwds).
+            apply(_dynspread, **dn_kwds) if isinstance(x.get(0), chart) else x.get(1),
```
jbednar, line 84: Splitting `.` and `apply` here seems very confusing.

hvplot/plotting/scatter_matrix.py
```diff
@@ -38,19 +49,50 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **
...
         chart_opts['color_index'] = c
         grid = grid.map(lambda x: x.clone(vdims=x.vdims+[c]), 'Scatter')
         groups = _hv.operation.gridmatrix(data.groupby(c).overlay(),
                                           chart_type=chart,
                                           diagonal_type=diagonal)
-        grid = (grid * groups).map(lambda x: x.get(0) if isinstance(x.get(0), chart) else x.get(1),
+        if datashade == True:
+            grid = (grid * groups).map(lambda x: x.get(0).apply(_datashade, **ds_kwds) if isinstance(x.get(0), chart) else x.get(1),
+                                   _hv.Overlay)
+        elif dynspread == True:
```
jbednar, line 81: I don't think this works; `dynspread` has to be applied *after* `datashade` or `rasterize`; it's not an alternative to either one, and requires that one of those has already completed. So I think the logic needs to be something like:

```
if rasterize or datashade:
    if datashade: datashade()
    else: rasterize()
    if dynspread: dynspread()
    else if spread: spread()
else:
    non-datashade pathway
```

hvplot/plotting/scatter_matrix.py
ppwadhwa: Yes, you are right.  I meant to remove shade.  I had included that not knowing what it was before, and realized it didn't work.

hvplot/plotting/scatter_matrix.py
ppwadhwa: that is an oops.  I fixed that and didn't push it.  I don't like that either.

hvplot/plotting/scatter_matrix.py
ppwadhwa: This makes sense.

ppwadhwa:
@jlstevens @jbednar thanks for looking this over!  @jbednar I made some updates per your comments.  Please take a look and tell me what you think.  thank you!

hvplot/plotting/scatter_matrix.py
```diff
@@ -38,19 +49,74 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **
     diagonal = HoloViewsConverter._kind_mapping[diagonal]
     chart = HoloViewsConverter._kind_mapping[chart]
 
+    #remove datashade kwds
+    if datashade == True or rasterize == True:
```
jbednar, line 53:
```suggestion
    if datashade or rasterize:
```

hvplot/plotting/scatter_matrix.py
```diff
@@ -38,19 +49,74 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **
...
+            .apply(_dynspread, **dn_kwds) if isinstance(x.get(0), chart) else x.get(1),
+                                _hv.Overlay)
+        elif spread:
+            grid = grid.map(lambda x: x.get(0)
+            .apply(_spread, **sp_kwds) if isinstance(x.get(0), chart) else x.get(1),
+                                _hv.Overlay)
+    else:
+        grid = grid.map(lambda x: x.get(0) if isinstance(x.get(0), chart) else x.get(1),
+                                _hv.Overlay)
+
```
jbednar, line 116: I haven't tested this code, but I'd find it easier to follow if it was pulled out as a named function (preferably with a docstring), so that I could follow the logic and parameters more easily. E.g.:

```suggestion
   def apply_to_charts(grid, op, kwds):
        return grid.map(lambda x: x.get(0).apply(op, **kwds)
                        if isinstance(x.get(0), chart)
                        else x.get(1), _hv.Overlay)
    
   if datashade or rasterize:
        if datashade: grid = apply_to_charts(_datashade, **ds_kwds)
        else:         grid = apply_to_charts(_rasterize, **ds_kwds)
        if dynspread: grid = apply_to_charts(_dynspread, **dn_kwds)
        elif spread:  grid = apply_to_charts(_spread,    **sp_kwds)
    else:
        grid = apply_to_charts(lambda x: x)
```

hvplot/plotting/scatter_matrix.py
```diff
@@ -38,19 +49,74 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **
...
     if c:
         chart_opts['color_index'] = c
         grid = grid.map(lambda x: x.clone(vdims=x.vdims+[c]), 'Scatter')
         groups = _hv.operation.gridmatrix(data.groupby(c).overlay(),
                                           chart_type=chart,
                                           diagonal_type=diagonal)
-        grid = (grid * groups).map(lambda x: x.get(0) if isinstance(x.get(0), chart) else x.get(1),
-                                   _hv.Overlay)
+        grid = grid * groups
```
jbednar, line 94: I'm having trouble following the logic here, but in the old code `.map` was applied only under the `if c:` branch, but in the new code the indentation indicates that it's applied regardless of `c`.  Is that a problem?

hvplot/plotting/scatter_matrix.py
```diff
@@ -4,10 +4,20 @@
 
 from ..converter import HoloViewsConverter
 from ..util import with_hv_extension
+from holoviews.operation.datashader import datashade as _datashade 
+from holoviews.operation.datashader import spread as _spread 
+from holoviews.operation.datashader import dynspread as _dynspread 
+from holoviews.operation.datashader import rasterize as _rasterize
```
jbednar, line 10: I think it would be slightly cleaner to do:

```suggestion
import holoviews.operation.datashader as hd 
```

And then you could use `hd.rasterize`, `hd.spread`, etc. below. That way I wouldn't have to wonder if the meaning of _spread was changed somewhere between here and where it's used, as it's just standard namespace usage.

jbednar (commented):
Looks good!

hvplot/plotting/scatter_matrix.py
jbednar: The new changes look good, thanks!  I still don't see the answer to this question; it seems like the code now in `apply_to_charts` was previously run only when `c` was non-None, and now it's always run. Is that intentional?  How was it working before if so? @philippjfr ?

ppwadhwa: I'm still working on this.  I think Jean-Luc agreed with you about moving the .map part this back under the c requirement.  I think I moved it, because it wouldn't apply the datashade/rasterize/etc if you didn't have c specified.  C was giving it a column to color code by.  Can you datashade without this color coding requirement?  If not, then the previous way was correct.  
anyway, this is not working correctly anyway with the fixes.  Right now, it seems to do something with datashade, but when I try to rasterize or give it a new value for dynspread, for example, it seems to ignore it.  So, I'm trying to figure out where I had it working and what I did wrong in the process of making corrections.

jbednar: My guess is that because of this change, something either won't work when `c` is specified either in the old code or the new code. I can't quite tell what the old code was meant to do, but hopefully @philippjfr can.

ppwadhwa: I made some comments about what each step is doing in this `if c` block.  All these lines are necessary to get the groups to color properly within each element in the scatter matrix.  I have moved things so that the datashade functions are performed after all the coloring is complete.

hvplot/plotting/scatter_matrix.py
```diff
@@ -38,19 +46,82 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **
...
+            grid = grid.map(lambda x: x.apply(hd.datashade, **ds_kwds)
+                                        .apply(hd.spread, **sp_kwds)
+                                        if isinstance(x, chart) else x)
+        else:
+            grid = grid.map(lambda x: x.apply(hd.datashade, **ds_kwds) 
+                            if isinstance(x, chart) else x)
+
+    elif rasterize:
+        grid = grid.map(lambda x: x.apply(hd.rasterize, **ds_kwds)
+                                        if isinstance(x, chart) else x)
```
ppwadhwa, line 122: I ended up removing the `apply_to_charts` function, because you need to apply the options (dynspread or spread) for datashade right after performing the datashade function. If you try to run `apply_to_charts` once, and then do it a second time to apply dynspread or spread, it will check for elements that are of the type `scatter`.  These are no longer there after they were datashaded, so it doesn't do the second step. For some reason, trying to change things around to check for whether an element is of the type "hist" (or "histogram") doesn't work.  I was trying to keep the "apply_to_charts" function, by doing something like this:
```
def apply_to_charts(grid, op, kwds):
        return grid.map(lambda x: x
                        if isinstance(x, diagonal)
                        else x.apply(op, **kwds))```
or even:
```
def apply_to_charts(grid, op, kwds):
        return grid.map(lambda x: x.apply(op, **kwds)
                        if not isinstance(x, diagonal)
                        else x)```
where diagonal is the type for the diagonal elements.
This would give me an error that the histogram type did not any 'bounds'.

ppwadhwa:
Note on a need to do:  I don't have dynspread or spread working for the rasterized plot.  @jlstevens just told me that an updated version of datashader should support this, so I will update datashader, and add these options back in.

hvplot/plotting/scatter_matrix.py
```diff
@@ -38,19 +46,86 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **
...
+    #remove datashade kwds
+    if datashade or rasterize:
+        ds_kwds = {}
+        if 'aggregator' in kwds:
+            ds_kwds['aggregator'] = kwds.pop('aggregator')
+        
+    #remove dynspread kwds
+    if dynspread:
+        if hd.ds_version < '0.11.1':
+            print('WARNING: Any version of datashader less than 0.11.1 will not support dynspread.')
```
ppwadhwa, line 58: @jlstevens I added in some warnings about the datashader version, but I am unsure of the version exactly.  I tried to make my best guess based on the repo, but I haven't installed older versions to test it.  Also, should this be an error?  Or should I be using a logger?

jlstevens: I'll get back to you on the exact version. As for warning vs exception, I think asking for `spread` or `dynspread` in the `rasterize` case is an error for those old versions as the user requested something that the function cannot deliver. I would raise an appropriate Exception (I haven't thought of what exactly) that includes the message you wrote above but also explains that the issue isn't asking for `dynspread` but asking for `dynspread` *with* `rasterize`.

jlstevens: I just confirmed that 0.11.1 also doesn't have the necessary changes 0.12.0 should though. So I would use `< '0.12.0` here. Also, is `hd.ds_version` also a string? If so, I would use one of the version comparison utilities as string version comparison is unsafe.

ppwadhwa: hd.ds_version is a LooseVersion.  I was just looking up some different ways to do this, and will make a change.

jlstevens: Ok great! Was just checking...

ppwadhwa: @jlstevens I looked around the repo and it seems that for all the looseVersions of various packages, its compared to a string.  Is there a different way I should do the version comparison?

jbednar: No, you're good; Jean-Luc was just checking that ds_version is already a LooseVersion, because if it were not, there would be problems.

ppwadhwa: Great, thanks @jbednar !

hvplot/plotting/scatter_matrix.py
```diff
@@ -38,19 +47,99 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **
...
+                                        if isinstance(x, chart) else x)
+        elif spread:
+            grid = grid.map(lambda x: x.apply(hd.datashade if datashade 
+                                            else hd.rasterize, **ds_kwds)
+                                        .apply(hd.spread, **sp_kwds)
+                                        if isinstance(x, chart) else x)
+        else:
+            grid = grid.map(lambda x: x.apply(hd.datashade if datashade 
+                                            else hd.rasterize, **ds_kwds) 
+                            if isinstance(x, chart) else x)
```
jbednar, line 141: Seems like this can all be done just once:

```suggestion
    if datashade or rasterize:
        aggregatefn = hd.datashade if datashade else hd.rasterize
        spreadfn    = hd.dynspread if dynspread else (hd.spread if spread else lambda z, **_: z)
    
        grid = grid.map(lambda x: x.apply(aggregatefn, **ds_kwds).apply(spreadfn, **sp_kwds) if isinstance(x, chart) else x)
```
It's slightly tricky to have to define a no-op when there is no spreading, but I still think it's better than copying all that grid.map code with subtle differences.

hvplot/plotting/scatter_matrix.py
```diff
@@ -38,19 +47,99 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **
...
+                raise RuntimeError('Any version of datashader ' +
+                                    'less than 0.12.0 does not support ' + 
+                                    'rasterize with dynspread or spread')
+    #remove datashade kwds
+    if datashade or rasterize:
+        ds_kwds = {}
+        if 'aggregator' in kwds:
+            ds_kwds['aggregator'] = kwds.pop('aggregator')
+        
+    #remove dynspread kwds
```
jbednar, line 62:
```suggestion
    #remove dynspread kwds
    sp_kwds={}
```

(Needed for changes below)

hvplot/plotting/scatter_matrix.py
```diff
@@ -38,19 +47,99 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **
...
+
+        dn_kwds = {}
+        if 'max_px' in kwds:
+            dn_kwds['max_px'] = kwds.pop('max_px')
+        if 'threshold' in kwds:
+            dn_kwds['threshold'] = kwds.pop('threshold')
+        if 'how' in kwds:
+            dn_kwds['how'] = kwds.pop('how')
+        if 'mask' in kwds:
+            dn_kwds['mask'] = kwds.pop('mask')
```
jbednar, line 78:
```suggestion
        if 'max_px' in kwds:
            sp_kwds['max_px'] = kwds.pop('max_px')
        if 'threshold' in kwds:
            sp_kwds['threshold'] = kwds.pop('threshold')
        if 'how' in kwds:
            sp_kwds['how'] = kwds.pop('how')
        if 'mask' in kwds:
            sp_kwds['mask'] = kwds.pop('mask')
```

(needed for changes below; only either dynspread or spread can be used at one time, so we can pool their options.)

hvplot/plotting/scatter_matrix.py
```diff
@@ -38,19 +47,99 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **
...
+            
+
+    if spread:
+        if datashade == False and rasterize == False:
+            warnings.warn(
+                "Datashade or Rasterize must be specified to use spread. " 
+                "Spread will not be applied to plots."
+                )
+        sp_kwds = {}
```
jbednar, line 87:
```suggestion
```

(needed for changes below)

jbednar (commented):
Looks good to me, thanks so much! I've suggested a way to reduce having so many code paths at the end by using function pointers (including a no-op for spreading when there is no spreading); hopefully you can see how it works. If it still works after applying these changes, I'm happy to merge it!

jbednar (commented):
I've reviewed this and tried it out. Seems to work great! The only problem I can see is that the plot isn't dynamic. Normally in hvplot, datashaded and rasterized plots set `dynamic=True` so that they will update automatically on zooming by default. Here, it doesn't seem to re-render on zooming, and also doesn't accept the `dynamic` parameter so we can't enable being `dynamic`. 

@philippjfr , what needs to be done to support `dynamic`?

ppwadhwa:
Thanks for doing all this, @jbednar !  It looks much prettier with your changes.  :)

philippjfr:
So I finally tracked down why this was such a nightmare. Specifically on the HoloViews side any layout or grid of plots tries to merge stream callbacks that have the same Bokeh models as their source. The problem was that it even merged stream callbacks that only partially overlapped in their handles, e.g. if you had two RangeXY callbacks attached to separate plots, if both were attached to the same Range1D object they would get merged. This is obviously incorrect since a) the fact that the x_range is the same does not mean the y_range will also match and b) even if x_range and y_range match they might be transposed.

I've pushed https://github.com/holoviz/holoviews/pull/5133 which will have to be backported to HoloViews 1.14.6

philippjfr:
@maximlt will rebase the PR and then let's merge it. Fixing the bug in HoloViews doesn't have to hold up this PR.

jlstevens:
Will you want to pin HoloViews before the next hvplot release or maybe just warn when using `hvplot.scatter_matrix` when older HoloViews versions are in use?

philippjfr:
No pins but warning seems reasonable, but only specifically when using `scatter_matrix` with `holoviews<1.14.7 ` and with `datashade`/`rasterize` enabled.

maximlt:
Now the CI is green (well there's some coverage decrease, despite the fact that I've added quite a number of tests 🤔 ) does any of one of you @philippjfr or @jbednar want to review it?

The biggest change I've made was to wrap the rasterize/datashade and dynspread/spread operations into a partial to allow to pass keyword arguments to the operations, since `grid` doesn't accept keyword arguments.

examples/user_guide/Pandas_API.ipynb
```diff
@@ -699,6 +706,24 @@
     "scatter_matrix(df, alpha=0.2, diagonal='kde')"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Since scatter matrix plots may generate a large number of points (the one above renders 120,000 points!), you man want to take advantage of the power provided by [Datashader](https://datashader.org/) to rasterize the off-diagonal plots:"
```
jbednar, line 713:
```suggestion
    "Since scatter matrix plots may generate a large number of points (the one above renders 120,000 points!), you may want to take advantage of the power provided by [Datashader](https://datashader.org/) to rasterize the off-diagonal plots into a fixed-resolution representation:"
```

examples/user_guide/Pandas_API.ipynb
```diff
@@ -699,6 +706,24 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "df = pd.DataFrame(np.random.randn(10000, 4), columns=['a', 'b', 'c', 'd'])\n",
+    "\n",
+    "scatter_matrix(df, datashade=True, dynspread=True)"
```
jbednar, line 724: Is there a reason to use `datashade=True` here instead of `rasterize=True`? Generally we steer people towards `rasterize` unless there is a compelling reason not to, given that `rasterize` supports colorbars, hover, etc.

examples/user_guide/Pandas_API.ipynb
```diff
@@ -1058,9 +1081,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "scrolled": true
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "# Doesn't currently work as expected. Work-around:\n",
```
jbednar, line 1087: What is the workdaround here? Seems like a comment that needs deleting? Also "Colormaps can also be used other plot types" should maybe be "Colormaps can also be used with other plot types", but GH won't let me edit that since it presumably predates this PR.

hvplot/plotting/scatter_matrix.py
```diff
@@ -43,7 +46,22 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist',
         Default is `Category10 <https://github.com/d3/d3-3.x-api-reference/blob/master/Ordinal-Scales.md#category10>`.
     diagonal_kwds/hist_kwds/density_kwds: dict, optional
         Keyword options for the diagonal plots
-    kwds: Keyword options for the off-diagonal plots, optional
+    datashade (default=False):
+        Whether to apply rasterization and shading using datashader
+        library returning an RGB object
+    rasterize (default=False):
+        Whether to apply rasterization using the datashader library
+        returning an aggregated Image
```
jbednar, line 54:
```suggestion
    datashade (default=False):
        Whether to apply rasterization and shading (colormapping) using
        the Datashader library, returning an RGB object instead of 
        individual points
    rasterize (default=False):
        Whether to apply rasterization using the Datashader library,
        returning an aggregated Image (to be colormapped by the 
        plotting backend) instead of individual points
```

hvplot/plotting/scatter_matrix.py
```diff
@@ -43,7 +46,22 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist',
...
+    dynspread (default=False):
+        Allows plots generated with datashade=True or rasterize=True 
+        to increase the point size to make sparse regions more visible.
+        kwds supported include ``max_px``, ``threshold``,  ``shape``, ``how`` and ``mask``.
+    spread (default=False):
+        Allows plots generated with datashade=True or rasterize=True 
+        to increase the point size to make sparse regions more visible, by
```
jbednar, line 61:
```suggestion
        Make plots generated with datashade=True or rasterize=True 
        increase the point size to make points more visible, by
```

hvplot/plotting/scatter_matrix.py
```diff
@@ -43,7 +46,22 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist',
...
+    datashade (default=False):
+        Whether to apply rasterization and shading using datashader
+        library returning an RGB object
+    rasterize (default=False):
+        Whether to apply rasterization using the datashader library
+        returning an aggregated Image
+    dynspread (default=False):
+        Allows plots generated with datashade=True or rasterize=True 
+        to increase the point size to make sparse regions more visible.
+        kwds supported include ``max_px``, ``threshold``,  ``shape``, ``how`` and ``mask``.
```
jbednar, line 58:
```suggestion
        For plots generated with datashade=True or rasterize=True, 
        automatically increase the point size when the data is sparse
        so that individual points become more visible.
        kwds supported include ``max_px``, ``threshold``,  ``shape``, ``how`` and ``mask``.
```

hvplot/plotting/scatter_matrix.py
```diff
@@ -65,20 +84,88 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist',
     diagonal = HoloViewsConverter._kind_mapping[diagonal]
     chart = HoloViewsConverter._kind_mapping[chart]
 
+    if rasterize or datashade:
+        try:
+            import datashader  # noqa
+        except ImportError:
+            raise ImportError("rasterize or datashade require "
```
jbednar, line 91:
```suggestion
            raise ImportError("rasterize and datashade require "
```

hvplot/plotting/scatter_matrix.py
```diff
@@ -65,20 +84,88 @@ def scatter_matrix(data, c=None, chart='scatter', diagonal='hist',
...
+        if hv_version <= '1.14.6':
+            warnings.warn(
+                "Versions of holoviews before 1.14.7 did not suppport "
+                "dynamic update of rasterized/datashaded scatter matrix. "
+                "Update holoviews to a newer version."
+            )
+    
+    if rasterize and datashade:
+        raise ValueError("Choose to either rasterize or "
+                         "datashade the scatter matrix.")
```
jbednar, line 103:
```suggestion
                         "datashade the scatter matrix, not both.")
```

hvplot/tests/plotting/testscattermatrix.py
```diff
@@ -0,0 +1,144 @@
...
+
+class TestDatashader(TestCase):
+
+    def setUp(self):
+        try:
+            import datashader # noqa
+        except:
+            raise SkipTest('Datashader not available')
+        if sys.maxsize < 2**32:
+            raise SkipTest('Datashader does not work on 32-bit systems')
```
jbednar, line 95:
```suggestion
            raise SkipTest('Datashader does not support 32-bit systems')
```

jbednar (approved):
Looks good! I made various trivial comments and improvements that should be able to be applied easily, and then I'm happy for this to be merged.

examples/user_guide/Pandas_API.ipynb
maximlt: No particular reason no, except that I wasn't aware of that recommendation and that I've seen instances of `datashade` being used in hvplot's docs, for instance in the Gridded data user guide.

maximlt:
```suggestion
    "scatter_matrix(df, rasterize=True, dynspread=True)"
```

examples/user_guide/Pandas_API.ipynb
maximlt: Yes it's unrelated to this PR. I will open a new PR with changes to the docs.

hvplot/tests/plotting/testscattermatrix.py
```diff
@@ -0,0 +1,144 @@
...
+import sys
+
+from parameterized import parameterized
+import numpy as np
+import pandas as pd
+# try:
+#     import numpy as np
+#     import hvplot.networkx as hvnx
+# except:
+#     raise SkipTest('NetworkX not available')
```
maximlt, line 11:
```suggestion
```

examples/user_guide/Pandas_API.ipynb
maximlt: https://github.com/holoviz/hvplot/pull/678

examples/user_guide/Pandas_API.ipynb
jbednar: There are still a couple of cases that only `datashade` covers, such as alpha-mapping an arbitrary color (with count nonlinearly mapped to the alpha channel rather than a colormap) and categorical color mixing (as in the census and ship_traffic examples). Any other case should use `rasterize`, though there are a lot of older examples written before Bokeh 2.3's datashader colormapping support that need to be updated not to use `datashade`.

philippjfr:
Going to merge! Sorry it took so long @ppwadhwa, turns out the issue was in HoloViews. Thank you!

maximlt:
Yes thanks to everybody who contributed to this PR!!!

PR #505, opened 2020-08-31
Implement interactive API for all data backends

philippjfr:
Implements interactive API as described in https://github.com/pydata/xarray/issues/3709.

Depends on https://github.com/holoviz/holoviews/pull/4578

- [x] Add documentation
- [ ] Add tests

### DIFF
--- a/hvplot/interactive.py
+++ b/hvplot/interactive.py
@@ -16,7 +16,7 @@ from panel.widgets.base import Widget
 
 class Interactive():
     """
-    Interactive is a wrapper around and object that lets users create
+    Interactive is a wrapper around a Python object that lets users create
     interactive pipelines by calling existing APIs on an object with
     dynamic parameters or widgets.
     """
@@ -267,7 +267,7 @@ class Interactive():
     def layout(self):
         """
         Returns a layout of the widgets and output arranged according
-        to the center and and widget location specified in the
+        to the center and widget location specified in the
         interactive call.
         """
         widget_box = self.widgets
--- a/hvplot/interactive.py
+++ b/hvplot/interactive.py
@@ -1,0 +1,362 @@
+"""
+interactive API
+"""
+
+import abc
+import operator
+import sys
+
+import holoviews as hv
+import panel as pn
+import param
+
+from panel.layout import Column, Row, VSpacer, HSpacer
+from panel.widgets.base import Widget
+
+
+class Interactive():
+    """
+    Interactive is a wrapper around and object that lets users create
+    interactive pipelines by calling existing APIs on an object with
+    dynamic parameters or widgets.
+    """
+
+    __metaclass__ = abc.ABCMeta
+
+    _fig = None
+
+    def __init__(self, obj, transform=None, plot=False, depth=0,
+                 loc='top_left', center=False, dmap=False, **kwargs):
+        self._obj = obj
+        self._method = None
+        if transform is None:
+            self._transform = hv.dim('*')
+        else:
+            self._transform = transform
+        self._plot = plot
+        self._depth = depth
+        self._loc = loc
+        self._center = center
+        self._dmap = dmap
+        self._kwargs = kwargs
+        ds = hv.Dataset(self._obj)
+        self._current = self._transform.apply(ds, keep_index=True, compute=False)
+
+    @property
+    def _params(self):
+        return [v for k, v in self._transform.params.items() if k != 'ax']
+
+    @property
+    def _callback(self):
+        @pn.depends(*self._params)
+        def evaluate(*args, **kwargs):
+            ds = hv.Dataset(self._obj)
+            obj = self._transform.apply(ds, keep_index=True, compute=False)
+            if self._plot:
+                return Interactive._fig
+            else:
+                return obj
+        return evaluate
+
+    def _clone(self, transform=None, plot=None, loc=None, center=None,
+              dmap=None, **kwargs):
+        plot = self._plot or plot
+        transform = transform or self._transform
+        loc = self._loc if loc is None else loc
+        center = self._center if center is None else center
+        dmap = self._dmap if dmap is None else dmap
+        depth = self._depth+1
+        kwargs = dict(self._kwargs, **kwargs)
+        return type(self)(self._obj, transform, plot, depth,
+                           loc, center, dmap, **kwargs)
+
+    def _repr_mimebundle_(self, include=[], exclude=[]):
+        return self.layout._repr_mimebundle_()
+
+    def __dir__(self):
+        return [d for d in dir(self._current) if d[0] != '_']
+
+    def __getattr__(self, name):
+        if name in dir(self):
+            self._method = name
+            self.__doc__ = getattr(self._obj, name).__doc__
+            return self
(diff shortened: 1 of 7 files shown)


hvplot/interactive.py
```diff
@@ -0,0 +1,362 @@
...
+import panel as pn
+import param
+
+from panel.layout import Column, Row, VSpacer, HSpacer
+from panel.widgets.base import Widget
+
+
+class Interactive():
+    """
+    Interactive is a wrapper around and object that lets users create
```
jbednar, line 19:
```suggestion
    Interactive is a wrapper around a Python object that lets users create
```

hvplot/interactive.py
```diff
@@ -0,0 +1,362 @@
...
+        other = other._transform if isinstance(other, Interactive) else other
+        transform = type(self._transform)(self._transform, operator.floordiv, other)
+        return self._clone(transform)
+    def __ge__(self, other):
+        other = other._transform if isinstance(other, Interactive) else other
+        transform = type(self._transform)(self._transform, operator.ge, other)
+        return self._clone(transform)
+    def __gt__(self, other):
+        other = other._transform if isinstance(other, Interactive) else other
+        transform = type(self._transform)(self._transform, operator.gt, other)
```
jbednar, line 148: Is there a way to avoid specifying each of these delegations explicitly like this?  The answer may well be no, but it seems like a shame to have so much boilerplate code here, which makes it hard to see the real code, and would cover up any special processing that might be happening in one of these boilerplate methods.

hvplot/interactive.py
```diff
@@ -0,0 +1,362 @@
...
+    def __rsub__(self, other):
+        other = other._transform if isinstance(other, Interactive) else other
+        transform = type(self._transform)(self._transform, operator.sub, other, reverse=True)
+        return self._clone(transform)
+    def __rtruediv__(self, other):
+        other = other._transform if isinstance(other, Interactive) else other
+        transform = type(self._transform)(self._transform, operator.truediv, other, reverse=True)
+        return self._clone(transform)
+
+    def plot(self, *args, **kwargs):
```
jbednar, line 245: Can you add a docstring explaining why `plot` needed this special code?  Does it mean that `.interactive` won't work with other .plot-API packages like cufflinks/pdvega, and/or related packages like Plotly Express?

hvplot/interactive.py
```diff
@@ -0,0 +1,362 @@
...
+
+    #----------------------------------------------------------------
+    # Public API
+    #----------------------------------------------------------------
+
+    @property
+    def layout(self):
+        """
+        Returns a layout of the widgets and output arranged according
+        to the center and and widget location specified in the
```
jbednar, line 270:
```suggestion
        to the center and widget location specified in the
```

hvplot/interactive.py
```diff
@@ -0,0 +1,362 @@
...
+        elif loc in ('top', 'bottom'):
+            widgets = Row(HSpacer(), widget_box, HSpacer())
+        elif loc in ('top_left', 'bottom_left'):
+            widgets = Row(widget_box, HSpacer())
+        elif loc in ('top_right', 'bottom_right'):
+            widgets = Row(HSpacer(), widget_box)
+        elif loc in ('left_top', 'right_top'):
+            widgets = Column(widget_box, VSpacer())
+        elif loc in ('left_bottom', 'right_bottom'):
+            widgets = Column(VSpacer(), widget_box)
```
jbednar, line 287: Seems like this layout code ought to be pulled out into its own function to make this method shorter and easier to follow.  Plus this functionality seems similar to `interact`, DynamicMaps, and other cases where widgets are generated and added to a layout, and it would be nice to have a consistent approach for controlling those.

hvplot/pandas.py
```diff
@@ -10,9 +10,15 @@ def patch(name='hvplot', extension='bokeh', logo=False):
...
-    setattr(pd.Series, name, patch_property)
+    plot_prop = property(_patch_plot)
+    setattr(pd.DataFrame, name, plot_prop)
+    setattr(pd.Series, name, plot_prop)
+
+    _patch_interactive = lambda self: Interactive(self)
+    _patch_interactive.__doc__ = Interactive.__call__.__doc__
+    interactive_prop = property(_patch_plot)
+    setattr(pd.DataFrame, name, interactive_prop)
+    setattr(pd.Series, name, interactive_prop)
```
jbednar, line 21: Is there a supported way to make the interactive support available without `import hvplot.pandas` or similar sideeffect-y imports?  Personally, I'll use `import hvplot.pandas`, but I'm curious if there is more awkward but less magic alternative, either to explicitly perform the monkeypatching or to have a non-monkeypatch alternative (e.g. a global function instead of a method).

hvplot/interactive.py
philippjfr: I'll add a comment, and no. It's really just matplotlib which has this implicit state nonsense, which requires me to work around. I also still have to add special handling for Pandas plot engines.

hvplot/pandas.py
philippjfr: `Interactive(ds)` will work, I can also provide a helper to do the monkey-patching manually.

martindurant:
I want to include these in Intake plot specs, so that source.plot() can produce interactive things! 
Any thoughts on how that might happen?

jbednar:
> I want to include these in Intake plot specs, so that source.plot() can produce interactive things!
> Any thoughts on how that might happen?

Good question. Intake plot specs are YAML, like:

```
      plots:
        dropoff_scatter:
          kind: scatter
          x: dropoff_x
          y: dropoff_y
          datashade: True
          width: 800
          height: 600
```
I would guess that it's already possible to get some limited interactivity, e.g. by adding `groupby: time` to generate a HoloMap over time.  But it would take some brainstorming to figure out how to support interactivity in the more general sense of this PR, e.g. to get a widget to select the column to plot as `x` and `y`, or to use for the value of `datashade` here. 

Maybe some syntax like `x: select [dropoff_x, pickup_x]` or `datashade: boolean [True]"` or `width: integer [1, 800, 2000]`?  It gets tricky even in this simple example, though, because x and y should be selected together, leading to wanting a selection [dropoff, pickup] and `x: select [dropoff, pickup]+_x` and `y: ...somehow same widget as x ...+_x`.  Obviously a declarative spec can't reasonably handle all the cases, but maybe there's a feasible and valuable subset.

martindurant:
Something like this?
```python
slider = pn.widgets.IntSlider(name='time', start=0, end=10)
ds.air.interactive.isel(time=slider).plot()
```
->
```yaml
plots:
  example:
    kind: pipeline
    pipeline:
      - getattr: air
      - isel:
        time:
          widget: IntSlider
          name: time
          start: 0
          end: 10
    plotargs: {}
```

For the API, it'd also be great to be able to call with some value for the widgets to get the output data that would be plotted. That would also cover the "process my data in a pipeline" request that we've wondered about for a while (i.e., use this model even when there are no interactive elements). 

Of course, this is quite long and verbose, and maybe doesn't all belong in the "plots" section, especially if we want to make the derived data available. In short, I think it's all doable, we just have to decide what feels convenient for the user.

PR #523, opened 2020-11-15
Made a pass over the Interactive user guide

jbednar:
Interactive.ipynb looks great! Here I've made mostly minor cleanups and clarifications. The biggest addition is a Panel app at the very end. I think it's important to show that .interactive isn't a dead end, i.e. that it can be reconfigured and used in a larger app the same way as interact() can. Remaining issues:

- [ ] The text doesn't explain "accessor"; it probably should!
- [ ] I think there should be an example where multiple plots are shown in a layout. I think that would require referring to an inferred DiscreteSlider widget so that it can be passed into another .interactive call, but that seemed a bit hairy to explain.
- [ ] The Panel app I made isn't gorgeous; @philippjfr may see some simple way to make something more compelling.

### DIFF
--- a/examples/user_guide/Interactive.ipynb
+++ b/examples/user_guide/Interactive.ipynb
@@ -10,6 +10,7 @@
     "import hvplot.xarray # noqa\n",
     "import hvplot.pandas # noqa\n",
     "import panel as pn\n",
+    "import panel.widgets as pnw\n",
     "import ipywidgets as ipw"
    ]
   },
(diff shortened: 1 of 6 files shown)


examples/user_guide/Interactive.ipynb
```diff
@@ -327,15 +325,21 @@
...
+    "    pn.Row(\n",
+    "        pn.panel(\"https://hvplot.holoviz.org/assets/hvplot-wm.png\", width=100), \n",
+    "        pn.Spacer(width=20),\n",
+    "        pn.Column(\n",
+    "            pn.panel(\"## Select a time and type of plot\", width=400),\n",
+    "            interactive.widgets\n",
+    "        ),\n",
+    "        pn.panel(\"https://panel.holoviz.org/_static/logo_stacked.png\", width=100)\n",
+    "    ),\n",
+    "    interactive.panel()\n",
```
jbednar, line 341: Is there a good reason why interactive.panel() has method-like syntax while interactive.widgets has attribute-like access? If so, fine, but it does seem confusing and hard to remember.

philippjfr: Good point, I'll standardize on method syntax.

PR #525, opened 2020-11-17
Provide color_key to datashade as well as cmap

philippjfr:
If no explicit color_key is supplied this also supplies the colormap to the color_key argument.

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -1097,6 +1097,8 @@ class HoloViewsConverter(object):
         style = {}
         if self.datashade:
             operation = datashade
+            if 'cmap' in opts:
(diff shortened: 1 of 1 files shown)


hvplot/converter.py
```diff
@@ -1097,6 +1097,8 @@ def method_wrapper(ds, x, y):
         style = {}
         if self.datashade:
             operation = datashade
+            if 'cmap' in opts:
+                opts['color_key'] = opts['cmap']
```
jbednar, line 1101: Won't this overwrite an explicit `color_key` that the user supplied?

philippjfr: Forgot to push my other change.

jbednar (approved):

PR #535, opened 2020-12-15
Add a warning message about interactivity not working on static website

lhoupert:
It took me a little bit of time to realise why the interactivity was not working on the website... 
Then I saw the warning message on the widget page (https://hvplot.holoviz.org/user_guide/Widgets.html). I think it would be worth having the same type of message on this webpage...

### DIFF
--- a/examples/user_guide/Interactive.ipynb
+++ b/examples/user_guide/Interactive.ipynb
@@ -25,6 +25,15 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks, this is definitely needed. However I believe @jlstevens is currently adding an option to all our websites that looks like the floating box you see on this page: https://pyviz-dev.github.io/panel/user_guide/Widgets.html

jlstevens:
Yes, this is being addressed: https://github.com/holoviz/hvplot/pull/550

PR #550, opened 2021-01-17
Disable interactivity warning on homepage

jlstevens:
Disables the warning on the homepage (when built with latest nbsite).


jlstevens:
The `nbsite build` command will also need to be updated when doc building is moved to GitHub Actions.

Issue #580, opened 2021-03-26
Compatibility with python 3.8 ?


loicduffar:
Thanks for contacting us! Please read and follow these instructions carefully, then delete this introductory text to keep your issue easy to read. Note that the issue tracker is NOT the place for usage questions and technical assistance; post those at [Discourse](https://discourse.holoviz.org) instead. Issues without the required information below may be closed immediately.

#### Additional context

According to [installation instructions](https://github.com/holoviz/hvplot), hvplot is supported by python 3.5-3.7 (and only 3.6 according to [setup.py](https://github.com/holoviz/hvplot/blob/master/setup.py) ).

I would like to migrate to python 3.8 (Anaconda), but I want to be sure all libraries I use are compatibles.

philippjfr:
This is simply a matter of not having updated the setup.py (a PR to fix it would be very welcome). I've been using 3.8 daily for months without issues.

loicduffar:
Thanks ! 
I am happy to hear that.

I'll try to post a PR to fix it ... but it's the first time for me, and I haven't figured out the principle yet...

loicduffar:
I have done the PR.
You may tell me if it is not the correct way ....

PR #581, opened 2021-03-26
Update setup.py

loicduffar:
#580
loicduffar 
According to installation instructions, hvplot is supported by python 3.5-3.7 (and only 3.6 according to setup.py ).

I would like to migrate to python 3.8 (Anaconda), but I want to be sure all libraries I use are compatibles.

philippjfr
 This is simply a matter of not having updated the setup.py (a PR to fix it would be very welcome). I've been using 3.8 daily for months without issues.


philippjfr:
Thanks @loicduffar! Appreciate it.

PR #590, opened 2021-04-13
Fix docstring for x_sampling/y_sampling

michaelaye:


philippjfr:
Thanks for catching that!

PR #591, opened 2021-04-14
Add cnorm option

philippjfr:


hvplot/converter.py
```diff
@@ -84,6 +84,8 @@ class HoloViewsConverter(object):
     ---------------
     colorbar (default=False): boolean
         Enables colorbar
+    cnorm (default='linear'): str
+        Color scaling which must be one of 'linear', 'log' or 'eq_hist'
```
jbednar, line 88: Given that hvPlot is not enforcing this "must be" clause, I'd think we should say something more like "Color scaling (normalization) as supported by the HoloViews cnorm option, such as 'linear', 'log', or 'eq_hist'.".  That way when HoloViews and/or Bokeh add additional normalization options, this docstring will remain correct.

PR #601, opened 2021-04-27
preserve function signature and docstring in `with_hv_extension`

### LINKED ISSUE #600: `hvplot.util.with_hv_extension` hides docstrings

  * hvplot = 0.7.0

However looking at the source code this issue should still be present in MASTER.

Functions wrapped with the `hvplot.util.with_hv_extension` decorator have their docstrings hidden by the `wrapper` docstring.

[`functools.wraps`](https://docs.python.org/3/library/functools.html#functools.wraps) can fix this.

Example
```Python
>>> import hvplot
>>> help(hvplot.scatter_matrix)
```

Result:

```
Help on function wrapper in module hvplot.util:

wrapper(*args, **kwargs)
```

This is the docstring from the `wrapper` function inside `hvplot.util.with_hv_extension`.

toddrjen:
This change allows `hvplot.utils.with_hv_extension` to preserve function signatures and docstrings, which are currently lost.  Fixes #600.

### DIFF
--- a/hvplot/util.py
+++ b/hvplot/util.py
@@ -6,6 +6,7 @@ from __future__ import absolute_import
 import sys
 
 from distutils.version import LooseVersion
+from functools import wraps
 from types import FunctionType
 
 import pandas as pd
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks @toddrjen!

PR #602, opened 2021-04-27
support setting options for diagonal plots in scatter_matrix

### LINKED ISSUE #599: Allow specifying keyword arguments for diagonal
#### Is your feature request related to a problem? Please describe.
Although it is possible to provide keyword arguments that are passed to off-diagonal (by default scatter) plots, it isn't possible to do the same with the diagonal (by default histogram) plots. This makes it hard to set things like the bin size for the histograms.

#### Describe the solution you'd like
An argument for the function, taking a dict, which is set as the keyword arguments for the diagonal plots.  Perhaps `diagonal_kwds` or `diag_kwds`.  So, for example, `hvplot.scatter_matrix(data, diag_kwds={'bins': 100})`. 

#### Describe alternatives you've considered
Manually specifying the options to the grid is possible, but cumbersome since it requires calculating which grid elements are the diagonals.

#### Additional context
The pandas `scatter_matrix` provides separate arguments to specify options for the diagonal and off-diagonal plots.  It uses `hist_kwds` and `density_kwds`, for 'hist' and 'kde' plots, respectively. This seems redundant to me since you an only ever use one or the other.

toddrjen:
This adds a new argument that allows the user to specify keyword arguments for the diagonal plots in `hvplot.plotting.scatter_matrix`.  Fixes #599.

### DIFF
--- a/hvplot/plotting/scatter_matrix.py
+++ b/hvplot/plotting/scatter_matrix.py
@@ -7,7 +7,8 @@ from ..util import with_hv_extension
 
 
 @with_hv_extension
-def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5, **kwds):
+def scatter_matrix(data, c=None, chart='scatter', diagonal='hist', alpha=0.5,
+                   diagonal_kwds=None, **kwds):
     """
     Scatter matrix of numeric columns.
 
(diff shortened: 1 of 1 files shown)


jbednar (approved):
Looks good, thanks! Should we consider also supporting the Pandas `hist_kwds` by code like `if hist_kwds in kwds: diagonal_kwds=kwds.pop('hist_kwds')`? The only reason to do so is if some specific hist_kwds that are valid in Pandas .plot would then be accepted here. If the actual options differ as well, then there'd be no point...

toddrjen:
Alright, I added support for both `hist_kwds` and `density_kwds`.  I also added support for specifying a colormap, which is currently hard-coded and cannot be overridden.  The current colormap is okay for `chart='scatter'` but not good for other `chart` types.

toddrjen:
The test failures don't seem to be related to this PR. Is there anything else anyone thinks I should change?

philippjfr:
Thanks @toddrjen

Issue #603, opened 2021-04-28
handle xarray dataarrays with no coordiantes


rabernat:
#### ALL software version info

hvplot 0.7.1
xarray 0.17.0
python 3.8

#### Description of expected behavior and the observed behavior

Many xarray datasets have labeled dimensions but no coordinate, e.g.

```python
da = xr.DataArray(np.ones((10, 10)), dims=['y', 'x'])
```

Hvplot often produces errors around these datasets that seem unnecessary.

#### Complete, minimal, self-contained example code that reproduces the issue

```
import xarray as xr
import hvplot.xarray

# no coordinates for 2d array: this works fine
da = xr.DataArray(np.ones((10, 10)), dims=['y', 'x'])
da.hvplot.image('x', 'y')

# 3d array with coordinates, also fine
da = xr.DataArray(np.ones((2, 10, 10)), dims=['time', 'y', 'x'],
                  coords={'time': ('time', np.arange(2)),
                          'x': ('x', np.arange(10)),
                          'y': ('y', np.arange(10))})
da.hvplot.image('x', 'y')

# 3d array with some missing coordinates: does not work
da = xr.DataArray(np.ones((2, 10, 10)), dims=['time', 'y', 'x'],
                  coords={'time': ('time', [0, 1])})
da.hvplot.image('x', 'y')
```

In my mind, this should work exactly the same as the prior example with coordinates.

#### Stack traceback and/or browser JavaScript console output

the final example raises

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/srv/conda/envs/notebook/lib/python3.8/site-packages/IPython/core/formatters.py in __call__(self, obj, include, exclude)
    968 
    969             if method is not None:
--> 970                 return method(include=include, exclude=exclude)
    971             return None
    972         else:

/srv/conda/envs/notebook/lib/python3.8/site-packages/holoviews/core/dimension.py in _repr_mimebundle_(self, include, exclude)
   1315         combined and returned.
   1316         """
-> 1317         return Store.render(self)
   1318 
   1319 

/srv/conda/envs/notebook/lib/python3.8/site-packages/holoviews/core/options.py in render(cls, obj)
   1403         data, metadata = {}, {}
   1404         for hook in hooks:
-> 1405             ret = hook(obj)
   1406             if ret is None:
   1407                 continue

/srv/conda/envs/notebook/lib/python3.8/site-packages/holoviews/ipython/display_hooks.py in pprint_display(obj)
[...]

PR #605, opened 2021-05-04
Adding default coordinates to Datasets with missing image coords

jlstevens:
This PR aims to address #603 by generating default image coordinates when they are missing.

Before the PR, using the example given in #603:

With the PR:

So far I've only thought about the `Dataset` branch, to finish this PR I'll need to handle the `DataArray` case as well.

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -1026,6 +1026,12 @@ class HoloViewsConverter(object):
             elif self.datatype == 'xarray':
                 import xarray as xr
                 if isinstance(data, xr.Dataset):
+                    if kind == 'image':
+                        dims_with_coords = set(data.coords.keys())
+                        missing_coords =  set(self.indexes) - dims_with_coords
+                        for missing_coord in missing_coords:
+                            new_coord = np.arange(data[missing_coord].shape[0])
+                            data = data.assign_coords(**{missing_coord:new_coord})
                     dataset = Dataset(data, self.indexes)
                 else:
                     name = data.name or self.label or self.value_label
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -1029,9 +1029,9 @@ class HoloViewsConverter(object):
                     if kind == 'image':
                         dims_with_coords = set(data.coords.keys())
                         missing_coords =  set(self.indexes) - dims_with_coords
-                        for missing_coord in missing_coords:
-                            new_coord = np.arange(data[missing_coord].shape[0])
-                            data = data.assign_coords(**{missing_coord:new_coord})
+                        new_coords = {coord: np.arange(data[coord].shape[0])
+                                      for coord in missing_coords}
+                        data = data.assign_coords(**new_coords)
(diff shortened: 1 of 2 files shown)


jlstevens:
Some datashader related tests seem to be failing on Python 3. Python 2 is passing because:

> test_plot_resolution_0_image (hvplot.tests.testoperations.TestChart2D) ... SKIP: xarray or datashader not available

jlstevens:
The datashader test failures are not being caused by this PR so that is a separate issue to fix. Restoring the original commit.

rabernat:
Amazing, thanks for your work on this!

I would make sure to test many possible combinations here (no coordinates, some coordinates, dask, no dask, different ndims, etc.) The example I reported in #603 was just one fail case. Every different scenario seemed to fail with a slightly different error message, and unfortunately I did not take the time to document all of them.

hvplot/converter.py
```diff
@@ -1026,6 +1026,12 @@ def __call__(self, kind, x, y):
             elif self.datatype == 'xarray':
                 import xarray as xr
                 if isinstance(data, xr.Dataset):
+                    if kind == 'image':
+                        dims_with_coords = set(data.coords.keys())
+                        missing_coords =  set(self.indexes) - dims_with_coords
+                        for missing_coord in missing_coords:
+                            new_coord = np.arange(data[missing_coord].shape[0])
+                            data = data.assign_coords(**{missing_coord:new_coord})
```
philippjfr, line 1034:
```suggestion
                        missing_coords =  set(self.indexes) - dims_with_coords
                        new_coords = {coord: np.arange(data[coord].shape[0])
                                                  for coord in missing_coords}
                        data = data.assign_coords(**new_coords)
```

hvplot/converter.py
```diff
@@ -1026,6 +1026,12 @@ def __call__(self, kind, x, y):
             elif self.datatype == 'xarray':
                 import xarray as xr
                 if isinstance(data, xr.Dataset):
+                    if kind == 'image':
```
philippjfr, line 1029: How about quadmeshes?

philippjfr: And contours and contourf?

jlstevens: Yup, all need to be handled. PR is still WIP and not ready for review.

jlstevens:
> I would make sure to test many possible combinations here (no coordinates, some coordinates, dask, no dask, different ndims, etc.) 

Absolutely! I'll make sure there are unit tests for these cases.

setup.py
```diff
@@ -31,7 +31,7 @@ def get_setup_version(reponame):
     'bokeh >=1.0.0',
     'colorcet >=2',
     'holoviews >=1.11.0',
-    'pandas',
+    'pandas <=1.2.2',
```
philippjfr, line 34:
```suggestion
    'pandas',
```

philippjfr:
Going to merge this as a first cut, please do follow up on all the other cases but I'm releasing 0.7.2 tonight so wanted to get this in.

PR #651, opened 2021-08-16
Improving the hvPlot UI

philippjfr:
This PR is working towards getting the hvPlot UI into a state where we can release it.

- [x] Separate DataFrame, Geometry and Gridded explorers while keeping `hvPlotExplorer` base class
- [ ] Expose all axis options
- [ ] Expose all style options
- [x] Expose all geo options
- [x] Expose all datashader options
- [ ] Expose all style mapping options

### DIFF
--- a/hvplot/ui.py
+++ b/hvplot/ui.py
@@ -1,24 +1,49 @@
-
+import numpy as np
 import panel as pn
 import param
 
+from holoviews.core.util import max_range
 from holoviews.plotting.util import list_cmaps
 from panel.viewable import Viewer
 
 from .converter import HoloViewsConverter as hvConverter
+from .util import is_geodataframe, is_xarray
 
-
-kinds = set(hvConverter._kind_mapping) - set(hvConverter._gridded_types)
-COLORMAPS = [cm for cm in list_cmaps() if not cm.endswith('_r_r')]
-MAX_ROWS = 10000
+# Defaults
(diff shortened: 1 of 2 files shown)


hvplot/ui.py
```diff
@@ -95,66 +197,133 @@ class hvPlotExplorer(Viewer):
...
+    style = param.ClassSelector(class_=Style)
+
+    def __new__(cls, data, **params):
+        if is_geodataframe(data):
+            cls = hvGeomExplorer
+        elif is_xarray(data):
+            cls = hvGridExplorer
+        else:
+            cls = hvDataFrameExplorer
+        return cls(data, **params)
```
jbednar, line 219: This code answers my immediate reaction to "Separate DataFrame, Geometry and Gridded explorers while keeping hvPlotExplorer base class", i.e., to ask for a factory that returns something of the correct type.  One step ahead of me!

Still, if hvGeomExplorer is not a subclass of hvDataFrameExplorer, it seems problematic to switch between those two here.  Conceptually, can't any dataframe become a geo dataframe (conceptually) by tacking on a geometry column?  And so shouldn't the capabilities of hvGeomExplorer include _all_ the capabilities of hvDataFrameExplorer, plus a few geo-related ones? I can't tell much from the implementations of these two classes about what their future is, but it seems at first glance that people could add functionality to hvDataFrameExplorer that would then not be available in hvGeomExplorer, which doesn't seem right to me.

PR #653, opened 2021-08-24
Implement backend options transformer

philippjfr:
Enables plotting hvPlot output with HoloViews Matplotlib and Plotly backends.

[Matplotlib](https://anaconda.org/philippjfr/plotting_matplotlib/notebook)
[Plotly](https://anaconda.org/philippjfr/plotting_plotly/notebook)


philippjfr:
This is now in a reasonable state but the list of transforms is not at all exhaustive. So ass you take over @maximlt here's what needs to be done.

Basically I've transformed some of the most important options but there is a **long** tail of options which are currently not handled. As a start it would be great if you could go through all user guides and reference notebooks and visually compare the plotly and matplotlib output with the bokeh output. Another useful approach is to insert this print statement:

```python
if opt not in el_opts.allowed_keywords:
    print(f'{opt} not supported')
    ...
```

and see which options are currently being skipped and either make sure they are also UNSET (if not supported) or define the appropriate transform to ensure they are supported.

maximlt:
Here is a small write-up to clarify our intent here. As of now `hvplot` has a unique backend which is `bokeh`. hvplot's API allows to pass in kwargs to customize a bokeh plot, on the condition that they are known to holoviews. For instance to plot a dashed line one should provide the bokeh option `line_dash="dashed"`. These options can be found by tab completion or with `hvplot.help('line')`.

hvplot relies obviously on holoviews which itself supports two additional backends: matplotlib and plotly. It seems natural then to add support in hvplot for these alternative backends. There's however code out there that surely relies on customizing plots with bokeh kwargs. Where this happens, hvplot should try its best to transform the bokeh options provided to options understood by the targeted backend. For instance, the `line_dash` equivalent option for Matplotlib would be `linestyle="dashed"`. Under the hood a mapping is used to store the options that can be linked or not between bokeh and matplotlib/plotly (TBD), and when they can be linked a transform function.

This allows for instance a user to turn a notebook with bokeh plots  to a notebook with matplotlib plots, while partly preserving the plot options originally defined for the bokeh backend. Note that this is not a dead-end, the underlying plot object can be accessed with `hv.render` and further customized.

```python
# These two lines change the backend to matplotlib
import holoviews as hv
hv.extension('matplotlib')

import hvplot.pandas

df = ...
plot = df.hvplot.line(x='x', y='y', line_dash='dashed')
# Display plot in a cell
plot

# Get the Matplotlib figure to further customize it if required
fig = hv.render(plot)
...
fig.savefig('fig.png')
```

What has not yet been done in this work is to directly deal with matplotlib or plotly options, e.g. what to do with `df.hvplot.line(linestyle="dashed")`. This could be supported in principle. As of now this would result in a warning being emitted due to an option unknown to hvplot being used.

maximlt:
Here's a list of the current issues observed while running all the notebooks with the PR in its current state.

### General:

  * date ticks are usually too long and not readable since partially overlapping
  * Can't get the `line_width` option to work

### User guides

* Plotting:
    * Bar plot with two y variables -> Display 'Year, Variable' in the x label ('Year' being x here), 'Variable' shouldn't be displayed there
    * hexbin: do not display an image, returns `:HexTiles   [airtime,arrdelay]`, note this uses a dask df
    * heatmap: display the annotations by default (default in holoviews)
    * hist: error with `by=...` but that could also be because of dask again
    * box: figure with no plot when `by=...`, dask again
    * datashader: lots of unecessary space around the plot
    * violin: displaying the axes only, but again could be because of dask
* Pandas API:
    * scatter: no colormap visible when `c='whatever'`
    * scatter_matrix, andrew_curves, parallel_coordinates: not working, options are not transformed in the PR
    * lag plots: no legend displayed by default while there's one with bokeh
    * tick formatter: `ValueError: ClassSelector parameter 'xformatter' value must be an instance of (str, Formatter, function), not DatetimeTickFormatter(id='1002', ...)`.
    * error bars: not working well at all, the error lines are next to the bars
    * colormap: `colormap='Greens'` not working, colormap not respected
* streams: not working `df.groupby('y').sum().hvplot.bar(x='y')` leads to a ZeroDivisionError
* subplots: grid size really small and cannot be increased
* time series: violin matplotlib plots have a cycled colormap while bokeh plots are just blue
* geographic: spatialpandas plot upside down

### Reference

* heatmap: annotations displayed by default
* ohlc: not working at all
* violin: cmap not respected
* contour and contourf: not working at all (clim=(None, None), is that expected?)
* image: cartopy hangs when downloading the coastline shapefile (this is being fixed by cartopy)
* quadmesh: the layout is kind of weird, too much padding around the plot
* geopandas points: Issue with tiles: `'No plotting class for Tiles type and matplotlib backend found'`

maximlt:
@philippjfr about the issue with `line_width` not being taken into account, I think I found why but I'm not sure what's the appropriate fix.

At this line
https://github.com/holoviz/hvplot/blob/483fe175cdf38102cda5b6fc060621ebf12fc63f/hvplot/converter.py#L958
the options I get for a `Curve` plot when I have declared that I want a Matplotlib plot (`import holoviews as hv; hv.extension('matplotlib'); import hvplot.pandas`) are:
```python
OrderedDict([('axiswise', False),
             ('color', Cycle()),
             ('framewise', True),
             ('height', 300),
             ('labelled', ['x', 'y']),
             ('line_width', 20),
             ('linewidth', 2),
             ('logx', False),
             ('logy', False),
             ('muted_alpha', 0.2),
             ('responsive', False),
             ('shared_axes', True),
             ('show_grid', False),
             ('show_legend', True),
             ('tools', ['hover']),
             ('width', 700)])
```
They **include** `linewidth=2`, which is the default option for a `Curve` for the Matplotlib backend in holoviews:
https://github.com/holoviz/holoviews/blob/5e058cadb7c9817216e6eeafb2588ee5c5a5bc8c/holoviews/plotting/mpl/__init__.py#L231

Is it expected that at this stage `.get()` is able to return options from both backends>

maximlt:
In the *Plotting* guide the call to create a Hexbin plot (`flights.hvplot.hexbin(x='airtime', y='arrdelay', width=600, height=500, logz=True)`) doesn't display any image, it just returns `:HexTiles [airtime,arrdelay]`.

Since the following code correctly displays an image with the Matplotlib backend I assume this is a **hvplot** issue.
```python
hv.HexTiles(flights, kdims=['airtime', 'arrdelay']).opts(fig_size=200, logz=True, colorbar=True)
```

The object returned by hvplot has this `.dataset` property:
```
:Dataset   [index]   (year,month,day,dayofweek,dep_time,crs_dep_time,arr_time,crs_arr_time,carrier,flight_num,tail_num,actual_elapsed_time,crs_elapsed_time,airtime,arrdelay,depdelay,origin,dest,distance,taxi_in,taxi_out,cancelled,cancellation_code,diverted,carrier_delay,weather_delay,nas_delay,security_delay,late_aircraft_delay)
```
while the one returned by holoviews has:
```
:Dataset   [airtime,arrdelay]   (year,month,day,dayofweek,dep_time,crs_dep_time,arr_time,crs_arr_time,carrier,flight_num,tail_num,actual_elapsed_time,crs_elapsed_time,depdelay,origin,dest,distance,taxi_in,taxi_out,cancelled,cancellation_code,diverted,carrier_delay,weather_delay,nas_delay,security_delay,late_aircraft_delay)
```

maximlt:
Executing this:
```python
fig = airports.hvplot.points('Longitude', 'Latitude', geo=True, color='red', alpha=0.2,
                       xlim=(-180, -30), ylim=(0, 72), tiles='ESRI')
```
works correctly with the Bokeh backend but fails with a horrible error message (https://github.com/holoviz/geoviews/issues/540) with the Matplotlib one.

Interestingly `xlim` and `ylim` appear to have been projected to Web Mercartor, while the *crs* of the Points element is still PlateCarree.
```
:Overlay
   .Tiles.I  :Tiles   [x,y]
   .Points.I :Points   [Longitude,Latitude]
    | Options(alpha=0.2, axiswise=False, color='red', data_aspect=1, framewise=True,
    |         legend_position='right', logx=False, logy=False, muted_alpha=0.2, responsive=False,
    |         shared_axes=True, show_frame=True, show_grid=False, show_legend=True,
    |         size=5.477225575051661, tools=['hover'], xlim=(-19981848.597392607, -3339584.8351176973),
    |         ylim=(0.1113194902619854, 11753184.255101029))
```

Anyway, `Tiles` elements don't have a matplotlib implementation in holoviews, so no background tiles would be plotted.

One way to adapt the code to support plotting background tiles with the Matplotlib backend may be to use the WMTS elements provided by geoviews instead of the Tiles elements provided by holoviews, given that the latter have no projection.

maximlt:
Plotly issues:
- [ ] The legend is (always?) shown while it's not with Bokeh or Matplotlib => Opened a Holoviews issue (https://github.com/holoviz/holoviews/issues/5217)

maximlt:
Before the backend option transforms work hvPlot users could already pass Bokeh options as kwargs such as `line_dash` in `df.hvplot.line(x='x', y='y', line_dash='dashed')`, without hvPlot knowing anything about the `line_dash` option.

After this work, when a user selects another backend such as Matplotlb and calls `df.hvplot.line(x='x', y='y', line_dash='dashed')`, if `line_dash` is part of hvPlot's internal transforms mapping then it will be converted (the key or the value or both) to a valid option definition for the targeted backend. If the option isn't part of the mapping then it'll be ignored. In that case `line_dash` is part of the Matplotlib mapping and would be converted to `linestyle='dashed'`. So the advantage of this system is that it facilitates executing existing code with another backend. If some Bokeh options were passed in that code as kwargs they may or may not be used by the backend (this depends on whether this option is actually available in the targeted backend and also on whether I've mapped it or missed it).

Now let's imagine a new hvPlot user who doesn't care at all about Bokeh plots but wants to generate Matplotlib plots only, and wants a dashed line. If this user knows Matplotlib's API a little then the natural thing to do would be  `df.hvplot.line(x='x', y='y', linestyle='--')`. This isn't going to work though, the plot will be displayed but with a warning: `WARNING:param.main: linestyle option not found for line plot; similar options include: []`. `hvplot.help('line')` will list the `line_dash` option and so once the user has figured out that it is a Bokeh option then they'll have to go to Bokeh's site (or maybe Holoviews?) to see what values `line_dash` accepts. To avoid this pattern which feels a little awkward, what should we recommend in the docs for now? Should we, for now, recommend the usage of `.opts()` for matplotlib&plotly and link to the relevant section of the HoloViews docs?

One subtlety about the option transforms is that when a transform exists for an option (e.g. `alpha` to `opacity` for plotly), in most cases the value isn't validated, it's just passed internally to the backend via `.opts`. Not validating the value means that a user could provide a combination of a valid bokeh option with a value that would be invalid for bokeh but valid for the backend. For example, a user could select the plotly backend and provide `color='hsl(0,100%,50%)` (red color) this `hsl(...)` value being valid for plotly. However switching to bokeh, this code would no longer work and raise an error as this sort of value isn't accepted by bokeh for defining the color. What should we do to avoid this? I would suggest documenting that only valid Bokeh option values are supported for now. And is there a way in holoviews that we could leverage to validate an option value before attempting to render the figure? (I assume not and that it's the responsibility of the backend libraries only)

jlstevens:
Maybe I am missing something but couldn't `linestyle='--'` (the matplotlib kwarg) just be passed through to matplotlib when the matplotlib backend is active, just the same way  `line_dash` gets passed to Bokeh without hvPlot knowing anything about it?

The issue with different backends accepting different types of values is tricky at the holoviews level as well, when trying to switch backends. I'm not sure it is a great idea but if we declare multiple backends are in use (as holoviews does is when loading the extension), keywords not recognized by the current backend could just be ignored instead of generating errors. e.g. `df.hvplot.line(x='x', y='y', linestyle='--', line_dash='dashed')` would be allowed. This *could* allow some more flexibility to switch backend but with a flat kwarg namespace you are out of luck if shared keywords with different value formats clash. Another type of solution would be something like `df.hvplot.line(x='x', y='y', matplotlib_kwargs={...}, bokeh_kwargs={...)` of course...

jbednar:
> Before the backend option transforms work hvPlot users could already pass Bokeh options as kwargs such as line_dash in df.hvplot.line(x='x', y='y', line_dash='dashed'), without hvPlot knowing anything about the line_dash option.

Can you be more specific about what previously happened to `line_dash` in this case? Did it become a HoloViews plot option, or was it somehow kept around to be applied later as a Bokeh plot hook?

> Maybe I am missing something but couldn't linestyle='--' (the matplotlib kwarg) just be passed through to matplotlib when the matplotlib backend is active, just the same way line_dash gets passed to Bokeh without hvPlot knowing anything about it?

That does sound reasonable to me, if that's how it works.

> Should we, for now, recommend the usage of .opts() for matplotlib&plotly and link to the relevant section of the HoloViews docs?

I would definitely try to avoid using .opts() wherever practical, in order to minimize the number of concepts needed for most people to get most of their work done, and to make it easier to switch from .plot to .hvplot.

> For example, a user could select the plotly backend and provide color='hsl(0,100%,50%) (red color) this hsl(...) value being valid for plotly. However switching to bokeh, this code would no longer work and raise an error as this sort of value isn't accepted by bokeh for defining the color. What should we do to avoid this?

I don't think we should work hard to avoid this. By default, if a user does that, then they've tied themselves to a particular plotting library, which seems fine by me! If a user did want to avoid that, they could follow @jlstevens' suggestion of being explicit with`df.hvplot.line(x='x', y='y', matplotlib_kwargs={...}, bokeh_kwargs={...}, ...)` to sort things out by backend. But very few users seem to want to dynamically switch backends, even on HoloViews where the options can be sorted out cleanly, so I don't expect such options to be used much.

> And is there a way in holoviews that we could leverage to validate an option value before attempting to render the figure? (I assume not and that it's the responsibility of the backend libraries only)

At least for now I'm happy to let the backend library do that validation. Eventually, sure, eager validation is more helpful for the user, but since few people will be creating plottable objects but not plotting them, validation should eventually happen already.

maximlt:
> Maybe I am missing something but couldn't linestyle='--' (the matplotlib kwarg) just be passed through to matplotlib when the matplotlib backend is active, just the same way line_dash gets passed to Bokeh without hvPlot knowing anything about it?

I wasn't quite clear when I wrote that hvPlot doesn't know anything about `line_dash`. `line_dash` isn't part of the API and not mentioned in any of the options available in hvPlot. However hvPlot will check that the options passed are valid options for the Bokeh backend and so at this moment it knows that `line_dash` is OK.

Basically how things work right now is that it's always assumed internally that the options passed are Bokeh options. It looks something like that:
```python

user_opts = {'line_dash': 'dashed'}
new_backend = 'matplotlib'

import holoviews as hv

hv.extension('bokeh')
hv.extension(new_backend)

# In hvplot the `_process_style` method is executed before setting the options
# on the element to parse the user options. It assumes the backend is bokeh.
# It will happily parse and let `line_dash` go further but will "block" `linestyle`
# since it's not one of the allowed options of the bokeh backend.

el = hv.Curve([[0,1], [1, 2]]).opts(backend='bokeh', **user_opts)

def transfer_opts(element, backend):
    bokeh_opts = el.opts.get(backend='bokeh', defaults=False)
    # Convert the bokeh options to another backend options using an internal mapping
    new_opts = {'linestyle': '--'}  # Just to get this snippet running
    return element.opts(**new_opts, backend=backend)

el = el.apply(transfer_opts, backend=new_backend)
```

The last line in the snippet is the command that is the last one executed in the code, before that everything is in Bokeh's world until it's converted into another backend world.

maximlt:
Philipp also designed another feature in this PR that I think is worth mentioning. A user can pass a `backends` parameter that allows to convert the options for all the backends requested, so that the object can be rendered using any of the available backends:
```python
import hvplot.pandas
# load data ...
hvplot.extension('plotly', 'matplotlib')
o = crime.hvplot.line(x='Year', y='Violent Crime rate', line_dash='dashed', backends=['plotly', 'matplotlib'])
# Now the object has the `line_dash` option converted to all the backends
# And this returns a matplotlib figure with a dashed line!
hvplot.render(o, backend='matplotlib')
```

maximlt:
> Maybe I am missing something but couldn't linestyle='--' (the matplotlib kwarg) just be passed through to matplotlib when the matplotlib backend is active, just the same way line_dash gets passed to Bokeh without hvPlot knowing anything about it?

I guess this should be possible but I don't know how tricky that'd be to implement, it might be that there are some things deeply tied to the Bokeh backend and its options (`_process_style` is about a 100 LOC). I wouldn't like this to block this PR any longer :/ What are your thoughts on this @philippjfr ? In the mean time I'll try to get the tests to pass and fix the docs build.

maximlt:
Setting pandas backend to `holoviews` currently defaults to Bokeh.
https://github.com/holoviz/hvplot/blob/0cf0d867cbb1548fa144f34f216482178f27340e/setup.py#L137-L143

Should we provide a way to set pandas backend to another extension (e.g. `'holoviews-matplotlib'`)?

maximlt:
A default Bokeh renderer is instantiated in `hvplot.utilities` and exposed in the main `__init__.py` file.
https://github.com/holoviz/hvplot/blob/0cf0d867cbb1548fa144f34f216482178f27340e/hvplot/utilities.py#L4

I don't see this documented anywhere. In this PR I've exposed `holoviews.render` which can be used as `hvplot.render(...)`. Could we deprecate `renderer` and remove it in a future version?

maximlt:
Some work's required before this is merged probably to refine how the size/aspect options are handled with the other extensions:
* hvplot sets default width/height options for bokeh plots, should these defaults be kept for matplotlib and plotly?
* in the current state of the PR `width` and `height` are supported by all the extensions even without compatibility. This is because they're part of the options that are transformed (there's some special logic to convert width/size to matplotlib holoviews plotting options)
* not sure what the current state of matplotlib is about setting the size but I guess it's done with `figsize` (*Width, height in inches.*), which is what pandas.plot() accepts. Note that this not an option directly supported by holoviews (which supports `fig_inches` and `fig_size` and others).

I'd suggest:
* `width` and `height` be the declared  hvplot's way of setting a plot size
* have special logic to convert `figsize` if provided to get closer to pandas' plot API

maximlt:
> `width` and `height` be the declared hvplot's way of setting a plot size

This is in fact already the case, since `width` and `height` are part of the Generic Options.

> have special logic to convert `figsize` if provided to get closer to pandas' plot API

Actually there was already some sort of handling of `figsize`. There's a warning (that should be emitted but is not, I'll fix that in this PR) that indicates that there's no concept of a Figure in hvPlot:

```python
        if 'figsize' in mismatches:
            mismatches.pop(mismatches.index('figsize'))
            param.main.param.warning('hvPlot does not have the concept of a figure, '
                               'and the figsize keyword will be ignored. The '
                               'size of each subplot in a layout is set '
                               'individually using the width and height options.')
```

This makes sense, as `figsize` for Matplotlib controls the size of the whole Figure/Layout while width/height in hvPlot controls the size of each individual (sub)plot. We could just ignore `figsize` in the case of subplots but I tend to think that for the scope of this PR we should just stop there, and improve things when we get some feedback.

maximlt:
@philippjfr @jbednar this is ready for a review.

hvplot/utilities.py
```diff
@@ -24,3 +25,38 @@ def show(obj, title=None, port=0, **kwargs):
         raise ValueError('%s type object not recognized and cannot be shown.' %
                          type(obj).__name__)
     _pn.pane.HoloViews(obj).show(title, port, **kwargs)
+
+
+class hvplot_extension(_hv.extension):
```
philippjfr, line 30: Hmm, how would you use this together with the import based extension loading mechanism? Does it end up embedding the JS/CSS twice?

hvplot/utilities.py
```diff
@@ -24,3 +25,38 @@ def show(obj, title=None, port=0, **kwargs):
...
+    _compatibility = None
+
+    def __call__(self, *args, **params):
+        # importing e.g. hvplot.pandas always loads the bokeh extension.
+        # so hvplot.extension('matplotlib', compatibility='bokeh') doesn't
+        # require the user or the code to explicitely load bokeh.
+        compatibility = params.pop('compatibility', None)
+        super().__call__(*args, **params)
+        backend = _hv.Store.current_backend
+        if compatibility in ['matpliotlib', 'plotly'] and backend != compatibility:
```
philippjfr, line 48:
```suggestion
        if compatibility in ['matplotlib', 'plotly'] and backend != compatibility:
```

maximlt:
@jlstevens could you please review the last changes wrt compatibility?

hvplot/utilities.py
```diff
@@ -31,9 +31,8 @@ def show(obj, title=None, port=0, **kwargs):
 
 class hvplot_extension(_hv.extension):
 
-    compatibility = param.Selector(
-        default=None, objects=['bokeh', 'matplotlib', 'plotly'],
-        allow_None=True, doc="""
+    compatibility = param.ObjectSelector(
+        allow_None=True, objects=['bokeh', 'matplotlib', 'plotly'], doc="""
             Backend used to parse styling options.""")
```
jlstevens, line 36: Maybe from a user perspective it should be something like 'Plotting library API used to process extra keyword arguments' (or similar)?

jlstevens:
Other than the suggestion to update the parameter docstring, the latest changes look fine to me!

examples/homepage.ipynb
```diff
@@ -160,6 +160,8 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "hvPlots will show widgets like the \"Time\" slider here whenever your data is indexed by dimensions that are not mapped onto the plot axes, allowing you to explore complex datasets much more easily than with the default plotting support.\n",
```
jlstevens, line 163: Would end this sentence with `allowing you to explore complex datasets much more easily.` (not need for the rest imho).

examples/homepage.ipynb
```diff
@@ -181,12 +183,44 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "hvPlots will show widgets like the \"Time\" slider here whenever your data is indexed by dimensions that are not mapped onto the plot axes, allowing you to explore complex datasets much more easily than with the default plotting support.\n",
-    "\n",
+    "hvPlot offers the possibility to create [Matplotlib](https://matplotlib.org/) and [Plotly](https://plotly.com/) plots. Switching the underlying plotting library/backend is achieved with the `extension` function."
```
jlstevens, line 186: Isn't 'switching' is done by `output`? I would say something like `Load the chosen plotting library with the extension function.` to be more accurate.

examples/reference/pandas/lagplot.ipynb
```diff
@@ -0,0 +1,66 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "np.random.seed(5)\n",
+    "x = np.cumsum(np.random.normal(loc=1, scale=5, size=50))\n",
+    "s = pd.Series(x, name='Time series')\n",
+    "\n",
+    "s.hvplot()"
```
jlstevens, line 38: Don't think this is really related to this PR, but the lack of an x-axis label bothers me. What are the units? Absolute time or presumably time deltas?

examples/reference/pandas/ohlc.ipynb
```diff
@@ -93,29 +93,9 @@
...
-   "pygments_lexer": "ipython3",
-   "version": "3.8.8"
-  },
-  "widgets": {
-   "application/vnd.jupyter.widget-state+json": {
-    "state": {},
-    "version_major": 2,
-    "version_minor": 0
-   }
+   "pygments_lexer": "ipython3"
```
jlstevens, line 98: Eesh! Good to see this removed.

examples/user_guide/Introduction.ipynb
```diff
@@ -6,15 +6,18 @@
...
+    "* [Streamz](https://streamz.readthedocs.io): DataFrame(s), Series(s) (streaming columnar data)\n",
+    "* [Intake](https://github.com/ContinuumIO/intake): DataSource (data catalogues)\n",
+    "* [GeoPandas](https://geopandas.org): GeoDataFrame (geometry data)\n",
+    "* [NetworkX](https://networkx.github.io/documentation/stable/): Graph (network graphs)\n",
     "\n",
     "Many of these libraries have the concept of a high-level plotting API that lets a user generate common plot types very easily. The native plotting APIs are generally built on [Matplotlib](http://matplotlib.org), which provides a solid foundation, but means that users miss out the benefits of modern, interactive plotting libraries for the web like [Bokeh](http://bokeh.pydata.org) and [HoloViews](http://holoviews.org).\n",
     "\n",
-    "hvPlot provides a high-level plotting API built on HoloViews and Bokeh that provides a general and consistent API for plotting data in all the abovementioned formats.\n",
+    "**hvPlot** provides a high-level plotting API built on HoloViews that provides a general and consistent API for plotting data in all the abovementioned formats.\n",
```
jlstevens, line 20:
```suggestion
    "**hvPlot** provides a high-level plotting API built on HoloViews that provides a general and consistent API for plotting data in all the formats mentioned above.\n",
```

Hope this suggestion doesn't mess up the notebook JSON!

examples/user_guide/Introduction.ipynb
```diff
@@ -128,13 +136,32 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "hvplot.extension('matplotlib')\n",
+    "\n",
+    "df.hvplot()"
```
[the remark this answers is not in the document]
jlstevens: Also, I would add a sentence pointing people to the user guide where they can get more information about working with multiple plotting backends.

examples/user_guide/Plotting_Extensions.ipynb
```diff
@@ -0,0 +1,205 @@
...
+   "outputs": [],
+   "source": [
+    "violent_crime.opts.info()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You can see that `line_dash='dashed'` has been internally converted to `linestyle='dashed'` that is a valid option for the Matplotlib plotting extension. Note however that all the Bokeh options haven't yet been mapped or can't be converted to options suited for Matplotlib or Plotly, in that case they will be ignored."
```
jlstevens, line 131: I would say something like 'Note that support for all Bokeh options is currently incomplete and will continue to be improved while equivalent support for Plotly and Matplotlib is planned for a future release'

examples/user_guide/Plotting_Extensions.ipynb
```diff
@@ -0,0 +1,205 @@
...
+   "outputs": [],
+   "source": [
+    "from plotly.graph_objects import Figure\n",
+    "\n",
+    "plotly_fig = hvplot.render(violent_crime, backend='plotly')\n",
+    "fig = Figure(plotly_fig).update_layout(title='Violent crime')\n",
+    "fig.show()"
+   ]
+  }
+ ],
```
jlstevens, line 196: Would be nice to end the notebook with a sentence or two as a conclusion.

examples/user_guide/Plotting_with_Matplotlib.ipynb
```diff
@@ -0,0 +1,672 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As we discovered in the [Introduction](Introduction.ipynb), HoloViews allows plotting a variety of data types. Here we will use the sample data module and load the pandas and dask hvPlot API:"
```
jlstevens, line 7: I would add a sentence along the lines of 'This notebook demonstrates the use of the Matplotlib plotting backend, the equivalent notebook demonstrating the Plotly backend may be found HERE (link)'

examples/user_guide/Plotting_with_Plotly.ipynb
```diff
@@ -0,0 +1,672 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "As we discovered in the [Introduction](Introduction.ipynb), HoloViews allows plotting a variety of data types. Here we will use the sample data module and load the pandas and dask hvPlot API:"
```
jlstevens, line 7: Same as for matplotlib. One sentence should be enough to link the contents of the notebook to the notebook title.

jlstevens:
Made some suggestions for the docs but overall this looks good!

examples/reference/pandas/lagplot.ipynb
maximlt: This is the exact same example as pandas' one, which doesn't mean it's perfect!
https://pandas.pydata.org/docs/reference/api/pandas.plotting.lag_plot.html

examples/user_guide/Introduction.ipynb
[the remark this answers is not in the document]
maximlt: Rotated by 30 degrees.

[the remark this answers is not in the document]
maximlt: > Also, I would add a sentence pointing people to the user guide where they can get more information about working with multiple plotting backends (Plotting Extensions).

Thanks, added that!

maximlt:
@jlstevens thanks for the review! I applied your suggestions in the last two commits.

jlstevens:
Tests look like they are passing and I'm happy to see it merge from my side!

PR #659, opened 2021-09-10
Add a title to xarray plots with scalar coords

### LINKED ISSUE #656: move scalar coords to title in hvplot.xarray
#### Is your feature request related to a problem? Please describe.
Upstreaming xarray .plot() methods to hvplot.xarray

#### Describe the solution you'd like
an xarray.Dataset with one data variable:
`ds["data_var"].plot()`
to match
`ds,hvplot()`

#### Additional context
```
import xarray as xr
import hvplot.xarray

ds = xr.tutorial.open_dataset("air_temperature").isel(lat=0, lon=0)
ds["air"].plot()
```

Screen Shot 2021-09-06 at 10 40 38 PM

```
ds.hvplot()
```
Screen Shot 2021-09-06 at 10 41 41 PM

maximlt:
Fixes #656 

Plots based on xarray objects that have scalar coords (e.g. `da.isel(lon=0)`) will get an informative title as one would get with xarray.plot() API. The generated title is equal to xarray's title since internally xarray's formatting utilities are used. This is part of their private API but this code hasn't changed in years.

There are a couple of differences with xarray:
1. If the data has more than two dimensions it will default to a histogram without providing it further hints. In that case xarray.plot() just sets `'Histogram'` as a title, while hvplot() will set the informative title (e.g. `lon = 250`).
2. Whenever widgets are generated to explore the data (groupby op) the title is not changed, it is a title generated by holoviews that is in sync with the widget's values and displays them ({label}: {value} {unit}, ...).

```python
import xarray as xr
import hvplot.xarray  # noqa

ds_air = xr.tutorial.open_dataset("air_temperature")

# DataArray

air = ds_air.air

# Same as before, the title reflects the widgets' state
air.isel().hvplot.line()

# New title: lat = 75.0, lon = 200.0
air.isel(lat=0, lon=0).hvplot.line()

# New title: lon = 285.0
air.sel(lat=[20, 40, 60], lon=285).plot.line(hue='lat')

# New title: time = 2013-06-01T12:00:00'
air.sel(time='2013-06-01 12:00').hvplot()

# New title: lon = 285.0
air.sel(lat=[25, 50, 75], lon=285).hvplot.kde('air', alpha=0.5)

# Dataset

# New title that includes already existing scalar coords and those
# obtained by the iselection 
ds_roms.isel(s_rho=0, ocean_time=0).hvplot.scatter(x='salt', y='zeta')
```

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -498,7 +498,7 @@ class HoloViewsConverter(object):
             plot_opts['global_extent'] = global_extent
         if projection:
             plot_opts['projection'] = process_crs(projection)
-        title = title if title else getattr(self, '_title', None)
+        title = title if title is not None else getattr(self, '_title', None)
         if title is not None:
             plot_opts['title'] = title
         if (self.kind in self._colorbar_types or self.rasterize or self.datashade or self._color_dim):
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -655,6 +655,9 @@ class HoloViewsConverter(object):
         elif is_xarray(data):
             import xarray as xr
             z = kwds.get('z')
+            if isinstance(data, xr.Dataset):
+                if len(data.data_vars) == 0:
+                    raise ValueError("Cannot plot an empty xarray.Dataset object.")
             if z is None:
                 if isinstance(data, xr.Dataset):
                     z = list(data.data_vars)[0]
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -498,6 +498,7 @@ class HoloViewsConverter(object):
             plot_opts['global_extent'] = global_extent
         if projection:
             plot_opts['projection'] = process_crs(projection)
+        title = title if title else getattr(self, '_title', None)
         if title is not None:
             plot_opts['title'] = title
         if (self.kind in self._colorbar_types or self.rasterize or self.datashade or self._color_dim):
@@ -703,6 +704,15 @@ class HoloViewsConverter(object):
 
             if groupby:
                 groupby = [g for g in groupby if g not in grid]
+
+            if not groupby and not grid:
+                # Set a title for coords of size 1 (e.g. da.isel(lon=0).hvplot())
+                if isinstance(da, xr.DataArray):
+                    self._title = da._title_for_slice()
+                elif isinstance(da, xr.Dataset):
+                    one_var = list(da.data_vars)[0]
+                    self._title = da[one_var]._title_for_slice()
+
             self.data = data
         else:
             raise ValueError('Supplied data type %s not understood' % type(data).__name__)
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -708,14 +708,14 @@ class HoloViewsConverter(object):
             if groupby:
                 groupby = [g for g in groupby if g not in grid]
 
+            # Add a title to hvplot.xarray plots that displays scalar coords values,
+            # as done by xarray.plot()  
             if not groupby and not grid:
-                # Set a title for coords of size 1 (e.g. da.isel(lon=0).hvplot())
                 if isinstance(da, xr.DataArray):
                     self._title = da._title_for_slice()
                 elif isinstance(da, xr.Dataset):
-                    one_var = list(da.data_vars)[0]
-                    self._title = da[one_var]._title_for_slice()
-
+                    self._title = partial(xr.DataArray._title_for_slice, da)()
+                        
             self.data = data
         else:
             raise ValueError('Supplied data type %s not understood' % type(data).__name__)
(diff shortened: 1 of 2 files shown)


hvplot/converter.py
```diff
@@ -498,6 +498,7 @@ def __init__(self, data, x, y, kind=None, by=None, use_index=True,
             plot_opts['global_extent'] = global_extent
         if projection:
             plot_opts['projection'] = process_crs(projection)
+        title = title if title else getattr(self, '_title', None)
```
jbednar, line 501: I don't know if numbers are allowed as plot titles, but if they are, seems like this logic will fail if someone passes 0 as the title (e.g. if they wanted to number a bunch of plots).  So maybe:

```suggestion
        title = title if title is not None else getattr(self, '_title', None)
```

That way people can also pass the empty string '' if they want to suppress the auto-generated title.

hvplot/converter.py
```diff
@@ -703,6 +704,15 @@ def _process_data(self, kind, data, x, y, by, groupby, row, col,
...
             if groupby:
                 groupby = [g for g in groupby if g not in grid]
+
+            if not groupby and not grid:
+                # Set a title for coords of size 1 (e.g. da.isel(lon=0).hvplot())
+                if isinstance(da, xr.DataArray):
+                    self._title = da._title_for_slice()
+                elif isinstance(da, xr.Dataset):
+                    one_var = list(da.data_vars)[0]
+                    self._title = da[one_var]._title_for_slice()
```
jbednar, line 714: The DataArray title seems clear, but I'm less sure about the Dataset one:  
- It seems to be assuming that there is always at least one data_var; is that always true? I.e., it's not possible to have a Dataset with coordinates but no data values? If so, fine; if not, should first check that the list is nonempty.
- Why does it only use the title from the first data_var? Is it known that will be the only one plotted?  Or should it concatenate all of the titles? 

If it's clear that there is always exactly one data_var, then these considerations don't apply but there should be a comment explaining why that's a safe assumption.

jbednar (commented):
Thanks!!

Looks good. Could use some tests and also a reproducible example here in the PR's markdown.

> This is part of their private API but this code hasn't changed in years.

Seems ok; if that interface does change I guess we could copy that code (and its associated license) in here, but for now at least it seems like a better bet that if xarray changes how it construct those titles, we'd want to inherit those changes, and so it seems better to reply on the private API than to copy it here.

> If the data has more than two dimensions it will default to a histogram without providing it further hints. In that case xarray.plot() just sets 'Histogram' as a title, while hvplot() will set the informative title (e.g. lon = 250).

The hvPlot behavior sounds more useful to me; feel free to propose that xarray change to match!

> Whenever widgets are generated to explore the data (groupby op) the title is not changed, it is a title generated by holoviews that is in sync with the widget's values and displays them 

Makes sense. When there are widgets, it's important to be able to see those widget values reflected in the title, so that we can be sure the plot actually does reflect the current widget settings. And if there are e.g. many, many dimensions but only a couple of widgets, we'd want the widget values in the title (as that's what's being explored), not the long list of other static dimension values. When there are no widgets, the only informative thing is the list of other static dimension values (long or not), so it seems reasonable to display that as a default. So I _think_ I agree with this behavior, though I suppose I'd have to see how it worked in realistic situations to be sure my intuition is correct.

maximlt:
> Looks good. Could use some tests and also a reproducible example here in the PR's markdown.

I'll add a few tests and will edit the PR's message with an example.

> The hvPlot behavior sounds more useful to me; feel free to propose that xarray change to match!

Yes I'll open an issue on their side. I also noticed that their code to format numbers actually doesn't work.

> Makes sense. When there are widgets, it's important to be able to see those widget values reflected in the title, so that we can be sure the plot actually does reflect the current widget settings. And if there are e.g. many, many dimensions but only a couple of widgets, we'd want the widget values in the title (as that's what's being explored), not the long list of other static dimension values. When there are no widgets, the only informative thing is the list of other static dimension values (long or not), so it seems reasonable to display that as a default. So I _think_ I agree with this behavior, though I suppose I'd have to see how it worked in realistic situations to be sure my intuition is correct.

Yes I agree with you, it's important to have the widgets values reflected in the title, even if it's a little redundant.

hvplot/converter.py
maximlt: > It seems to be assuming that there is always at least one data_var; is that always true? I.e., it's not possible to have a Dataset with coordinates but no data values? If so, fine; if not, should first check that the list is nonempty.

I checked and it's actually possible to have a Dataset with no data values. The code in that function already assumed (line 660) that Dataset objects have at least one data_var. I'll add a check which will raise an error in the event of a Dataset with no data_var.

> Why does it only use the title from the first data_var? Is it known that will be the only one plotted? Or should it concatenate all of the titles?

That was a simplification, I've found a way to plot the scalar coords of a Dataset directly.

PR #660, opened 2021-09-10
Fix a broken test due to PROJ unexpectedly matching 'foo'

### LINKED ISSUE #657: Tests failing on Python 3.7 due to CRS interpretation
The tests are currently failing on Python 3.7 because of the test `test_plot_with_crs_as_nonexistent_attr_str` that expects an error to be raised when a wrong CRS (`'foo'`) value is passed to an image plot.

`pyproj` doesn't seem to be bothered by interpreting `'foo'` as a correct CRS (or at least it doesn't raise an error). I've tried it with the version installed in the dev environment (2.6.1post1, *defaults* is lagging on pyproj releases) and with the latest version, it never raises an error and actually returns:
```
<Geographic 2D CRS: EPSG:4289>
Name: Amersfoort
Axis Info [ellipsoidal]:
- Lat[north]: Geodetic latitude (degree)
- Lon[east]: Geodetic longitude (degree)
Area of Use:
- name: Netherlands - onshore, including Waddenzee, Dutch Wadden Islands and 12-mile offshore coastal zone.
- bounds: (3.2, 50.75, 7.22, 53.7)
Datum: Amersfoort
- Ellipsoid: Bessel 1841
- Prime Meridian: Greenwich
```

The utility function that checks the CRS validity in hvplot seems to have been borrowed from `salem` where it has been updated. I've tried validating `'foo'` with salem`s updated utility function and again, it was interpreted as correct.

I've got no clue why `'foo'` is a valid CRS for pyproj (now, since it must have failed in the past for the test suite to pass). Fixing the test though is not so hard, for instance`pyproj.Proj('wrong')` raises an error, even if that's just a work around to get the test suite to pass...

maximlt:
Fixes #657 

The code uses PROJ under the hood that matches `'foo'` with a projection named `'Amersfoort'`. Replaced by '`name_of_some_invalid_projection'` which doesn't match and should never match any projection in the future.

### DIFF
--- a/hvplot/tests/testgeo.py
+++ b/hvplot/tests/testgeo.py
@@ -63,8 +63,11 @@ class TestProjections(TestGeo):
         self.assertCRS(plot)
 
     def test_plot_with_crs_as_nonexistent_attr_str(self):
-        with self.assertRaisesRegex(ValueError, "'foo' must be"):
-            self.da.hvplot.image('x', 'y', crs='foo')
+        # Used to test crs='foo' but this is parsed under-the-hood
+        # by PROJ (projinfo) which matches a geographic projection named
+        # 'Amersfoort'
+        with self.assertRaisesRegex(ValueError, "'wrong' must be"):
+            self.da.hvplot.image('x', 'y', crs='wrong')
 
     def test_plot_with_geo_as_true_crs_no_crs_on_data_returns_default(self):
         da = self.da.copy()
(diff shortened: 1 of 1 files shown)


hvplot/tests/testgeo.py
```diff
@@ -63,8 +63,11 @@ def test_plot_with_crs_as_attr_str(self):
         self.assertCRS(plot)
 
     def test_plot_with_crs_as_nonexistent_attr_str(self):
-        with self.assertRaisesRegex(ValueError, "'foo' must be"):
-            self.da.hvplot.image('x', 'y', crs='foo')
+        # Used to test crs='foo' but this is parsed under-the-hood
+        # by PROJ (projinfo) which matches a geographic projection named
+        # 'Amersfoort'
+        with self.assertRaisesRegex(ValueError, "'name_of_some_invalid_projection' must be"):
+            self.da.hvplot.image('x', 'y', crs='wrong')
```
jbednar, line 70:
```suggestion
            self.da.hvplot.image('x', 'y', crs='name_of_some_invalid_projection')
```

PR #679, opened 2021-11-19
Update the default internal value of `clim` to support Matplotlib

maximlt:
PR in preparation of the upcoming support of Matplotlib and Plotly.

`self._dim_ranges['c']` was set to either `clim` if not None or to `(None, None)`. This caused no issue with the Bokeh backend of holoviews, which accepts a `param.Tuple` for `clim` and handles properly when it includes `None`. However both the Matplotlib and Plotly backends accept a `param.NumericTuple` only.

The Bokeh backend used to accept a `param.NumericTuple` only as well, but that was changed in https://github.com/holoviz/holoviews/pull/4383. The right fix might be to add support to whatever the Bokeh backend supports to Matplotlib & Plotly with allowing them to accept a non-numeric tuple for `clim`, but since I don't know what exactly this PR was about I went for the simpler, and hopefully correct, solution.

An important question here would be, does the bokeh backend behave the same way when `clim` is set to `(None, None)`  compared to `(np.nan, np.nan)`?

@jlstevens pinging you since we talked about it briefly already 🙃

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -1949,12 +1949,20 @@ class HoloViewsConverter(object):
         levels = self.kwds.get('levels', 5)
         if isinstance(levels, int):
             opts['color_levels'] = levels
-        if self._dim_ranges['c'] != (None, None):
-            opts['clim'] = self._dim_ranges['c']
         return contours(qmesh, filled=filled, levels=levels).opts(**opts)
 
     def contourf(self, x=None, y=None, z=None, data=None):
-        return self.contour(x, y, z, data, filled=True)
+        contourf = self.contour(x, y, z, data, filled=True)
+        # The holoviews contours operation used in self.contour adapts
+        # the value dim range, so we need to redimension it if the user
+        # has asked for it by setting `clim`. Internally an unset `clim`
+        # is identified by `(None, None)`.
+        if self._dim_ranges['c'] != (None, None):
+            z_name = contourf.vdims[0].name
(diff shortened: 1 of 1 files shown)


maximlt:
Ok so my first attempt to solve this made the test suite fail :/ In the second and successful attempt, at least in light of the test suite, the `clim` option is set in the contours plot only if `c` was already computed/set. I have to try that now with the matplotlib backend.

maximlt:
It's working well with the matplotlib backend!

maximlt:
I think I've finally found an acceptable solution but let me know what you think about it.

It happens that what was done in #587, i.e. setting the `clim` opts to plots created by `.contour` (and `.contourf` since it uses `.contour`), was only relevant for `.contourf`. `.contour` has a first step that consists in calling `.quadmesh` which applies a redim to the object it returns, this effectively takes into account the `clim` parameter a user can pass in the hvplot API. However, it happens that the `contours` holoviews operation **overrides** the dimension range in the case of a filled contour:

https://github.com/holoviz/holoviews/blob/4ea310adcfb8fe061faf5acc732522e366dc01b1/holoviews/operation/element.py#L591-L593

So I've removed the line added by #587, and replaced it by a `.redim.range` only applied to the element returned by `.contourf`. Using `.redim.range` allows to fix the issue solved originally by this PR, i.e. to prepare the code for the Matplotlib backend which doesn't allow the `clim` tuple to contain any `None`.

PR #690, opened 2021-12-16
Allow to manually trigger a docs build

maximlt:


maximlt:
Merging directly since this doesn't affect the tests.

PR #701, opened 2022-01-24
CI: Test Github actions test workflow with py3.8-3.10

MridulS:
[Just wanted to test the pipelines with newer py versions]

PR #702, opened 2022-01-26
MAINT: Drop support for py2.7, min supported version is py3.6 now

MridulS:
I went a bit aggressive with dropping older python version support 😅
This PR makes 3.7 the min supported version.

Is this reasonable? 

The test pipelines do not work with 3.8, 3.9 and 3.10 (https://github.com/holoviz/hvplot/pull/701) but I think the code itself should work.


jlstevens:
> Is this reasonable?

Thanks for dropping Python 2 support (finally!)

As for dropping 3.6 right now, I do think that is too aggressive. I do know various projects are also aggressively dropping support within the 3.x series but I do think there are still plenty of users still using python 3.6...

MridulS:
The syntax should be fine with 3.6 too.

jlstevens:
Thanks!

Once the tests pass, I think it looks good to merge!

maximlt:
Great thanks for updating it @MridulS , merging!

PR #703, opened 2022-01-26
Improve test workflow on GHA

maximlt:
- [x] Run it on Py38 and Py39
- [x] Run it on MacOs
- [x] Switch to pytest
- [x] Switch to codecov
- [x] Check the tests skipped
- [x] Run the CI on Py36


maximlt:
Ok I **think** I found the problem we've observed on Macos in this repo and elsewhere (currently breaking Panel's CI for instance). This happens when Python is updated while `doit develop_install ...` runs. Which is actually kind of crazy but appears to pose no problem on Windows and Linux.

setup.cfg
```diff
@@ -1,2 +1,7 @@
 [metadata]
 license_file = LICENSE
+
+[tool:pyctdev.conda]
+; Install geopandas-base on conda to avoid environment solving
+; to fail on defaults with fiona (this failed conda install netcdf4 fiona)
+namespace_map = geopandas=geopandas-base
```
philippjfr, line 7: ❤️

maximlt:
I'll merge this now since the test suite finally passes:
- on Python 3.6 I haven't been able to get conda to solve so I've had to limit the tests to execute to the unit tests, discarding the examples test (smoke test the notebooks). mamba did manage to solve (locally) so it's likely that when conda gets the libmamba solver we will be able to run the examples test again on 3.6. Funny fact, I once saw an 137 exit code while conda was solving, meaning that the container ran out of memory!
- the main difference with how things used to work is that now the conda channels used in the whole job are configured in the `setup-miniconda` action, as long as with the strict channel parameter. All the later calls are then bare `conda install package` without any `-c channel1` since they're already set.
- Updated some projection-related code by copying it from `salem` (this was already the case) since the ecosystem has seen some change in this field (cartopy, pyproj)

PR #705, opened 2022-02-14
Use nbval and nbqa for managing notebooks

MridulS:
Update: Moved to use the `flake8` linter instead, it respects `# noqa`and it's passing the CI without complaining. For`nbsmoke_skip_run`, I've used a `pytest` setting `ignore-glob`.

 ----------------------------------------------------
Currently the `pylint` module for `nbqa` complains a lot as `pylint` checks for many more things as compared to `pyflakes`.
We can also turn on other linters like `black`, `isort`, `pyupgrade` for the notebooks.

One thing that doesn't directly translate from `nbsmoke` to `nbval` is the the `nbsmoke_skip_run` pytest setting. We can use the explicit `ignore` while executing `pytest --nbval-lax` but it won't accept regex (I think?)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,8 +75,9 @@ extras_require = {
         'parameterized',
         'pytest',
         'pytest-cov',
-        'nbsmoke >=0.2.0',
-        'numpy >=1.7'
+        'numpy >=1.7',
+        'nbqa',
+        'nbval',
     ],
     'examples': _examples,
     'examples_extra': _examples_extra,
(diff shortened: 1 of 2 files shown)


maximlt:
@MridulS could you please push an empty commit to trigger the CI?

MridulS:
Hmm, I am getting a "Startup failure" while running the job on Github Actions https://github.com/holoviz/hvplot/actions/runs/2027074647. Interesting, I haven't seen this before.

maximlt:
I've never seen that too and the error reported at line 67 seems just wrong, there's no `cache` identifier in this workflow. Do you want me to open a new PR with your changes to see how that goes?

MridulS:
yeah sure! please do that.

PR #709, opened 2022-02-21
CI: Cache conda builds

MridulS:
Just testing out if we can make the github action builds a bit quicker.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -16,6 +16,13 @@ jobs:
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
         python-version: [3.6, 3.7, 3.8, 3.9]
+        include:
+        - os: ubuntu-latest
+          path: /usr/share/miniconda3/envs/test
+        - os: macos-latest
+          path: /Users/runner/miniconda3/envs/test
+        - os: windows-latest
+          path: C:\Miniconda3\envs\test
     timeout-minutes: 60
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


MridulS:
This should work as expected now, the cache is invalidated every day once (date is used in the key for the cache). It will also rerun if there is any change to setup.py.

There is one thing that I don't like is the editable reinstall of hvplot in the case there isn't a cache hit.

Some basic improvements:
Linux-py3.6 with cache goes from 10mins 42 secs to 1min 26 secs
Linux-py3.8 with cache goes from 26mins to 15 mins.

If this looks okay I guess we can merge it, I can babysit this next couple of days to check if everything is in order.

.github/workflows/test.yaml
```diff
@@ -72,6 +91,7 @@ jobs:
         run: |
           eval "$(conda shell.bash hook)"
           conda activate test-environment
+          python -m pip install -e .
```
maximlt, line 94: It's unfortunate to have to add this command but I guess it's unavoidable! I think the exact command run by pyctdev is `pip install --no-deps --no-build-isolation -e .` so we may want to add these two flags too, mostly for the sake of being consistent.

(I'm actually not sure that installing in editable mode is required, this is a practical default of pyctdev when running it locally to install a dev env because this is what you want but in a CI context it certainly doesn't make that much sense, if any).

maximlt (commented):
Great @MridulS ! These time improvements are nice, we'll be better oss citizen when it's merged :) I've got a couple of comments/questions:
* is there a way to invalidate the cache if need be? Maybe by writing a commit message including a special flag (e.g. `my commit [invalidate_cache]`) or via the GitHub UI?
* I tend to think that if the workflow is triggered after a tag was set then the cache shouldn't be used, is there a way to get that behavior?

.github/workflows/test.yaml
MridulS: It is required to do an editable install in this case as somewhere in the tests there is an assumption that hvplot is pip installed.

maximlt: I meant I think it'd be possible to pip install it without the editable flag, but it's fine as it is since it's anyway what `pyctdev` does.

MridulS:
> is there a way to invalidate the cache if need be? Maybe by writing a commit message including a special flag (e.g. my commit [invalidate_cache]) or via the GitHub UI?

Unfortunately no, there is no native way of doing this https://github.com/actions/cache/issues/2
We need to add some whitespace character in setup.py to force rebuild this.

> I tend to think that if the workflow is triggered after a tag was set then the cache shouldn't be used, is there a way to get that behavior?

Hmm, I do think we should do that but I'm not sure if that's possible currently (unless doing the whitespace hack).
Hopefully the invalidate cache is fixed by github (it seems to be on there Q3 22 roadmap)

maximlt:
How about defining a CACHE_INDEX environment variable, defaulting to 0, that would be used to create the cache hash? I think with that:
* we could bump it in Github's settings (per repo) to declare the cache a invalid at a point in time
* if the action is triggered with a special flag in the commit message then there could be a step that changes CACHE_INDEX to bump it or to a uuid. Same for if the action is triggered by a new tag.
What do you think?

MridulS:
I like that!

PS: The failing tests doesn't seem related.

maximlt:
Thanks @MridulS , I'll merge and we'll get the chance to try this very soon in hvPlot for its next release.

PR #711, opened 2022-02-25
Update datasets.yaml

Hoxbro:
What is defined in regex is removed from the filename not kept.
Before: `/home/shh/.intake/cache/7f9708758ba435683113a36d5d76f3db/assets.holoviews.org/data/.parq`
After: `/home/shh/.intake/cache/032a8a7804d6ce7538724a8e4880f80e/airline_flights.parq`

### DIFF
--- a/examples/datasets.yaml
+++ b/examples/datasets.yaml
@@ -4,7 +4,7 @@ sources:
     driver: parquet
     cache:
       - argkey: urlpath
-        regex: 'airline_flights'
+        regex: 'assets.holoviews.org/data/'
         type: file
     args:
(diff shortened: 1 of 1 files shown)


jbednar:
Thanks! What's the symptom that this fix addresses? Caching failing to recognize the file having already been downloaded?

Hoxbro:
This solves nothing related to the caching this should be solved upstream. 

The current implementation works and will download a file, my suggested changes will give it the correct filename in the cache folder e.g. `airline_flights.parq` instead of the current implementation which creates two folders and have an filename with only the extension in it  `assets.holoviews.org/data/.parq`.

PR #714, opened 2022-03-20
Improve array-like handling for x and y inputs

### LINKED ISSUE #713: Improve handling of arraylike as input for hvplot
``` 
pandas    1.3.5
hvplot    0.8.0a11.post9+gfbcc04e
```

When passing an array type like `df.columns` into `hvplot` it raises `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`. 

I think there is two solution for this problem; 
* one is either to support array-like formats like `numpy` or `pandas` as input for `hvplot`, 
* the other options is to raise a better error message when passing in a array-like type. 

``` python
import pandas as pd
import hvplot.pandas

df = pd._testing.makeDataFrame()
df.hvplot(y=df.columns)  # Raises ValueError
df.hvplot(y=list(df.columns))  # Works fine
```

Hoxbro:
Fixes #713

Properly one of the simplest ways to support array-like in hvplot.

### DIFF
--- a/hvplot/plotting/core.py
+++ b/hvplot/plotting/core.py
@@ -8,7 +8,7 @@ except:
     panel_available = False
 
 from ..converter import HoloViewsConverter
-from ..util import process_dynamic_args
+from ..util import is_list_like, process_dynamic_args
 
 
 class hvPlotBase:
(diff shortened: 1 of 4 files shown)


maximlt:
Thanks @Hoxbro for the initial push, I've just added a utility function that checks if an object is list-like (basically copied from pandas) instead of relying on checking for the `all` method.

PR #722, opened 2022-03-27
Added white background to diagram

Hoxbro:
I have installed [Dark Reader](https://addons.mozilla.org/da/firefox/addon/darkreader/) and noticed that the current diagram is missing a white background:

image

(On a small side note, it is awesome how GitHub can now compare pictures)


jbednar (approved):
Thanks! Yes, the new GH image compare tools are super useful!

PR #724, opened 2022-03-30
Fix the CI on Python 3.6

maximlt:
By installing importlib_resources, dependency required by the latest tqdm but not declared in its recipe (https://github.com/conda-forge/tqdm-feedstock/pull/114).


maximlt:
The run on MacOS / Python 3.7 is failing due to some issue with fiona, I've opened an issue on the conda-forge repo of fiona (https://github.com/conda-forge/fiona-feedstock/issues/194).

Issue #706, opened 2022-02-17
hvplot is not using lazy loading for zarr-backed xarray dataset


rabernat:
#### ALL software version info

```
xr.__version__
'0.21.1'
hvplot.__version__
'0.7.3'
holoviews.__version__
'1.14.7'
```

#### Description of expected behavior and the observed behavior

When I visualize zarr-backed xarray datasets with hvplot, I expect the loading to be lazy. I don't want to load the entire array into memory. I expect this to happen _even without dask being involved_.

Instead, I am noticing that hvplot is loading data eagerly.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import xarray as xr
import hvplot.xarray

url = "https://ncsa.osn.xsede.org/Pangeo/pangeo-forge/swot_adac/FESOM/surf/fma.zarr"
ds = xr.open_dataset(url, engine='zarr') # note that ds is not chunked but still uses lazy loading

# this will trigger loading the whole array into memory
ds.sst.hvplot(x='lon', y='lat', dynamic=True, rasterize=True)
```

Note that lazyness can be achieved by chunking the xarray dataset. But I don't want to have to do that. Dask introduces additional latency that I would like to avoid here.

jbednar:
Do you have a non-hvplot example of accessing this data and extracting a bit of it lazily, and can you explain precisely how it's lazy (i.e. what parts won't be loaded into memory)? Here it looks like you're forcing hvplot to read the entirety of `sst`, at least, since it's plotting all of it.

rabernat:
There is a time dimension. 

```python
ds.isel(time=0)
```

will lazily grab a single time step, which is all that should be required for a single image plot.

rabernat:
If you don't explicitly chunk the data, the underlying data are Xarray's custom LazilyIndexArray types. One thing that might be triggering this would be explicit coercion to a numpy array.

I could swear that this worked perfectly in a earlier version of hvplot. So I'm confused why I am getting issues now.

jbednar:
[...]

PR #741, opened 2022-04-19
Ensure zarr based DataArrays are not loaded into memory

philippjfr:
Fixes https://github.com/holoviz/hvplot/issues/706

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -599,9 +599,9 @@ class HoloViewsConverter:
 
         if is_xarray(self.data):
             # chunks mean it's lazily loaded; nanquantile will eagerly load
-            if self.data.chunks:
-                return False
             data = self.data[self.z]
+            if not getattr(data, '_in_memory', True) or data.chunks:
+                return False
             if is_xarray_dataarray(data):
                 if data.size > check_symmetric_max:
                     return False
--- a/hvplot/tests/testgridplots.py
+++ b/hvplot/tests/testgridplots.py
@@ -213,6 +213,7 @@ class TestGridPlots(ComparisonTestCase):
             # If a DataArray is not in memory, computing whether it's symmetric should
             # not be done and return False.
             assert not plot_opts.kwargs['symmetric']
+            ds.close()
     
     def test_symmetric_dataset_in_memory(self):
         da = xr.DataArray(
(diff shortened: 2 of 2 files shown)


jbednar:
Thanks. Is there a way to demonstrate whether this fixes the issue?

maximlt:
I've added a test to check that, when the data is "symmetric" but is "not in memory", `_process_symmetric` returns False early to avoid computing `nanquantile` on the whole data which would eagerly loads it if it was not yet already done.

This PR is actually a bigger change than I thought. For instance, opening a local NetCDF file as such `import xarray as xr; ds = xr.open_dataset('data.nc')` will not fully load it "in memory", i.e. `ds.variable._in_memory` will be `False` (this is what I've done to build the test by the way). If the `ds.variable` data happened to be symmetric then it will no longer be displayed with a diverging color map after this PR, at least not without passing directly `symmetric=True`.

Thoughts on this @jbednar, @philippjfr ?

jbednar:
@rabernat, do you have any suggestions?

hvplot/converter.py
```diff
@@ -599,9 +599,9 @@ def _process_symmetric(self, symmetric, clim, check_symmetric_max):
 
         if is_xarray(self.data):
             # chunks mean it's lazily loaded; nanquantile will eagerly load
-            if self.data.chunks:
-                return False
             data = self.data[self.z]
+            if not getattr(data, '_in_memory', True) or data.chunks:
+                return False
```
rabernat, line 604: This seems reasonable to me insofar as it treats non-in-memory data the same as chunked data

rabernat (approved):
Everything here looks good to me!

rabernat:
> If the `ds.variable` data happened to be symmetric then it will no longer be displayed with a diverging color map after this PR, at least not without passing directly `symmetric=True`.

Is it possible for this behavior to be dependent on whether this is a "dynamic" plot or not? Or some way for the user to explicitly indicate that they do not want or expect the data to be loaded eagerly.

maximlt:
This is indeed a change in behavior, but we believe that most hvPlot users are more likely to be in a space where the new behavior is to their advantage, such as climate/earth scientists that use hvPlot to display very large datasets and for who loading the full data is just plain wrong.

PR #758, opened 2022-05-18
Make `rescale_discrete_levels` a plot option

maximlt:
It was a datashader operation option, it's now a plot option that has no default value, like `cnorm`.

I assumed https://github.com/holoviz/holoviews/pull/5268 would be merged, setting `rescale_discrete_levels` to `True` by default in the `shader` operation. If not, and if the intent is to have it set by default to `True` in hvPlot but not in HoloViews (I'd prefer them having the same default), this PR will have to be adapted.

```python
import numpy as np, pandas as pd
import hvplot.pandas

def random_walk(n, f=5000):
    """Random walk in a 2D space, smoothed with a filter of length f"""
    xs = np.convolve(np.random.normal(0, 0.1, size=n), np.ones(f)/f).cumsum()
    ys = np.convolve(np.random.normal(0, 0.1, size=n), np.ones(f)/f).cumsum()
    xs += 0.1*np.sin(0.1*np.array(range(n-1+f))) # add wobble on x axis
    xs += np.random.normal(0, 0.005, size=n-1+f) # add measurement noise
    ys += np.random.normal(0, 0.005, size=n-1+f)
    return np.column_stack([xs, ys])

paths = random_walk(2000,30)
df = pd.DataFrame(paths, columns=['x', 'y'])

df.hvplot.paths(x='x', y='y', aspect=1, datashade=True, cnorm='eq_hist') + \
df.hvplot.paths(x='x', y='y', aspect=1, rasterize=True, cnorm='eq_hist')
```

image


jbednar (approved):
Thanks. I think this is ready to merge, along with the corresponding HoloViews PR.

hvplot/converter.py
```diff
@@ -572,6 +578,8 @@ def __init__(
                 pass
         if cnorm is not None:
             plot_opts['cnorm'] = cnorm
+        if rescale_discrete_levels is not None:
+            plot_opts['rescale_discrete_levels'] = rescale_discrete_levels
```
philippjfr, line 582: Need to guard this with a version check no?

maximlt: Yes indeed, I've added a version check for HoloViews>=1.15.0 in https://github.com/holoviz/hvplot/pull/758/commits/c5f325ac96cf45308ec411e50d95a885ac0c7e2f

PR #761, opened 2022-05-30
Fixes subplots=True for geo=True and with by-statement

Hoxbro:
The second part (2/2) of the fix for https://discourse.holoviz.org/t/small-multiples-aka-facetted-plots/3760/2

This will make it possible to run (after https://github.com/holoviz/holoviews/pull/5325 is merged):
``` python
import geopandas as gpd
import hvplot.pandas

countries = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))
countries.hvplot(geo=True, by="continent", subplots=True)
```

Before fix (with https://github.com/holoviz/holoviews/pull/5325):
image

After fix:
image


jbednar (approved):
Good catch!

maximlt:
Added an `xfail` test in https://github.com/holoviz/hvplot/pull/761/commits/f30077c58841a6d6bb05404bf98f6f231a7074c0 that was passing with https://github.com/holoviz/holoviews/pull/5325.

We've discussed about https://github.com/holoviz/holoviews/pull/5325 recently and the fix on the HoloViews side is likely going to be different. But that doesn't really matter for hvPlot, I think. @Hoxbro do you see any reason not to merge this?

maximlt:
Great thanks, merging then.

PR #768, opened 2022-07-05
Reduce bound input function in Interactive calls to one

Hoxbro:
Fixes #763

Most extra function calls were done inside `__init__`. I have added the `Watcher` to the `watchers` list if it is not already there, and if it is on the list, I `unwatch` it. I have a suspicion this is not the best practice and could be done more elegantly.  

The change in `__new__` reduces the bound function call from two to one (after the change in `__init__`).

PR #772, opened 2022-07-08
Set interactive function to lazy to avoid calling it twice

Hoxbro:
Splits up the fixes in #768 as this change is unrelated to the warning.  

I have commented out part of the test related to the changes in #768, that PR should uncomment those lines.


Hoxbro:
The failing tests were related to a data file no longer being at the https://github.com/mapbox/rasterio/raw/master/tests/data/RGB.byte.tif, but here https://github.com/rasterio/rasterio/raw/main/tests/data/RGB.byte.tif. A change from master --> main. 

I did add the data file to the commit instead of using the URL as before to avoid having a problem like this in the future. But I'm wondering if this could affect the package size and if I should go back to using the URL.

jbednar:
1.67MB does seem like a lot just to get an example image, but I'm not sure what fraction of the overall package that represents.

Hoxbro:
What if I exclude the file in `MANIFEST.in`? I can then, in a follow-up PR, change the path to a hvplot URL. 

That way, we have complete control over the file and the data in it, and people will still be able to run the test locally.

MarcSkovMadsen:
With pyscript and python

philippjfr:
Can we use a more minimal tif file?

Hoxbro:
> Can we use a more minimal tif file?

Properly, but I don't know how comprehensive it will be to create and update the code. I will change it back to a URL and make an issue about updating the file testing.

maximlt:
This looks good to me, thanks!

Issue #773, opened 2022-07-10
Extension not loaded when importing `hvplot.pandas` after running `panel.extension()`


maximlt:
Recording a bug observed on the HoloViz tutorial after upgrading its pins to the latest hvplot/panel/holoviews/etc. Not sure it's a hvplot bug so this issue might have to be transferred to the right place after more investigation.

<img width="283" alt="image" src="https://user-images.githubusercontent.com/35924738/178141272-df6c531f-5e26-4376-a56e-ee5849d3c18c.png">

philippjfr:
Bug probably caused by the change to extension loading but the fix is indeed in hvPlot.

philippjfr:
Hmm, actually pretty confused by this. Don't see how this ever worked given the current logic.

Cyb3r-Monk:
@maximlt hvplot is still on 0.8.0 in `pkgs/main/osx-arm64`channel. Would it be possible to update it? I'd been having this issue and randomly found this bug and saw that osx-arm64 channel still had the buggy version.

Hoxbro:
It is not possible to update that channel by anyone on the holoviz team, but you can install it through the pyviz channel: `conda install -c pyviz hvplot=0.8.1`.

PR #774, opened 2022-07-10
Ensure that repr is initialized on hvplot backend import

philippjfr:
Previously if the panel extension had been loaded this `post_patch` hook would not call `hv.extension` which meant that the reprs required for rendering HoloViews components were not registered. I'm not entirely sure how this ever worked but now that we avoid repeat cell execution it is perfectly fine to call `hv.extension` multiple times.

Fixes https://github.com/holoviz/hvplot/issues/773


maximlt:
Just to be sure, what we don't and maybe can't cover is the case where e.g. a user loads `hv.extension('bokeh')`, then somehow manages to get rid of the loaded Javascript/CSS by deleting the cell and reloading the browser tab, then imports `hvplot.pandas`. In that case the remaining optimization `if` will still identify the extension as loaded so the loading of the Javascript/CSS won't happen again and will still be missing, any later call to `df.hvplot()` will not display anything.

This differs from the behavior of `hv.extension()`. I.e. running`hv.extension('bokeh')` in different cells will load the Javascript/CSS everytime, while running `import hvplot.pandas` in different cells will only load in the first cell.

I just want to make sure we're clear about the behavior we want to get when executing theses statements in different cells. In a single cell context it's clear from the update made to pyviz_comms that the intention was to only make one Javascript/CSS load whatever the number of statements that would have this side effect.

philippjfr:
That's right, we explicitly tell users that an extension load is needed so users that delete the cell with the loaded extension can't expect things to work.

Issue #771, opened 2022-07-07
.interactive dataframe slice selection by row index or position not possible


JanHomann:
#### Is your feature request related to a problem? Please describe.
I was in the assumption that the .interactive interface for dataframes should mimic the interface of pandas. Yet not everything seems to work.

Given the code
```python
import hvplot.pandas
from bokeh.sampledata import airports
df = airports.data
dfi = df.interactive()

p = pnw.IntSlider(start=10, end=40)
dfi[5:p]
```
 I would like to see a slider and a dynamic dataframe representation showing index 5...p-1 (that dataframe has an integer index). Instead I get a static output without slider. I can make one in another cell, but when I move it, the dataframe output does not update. I need to execute the cell manually again to get the update. 

This here doesn't work at all and gives an InvalidIndexError
```python
dfi.loc[5:p,:]
# InvalidIndexError: IntSlider(end=40, start=10, value=10)
```

Those here give no slider and don't update when I create a slider in another cell:
```python
dfi.loc[:p]

dfi.iloc[5:p]
dfi.iloc[slice(5,p)]
dfi.values[5:p]
```

This here gives a type error
```python
dfi.loc[dfi.index[5:p]]
# TypeError: slice indices must be integers or None or have an __index__ method
```

A current workaround for `dfi.loc[5:p,:]` is:
```python
dfi[(dfi.index>=5) & (dfi.index<p)]
```

For `dfi.loc[5:p,:]` I have no reasonable workaround (except for resetting the index, selecting via above method and setting it again).

Given that neither `dfi[:p]` nor `dfi.iloc[:p,:]` nor `dfi.values[5:p]` nor `dfi.loc[dfi.index[5:p]]` work, I don't know how I would even select rows by position. 
Literally the first example in the interactive tutorial (https://hvplot.holoviz.org/user_guide/Interactive.html) has indexing by position in an xarray with `.isel`, which I think i have tried and it works. So it is surprising to me that there is no mechanism to select a positional range in dataframes. Or maybe there is some Pandas trick that I don't know.

JanHomann:
Yes, I suppose the mechanism by which this works in Pandas is that a slice object is passed to the dataframe method. Makes sense. Is there currently a workaround for positional indexing?

JanHomann:
[...]

PR #776, opened 2022-07-11
Support slice in interactive

Hoxbro:
Part 2 of the solution for https://github.com/holoviz/hvplot/issues/771.

It needs https://github.com/holoviz/holoviews/pull/5352 to work.

The panel check is removed because `panel` is already imported at the top. Therefore, I think the check is unnecessary, but correct me if I'm wrong.


maximlt:
`test_interactive_slice` is failing on Python 3.6. In these jobs the HoloViews version is 1.14.8 which doesn't yet have the fix to support slices that was added in https://github.com/holoviz/holoviews/pull/5352 and made available in HoloViews 1.15.1. @Hoxbro could you a guard checking HoloViews version after `iff isinstance(op_arg, slice):` and raising an error if it's too old, and skip that test for older HoloViews version?

Issue #777, opened 2022-07-11
No colorbar displayed when `cmap` is set


maximlt:
See while running the tutorial, when `cmap` is defined the colorbar isn't rendered. 

```python
df[df.mag>5].hvplot.scatter(x='longitude', y='latitude', rasterize=True, cnorm='eq_hist', cmap='Reds')
```

<img width="720" alt="image" src="https://user-images.githubusercontent.com/35924738/178339795-ec80c8e7-006e-4a45-b7e4-441742ad1bdd.png">

PR #782, opened 2022-07-21
Display a color bar when rasterize is enabled

maximlt:
Fixes https://github.com/holoviz/hvplot/issues/777

Before https://github.com/holoviz/hvplot/pull/206 the `cmap` style was handled as follows, i.e. when datashader was involved it was set to the default linear colormap:

```python
        if cmap is not None:
            style_opts['cmap'] = cmap
        elif self.rasterize or self.datashade:
            style_opts['cmap'] = self._default_cmaps['linear']
```

https://github.com/holoviz/hvplot/pull/206 made hvplot behave more closely to `xarray.plot()` by using a symmetric colormap if the data is found to be divergent. https://github.com/holoviz/hvplot/pull/206 replaced the code that set the colormap to linear, by setting the color bar, to `True` by default:

```python
        if cmap is not None:
            style_opts['cmap'] = cmap
        elif self.rasterize or self.datashade:
            plot_opts['colorbar'] = plot_opts.get('colorbar', True)
```

This PR:
* moves up the logic that handles the datashaded path, as it seems it could be set up-front
* sets the color bar to `True` by default when `rasterize=True`, whatever the value of `cmap`. This fixes the related issue. I'm not entirely sure this is the correct behavior for all plot kinds.
* doesn't set the color bar when `datashade=True` as a color bar isn't relevant with this option.

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -526,6 +526,8 @@ class HoloViewsConverter:
             plot_opts['colorbar'] = colorbar
         elif self.kind in self._colorbar_types:
             plot_opts['colorbar'] = True
+        elif self.rasterize or self.datashade:
+            plot_opts['colorbar'] = plot_opts.get('colorbar', True)
         if 'logz' in kwds and 'logz' in self._kind_options.get(self.kind, {}):
             plot_opts['logz'] = kwds.pop('logz')
         if invert:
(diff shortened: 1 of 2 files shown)


jbednar:
Seems ok to me, given that hvplot seems to default to using a colorbar whenever it's relevant, and it's certainly relevant when  datashading! Presumably specifying `cnorm='eq_hist'` overrides the `linear` setting?

maximlt:
> Presumably specifying `cnorm='eq_hist'` overrides the `linear` setting?

Yes it does:

image

philippjfr:
Perfect, fix looks good!

Issue #789, opened 2022-07-24
Make hvPlot tooltips useful in modern editors and IDEs


MarcSkovMadsen:
## Request

Make sure tooltips shown in modern Editors and IDEs like VS Code and PyCharm are available and useful.

## Motivation

I'm trying to help a user of hvPlot here https://discourse.holoviz.org/t/linked-plot-is-exceedingly-slow/3966. I am a user of VS Code. Its just a very slow process helping the user because I need to search and read things on the hvPlot website instead of being able to stay in my editor.

The issue is that the tooltips shown by VS Code are either empty or not useful. My guess is that its because no one has ever considered to make sure the docstrings of hvPlot works in modern editors and IDEs.

The consequence is that hvPlot is much, much harder to learn and work with than it has to be for editor and IDE users.

## scatter

`.hvplot.scatter` has no tooltip.

![image](https://user-images.githubusercontent.com/42288570/180633157-7e39aacf-78e5-4829-bad6-f6e0ced3b672.png)

maximlt:
I agree that better docstrings and documentation is required.

> I am a user of VS Code. Its just a very slow process helping the user because I need to search and read things on the hvPlot website instead of being able to stay in my editor.

Of course the more you can stay working in your editor the better. However don't forget that because of the flat API of the `.plot()` API of Pandas, which applies to the `.hvplot()` API of hvPlot, the provided methods accept a large number of arguments, and this by design. See for instance the tooltip of Pandas `df.plot.scatter()`, which doesn't list all the kwargs it accepts.

![image](https://user-images.githubusercontent.com/35924738/180636880-ed2b58ce-c0e9-4a6e-bdb4-17c6a84601b5.png)

In most cases then you will have to go to Pandas website and look for what kwargs you can pass. The same applies to hvPlot, even more so as it provides more powerful arguments for which a simple explanation in a tooltip may not be enough to be understood properly (e.g. `rasterize`).

Also note that Pandas docstring doesn't link to the reference page and the *Examples* section doesn't display plots. Even projects like Pandas with about 100M downloads/month have difficulties to make this right.

![image](https://user-images.githubusercontent.com/35924738/180637132-ad1d684c-326a-43f2-8d39-315562c0285a.png)

[...]

PR #793, opened 2022-07-24
Improved module level docstrings

MarcSkovMadsen:
Addresses a small part of https://github.com/holoviz/hvplot/issues/789 by adding module level docstrings.

The aim here is to provide

- an elevator pitch for hvplot
- a very basic intro and reminder of the api
- info enough to easily navigate `hvplot` and its resources from the editor.

I believe all the `hvplot.pandas`, `hvplot.xarray`, `hvplot.ibis` needs docstrings that are similar but specialized to the specific data source. These are the entry points for users. Not the top-level `hvplot` module.

There is one issue though. VS Code does not show docstring tooltips for nested imports like `import hvplot.pandas`. It only shows for `from hvplot import pandas`. See https://github.com/microsoft/pylance-release/issues/3093.

from-hvplot-import-pandas (1)


hvplot/__init__.py
```diff
@@ -1,3 +1,110 @@
...
+
+and list the available options using `hv.help´.
+
+>>> import holoviews as hv
+>>> hv.extension("bokeh")
+>>> hv.help(curves)
+
+# Todo: fix https://github.com/holoviz/holoviews/issues/5364
+
+How to create data apps
```
maximlt, line 95: I find this part a little too specific for a high-level pitch, hvPlot users can live happily without knowing anything about Panel (the opposite holds too). Also it's suddenly serving a script while I tend to think the main usage of hvPlot is in a notebook for data exploration.

Maybe adding a similar section in a second step would be useful once hvPlot's documentation is improved, with a section that clearly explains how to create an app with Panel. The [Viewing](https://hvplot.holoviz.org/user_guide/Viewing.html) section does some of that but mixes different concepts.

maximlt:
Thanks Marc!

> There is one issue though. VS Code does not show docstring tooltips for nested imports like `import hvplot.pandas`. It only shows for `from hvplot import pandas`. See [microsoft/pylance-release#3093](https://github.com/microsoft/pylance-release/issues/3093).

Oh I would not advise anyone to make such an import! `from hvplot import pandas` is confusing as it gives the impression you're importing pandas. Until the Pylance issue is resolved I don't see much point in adding high-level pitch docstring to the modules like `pandas.py`.

(I think the tests are failing because of the latest xarray)

hvplot/__init__.py
```diff
@@ -1,3 +1,110 @@
...
+>>> idx = pd.date_range('1/1/2000', periods=1000)
+>>> df  = pd.DataFrame(np.random.randn(1000, 4), index=idx, columns=list('ABCD')).cumsum()
+
+Import the hvplot extension for your data source
+
+>>> import hvplot.pandas
+
+Use the `.hvplot` api as you would use the Pandas `.plot` api.
+
+>>> curves=df.hvplot()
```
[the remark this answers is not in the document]
MarcSkovMadsen: Thanks for this one. I'm just so used to using black in my work that I never think about formatting. Would love to use black with HoloViz one day.

hvplot/__init__.py
```diff
@@ -1,3 +1,110 @@
...
+A, B, C and D time series.
+
+For more check out the user guide https://hvplot.holoviz.org/user_guide/index.html
+
+How to select the kind of plot
+------------------------------
+
+You can add the `kind` argument
+
+>>> df.hvplot(kind="scatter")
```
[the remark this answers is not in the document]
MarcSkovMadsen: What I am trying to achieve is that you can look at the docstring and learn the basics. The first thing I believe you want after seeing the basic example is to understand how to customize. And `kind` or "`.kind`" is the way to go.

I also believe its useful for users who have learned the basics but its been a while since they used hvPlot to be able to get a quick refresher.

Maybe this is too much based on my experience and too much to add?

hvplot/__init__.py
MarcSkovMadsen: I feel strongly for this section. But its probably my personal preference. The reason why I put it in is 

1. I believe having a server with autoreload is a very efficient way of working that you as a user should now about. Its really a great starting point if you are trying to create something reusable.
2. Together with the HoloViews section is shows how you can drop into other parts of the ecosystem, such that there are no dead ends.

Maybe this is too much based on my experience and too much to add?

MarcSkovMadsen:
> Thanks Marc!
> 
> > There is one issue though. VS Code does not show docstring tooltips for nested imports like `import hvplot.pandas`. It only shows for `from hvplot import pandas`. See [microsoft/pylance-release#3093](https://github.com/microsoft/pylance-release/issues/3093).
> 
> Oh I would not advise anyone to make such an import! `from hvplot import pandas` is confusing as it gives the impression you're importing pandas. Until the Pylance issue is resolved I don't see much point in adding high-level pitch docstring to the modules like `pandas.py`.
> 
> (I think the tests are failing because of the latest xarray)

I still see much value :-)

- Modules should have docstrings. Its best practice and helpful.
- It still creates value in Jupyter and VS Code as shown below.
- You have someone willing to contribute the docstrings now, so why not just get them in?

## Before

image
image


## After

image
image
If you *go to definition* you will also find the documentation
image

MarcSkovMadsen:
I believe the *intro section*, *how to .. in 3 simple steps* and *help* is the most important sections. I can give up on the rest of the sections. "Short" is also nice.

Ending the docstring with a link to https://hvplot.holoviz.org/user_guide/index.html is also nice if then that makes it easy for the user to get to a place that explains the rest.

Let me know what to do. Thanks.

hvplot/__init__.py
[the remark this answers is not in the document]
MarcSkovMadsen: I've removed and simplified

hvplot/__init__.py
```diff
@@ -1,3 +1,61 @@
+"""
+hvPlot makes data analysis and visualization simple
+===================================================
+
+hvPlot provides an easy to use, high-level, extended 🐼 Pandas .plot like API
```
jbednar, line 5:
```suggestion
hvPlot provides an easy to use high-level API for visualization, based on the 🐼 Pandas .plot() API
```

hvplot/__init__.py
```diff
@@ -1,3 +1,61 @@
...
+that works across a wide range of data sources and plotting backends.
+
+hvPlot
+
+- supports a wide range of data sources including Pandas, Dask, XArray
+Rapids cuDF, Streamz, Intake, Geopandas, NetworkX and Ibis.
+- supports the plotting backends Bokeh (default), Matplotlib and Plotly.
+- is built on top of HoloViews and allows you to drop into the rest of the
+HoloViz ecosystem when more power or flexibility is needed.
```
jbednar, line 14:
```suggestion
- exposes the powerful tools from the HoloViz ecosystem in a familiar and convenient API, while letting you drop down into the underlying HoloViz tools when more power or flexibility is needed.
```

jbednar (commented):
Looks great; thanks so much!

maximlt:
Thanks @MarcSkovMadsen for the updates and taking into account my suggestions!

I've just realized that starting from Python 3.7 modules can implement `__getattr__` and `__dir__` (https://peps.python.org/pep-0562/). This means we could easily get rid of all of the shim modules that were required before with a cleaner approach, and this when hvPlot will require 3.7 as a minimum version (now it's 3.6 but I guess not for too long). This means that the shim module doctsrings, like the one you added in `pandas.py`, will go away with this new implementation. I'm sorry for the extra work this caused you but I'd suggest you already remove them in this PR? I felt they were anyway very close to the __init__.py module docstring which I already find very useful!

MarcSkovMadsen:
I've removed the *shim* module docstrings @maximlt.

Why do we actually not do `hvplot.extension("pandas")` or `hvplot.extension("xarray")` similarly to `pn.extension` and `hv.extension('matplotlib')`?

That would be consistent with the rest of the ecosystem and work better with static type checking tools.

maximlt:
> I've removed the shim module docstrings

Thanks!

> Why do we actually not do hvplot.extension("pandas") or hvplot.extension("xarray") similarly to pn.extension and hv.extension('matplotlib')?

You can run `hvplot.extension('matplotlib')` to set the Matplotlib backend, which is consistent with HoloViews. I don't know whether allowing `hvplot.extension('pandas')` would make it more consistent with the ecosystem. I find `hv.extension()` and `pn.extension()` quite confusing, too implicit and possibly overloaded. We see users calling them without really knowing what they do, that might be a documentation issue, I don't know.

> work better with static type checking tools.

How?

maximlt:
> I've removed the shim module docstrings @maximlt.

Arf actually I see they're still around. Did you maybe forget to commit?

MarcSkovMadsen:
I've pushed the changes now @maximlt

maximlt:
There were still two weird files on your branch. I took the freedom to delete them, they seemed totally unrelated to that PR. Merging, thanks Marc!

PR #794, opened 2022-07-24
Update developer install instructions

### LINKED ISSUE #790: update developer instructions to use supported Python version
## Request

Please mention recent and modern Python versions in the developer guide

## Motivation

The developer instructions mentions Python 3.6. It reached end of life last year. It could make someone get the wrong impression that hvPlot is outdated and not actively supported to see that.

https://hvplot.holoviz.org/developer_guide/index.html#conda-environments

image

maximlt:
Fixes #790 

Trying a slightly new approach that consists in creating a totally empty conda environment, then setting the channels of this environment only and not globally, and then simply calling `doit/pyctdev` and pure `conda` commands without providing any channels as `-c channel`, just relying on the ones defined at the environment level.


maximlt:
Merging as this is a simply doc update and that the workflow changes have had no bad effects.

Issue #764, opened 2022-06-14
Improve the API of the hvPlotExplorer


maximlt:
There are mostly suggestions that occurred to me while writing the release blog post:

* by providing an easy entry point such as `hvplot.explore(df)` instead of having to instantiate the class `hvPlotExplorer`. I wasn't convinced by the name `hvPlotExplorer` either but it's fine if there's an easy and short entry point instead.
* by providing a way to get a handle on the plot. Should maybe get inspiration from `Interactive` that defines methods to get a handle of the various outputs it offers, such as `.widgets()`, `.panel()`, `.holoviews()`, etc.. Here I'd say I expect a `.plot()` method at least.

maximlt:
Just noting that GeoPandas has a `.explore` method that returns an interactive Folium map: https://geopandas.org/en/stable/docs/reference/api/geopandas.GeoDataFrame.explore.html

jbednar:
I think it's fine for different libraries to have different `explore` functionality.

MarcSkovMadsen:
For me `hvPlotExplorer` has the potential to provide quick BI like visualization of dataframes similar to what is possible with drag and drop in Power BI and Tableau. And furthermore the potential to autogenerate python code.

An updated name could be inspired by some kind of combination of `hvPlot`, `Data`, `DataFrame`, `Explorer`, `Chart`, `Editor`, `Maker`.

Its similar to

- https://www.highcharts.com/blog/products/highcharts-editor/
- https://www.trymito.io/
- https://chart-studio.plotly.com/create/#/

maximlt:
Yes it's interesting because I think when I read `hvPlotExplorer` I understand that it is to explore data, while in practice I would more likely use that component to edit a plot. In that case I already have an idea of the kind of plot I want to get, it's just that the `.hvplot` API has so many parameters that building a plot from this component makes my life easier.

I remember seeing packages that offer a GUI to edit Matplotlib plots. Found one (not sure it's the one I was looking for) that does that: [plt-editor-tool](https://github.com/RichardCouperthwaite/plt-editor-tool). It has an accompanying user guide in PDF from which I've made this screenshot:
<img width="901" alt="image" src="https://user-images.githubusercontent.com/35924738/175488426-1f97f7fd-10c9-492c-8467-6acb6f69b45e.png">

[...]

PR #799, opened 2022-07-26
Make the explorer easier to use

### LINKED ISSUE #787: Add options for saving hvPlotExplorer settings and only show the plot
Maybe I have overlooked something, but it could be nice to be able to save the settings from `hvPlotExplorer` so it can easily be recreated in another session.  

I'm thinking about something like this:

``` python
## First input cell in a notebook
import hvplot.ui
import hvplot.pandas
import panel as pn
pn.extension()

df = penguins.data
dfe = hvplot.ui.hvPlotExplorer(df)

## First output cell
# Play around with the Explorer until you have the plot you want

## Second input cell
dfe_settings = dfe.save()  # Returns a dict with the settings, which could be saved to a JSON file.

## Third input cell or another notebook where you have loaded the settings
# Load the settings into a new hvPlotExplorer
dfe2 = hvplot.ui.hvPlotExplorer.load(df, dfe_settings)
# or 
dfe2 = hvplot.ui.hvPlotExplorer(df, settings=dfe_settings)
```

Here a natural thing would be to only show the plot because all the settings have already been determined. Maybe with a function call like `dfe2.output()` or `dfe2.plot()`.

maximlt:
Fixes https://github.com/holoviz/hvplot/issues/764 
Fixes #787

- [x] ~~Should the accessors be patched by default when `hvplot.explorer` is called? I think so.~~ I changed my mind, less side effects are better, at least for a start.
- [x] ~~Should the explorer allow to switch the plotting backend? I'm not sure.~~ Let's keep it simple for now.
- [x] Discuss and extend the API
- [x] Add tests
- [x] Add docs. I suggest marking the explorer as **experimental** on its doc page.

image

@philippjfr  in https://github.com/holoviz/hvplot/commit/87b07b424067c4a413edbf4ab46360915c6a51e3 I've removed the `__new__` methods. They were causing `__init__` to be called twice, which wasn't really an issue as far as I could tell but was for sure undesired. I found no way to fix that by keeping them, so replaced them by a factory class method `from_data` that I use in the `explorer` function. This is a breaking change, which might be fine given that the explorer has been released recently and never documented. If this breaking change is unwanted, I could add a warning when an explorer isn't instantiated with `from_data`, or keep digging to try to fix the issue with `__new__`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -124,3 +124,6 @@ hvplot/.version
 
 # MacOS
 .DS_Store
+
+# Examples artefacts
+examples/user_guide/plot.html
--- a/doc/user_guide/index.rst
+++ b/doc/user_guide/index.rst
@@ -62,6 +62,9 @@ rather than Matplotlib.
 * `Plotting Extensions <Plotting_Extensions.html>`_
   Changing the plotting extension.
 
+* `Exploring data <Explorer.html>`_
+  Exploring data with user interface.
+
 * `Viewing <Viewing.html>`_
   Displaying and saving plots in the notebook, at the command prompt, or in scripts.
 
@@ -102,6 +105,7 @@ rather than Matplotlib.
     Interactive <Interactive>
     Widgets <Widgets>
     Plotting Extensions <Plotting_Extensions>
+    Exploring data <Explorer>
     Viewing <Viewing>
     Subplots <Subplots>
     Streaming <Streaming>
--- a/examples/user_guide/Explorer.ipynb
+++ b/examples/user_guide/Explorer.ipynb
@@ -51,7 +51,7 @@
     "\n",
     "* `save(filename, **kwargs)` to save the plot to file\n",
     "* `settings()` to obtain a dictionary of your customized settings\n",
-    "* `repr(var_name)` to obtain a formatted string to copy/paste that includes your customized settings\n",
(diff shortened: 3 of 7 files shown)


philippjfr:
Will probably look into the `__new__` issue myself. Please remind me when this is otherwise ready.

maximlt:
@philippjfr this is ready to be reviewed, if you want to have a look into `__new__`.

maximlt:
Made a few changes after some discussion:
- Raise an error when the data type given to `hvplot.explorer` is a geodataframe or an xarray object. This is because I found their support not good enough to be shared.
- Similarly the Geo tab has been removed, it did not work quite well from what we've seen (e.g.https://github.com/holoviz/hvplot/pull/810)
- renamed the `repr` method to `plot_code`, `repr` was misleading
- improved the docs in general

Hoxbro:
If #810 is not expected to be in the next release, can you add [these changes](https://github.com/holoviz/hvplot/pull/810/commits/433100feab355b345a38adba86d1ec01d4b5eea9) to this PR?

maximlt:
@philippjfr do you still want to have a look at the `__new__` thing or can I merge?

philippjfr:
Go ahead and merge, `__new__` hackery is usually ugly and unexpected so I'm fine with the current approach.

philippjfr:
Thanks for your great work here!

Issue #786, opened 2022-07-22
hvPlotExplorer issues with Style pane


Hoxbro:
An issue I have noticed when working with the Style pane of the hvPlotExplorer

Code used:
``` python
import hvplot.ui
import hvplot.pandas
import panel as pn
pn.extension()

from bokeh.sampledata import penguins

df = penguins.data
hvplot.ui.hvPlotExplorer(df)
```

### 1. The marker option is always empty
![image](https://user-images.githubusercontent.com/19758978/180438697-1c6f7cc7-a043-4667-9e7b-de154af7a463.png)

PR #809, opened 2022-07-29
Remove the marker styling option of the explorer

maximlt:
Sort of fixes https://github.com/holoviz/hvplot/issues/786

The explorer had a `marker` styling option available that in practice wasn't useful as it was not listing the markers available, in light of the upcoming release this PR simply removes this option.

Handling the styling options is a lot trickier than the other plot options:

- they vary given the selected backend
- they vary given the select kind of plot
- hvPlot for sure doesn't know anything about the values a styling option like `marker` can accept. I don't think HoloViews does either, @jlstevens do you confirm?. And I don't know if the plotting backends have APIs that expose their allowed values.
- there can be many many options to set, in particular for the interactive backends

It seems like there's more thinking and work required before exposing styling options to the explorer. My take is that for a first release the explorer is already useful enough without the styling options, hence this PR.

### DIFF
--- a/hvplot/ui.py
+++ b/hvplot/ui.py
@@ -109,8 +109,6 @@ class Style(Controls):
 
     alpha = param.Magnitude(default=1)
 
-    marker = param.Selector()
-
 
 class Axes(Controls):
 


jlstevens:
You could maybe try using the `markers` dictionary here https://github.com/holoviz/holoviews/blob/master/holoviews/plotting/bokeh/styles.py#L45 that has the marker styles that map between Bokeh and Matplotlib.

maximlt:
Thanks Jean-Luc for pointing at this dict, it may be useful in the future but I think for now I'll merge this PR, a lot more work is required before styling is made possible in the explorer.

PR #811, opened 2022-07-30
Enhancement to Fields tab in explorer

### LINKED ISSUE #785: hvPlotExplorer issues with Fields pane
Some issues I have noticed when working with the Fields pane of the hvPlotExplorer.

Code used:
``` python
import hvplot.ui
import hvplot.pandas
import panel as pn
pn.extension()

df = penguins.data
hvplot.ui.hvPlotExplorer(df)
```
### 1. Some kinds only show questionsmark in HoverTools
Kinds are `area`, `density`, and `violin`

Screenshot 2022-07-22 14 11 51

### 2. Changing y to `index` as the first thing raises an exception that is not caught.

https://user-images.githubusercontent.com/19758978/180437152-de647b60-2833-4efa-a524-bae11657595b.mp4

### 3. Kind=`errorbars` fails with `Dimension 2 not found`

image

### 4. Kind=`heatmap` collapses the UI

https://user-images.githubusercontent.com/19758978/180438111-8bcf375e-2941-4ec3-80e4-b0d191e2e576.mp4

Hoxbro:
Partially fixes #785 (2 and 4)

Indexes are now only added to `x`. For `x` and `y` because they are `param.Selector` needs to have the first value in the object specified, else they are None, which can give problem later on, e.g., when plotting.  

`self._plot` was removed because it is called when triggering `kind` on the next line.

<s>I have added the empty plot. Because when something fails, I think it is wrong to show the last working plot. The current configuration of settings does not output that plot. If people disagree with this, I could remove it from this PR and open another issue/PR to discuss it in. </s>

### DIFF
--- a/hvplot/ui.py
+++ b/hvplot/ui.py
@@ -353,12 +353,21 @@ class hvPlotExplorer(Viewer):
 
     def _populate(self):
         variables = self._converter.variables
+        indexes = getattr(self._converter, "indexes", [])
+        variables_no_index = [v for v in variables if v not in indexes]
         for pname in self.param:
             if pname == 'kind':
                 continue
             p = self.param[pname]
             if isinstance(p, param.Selector):
-                p.objects = variables
+                if pname == "x":
+                    p.objects = variables
+                else:
+                    p.objects = variables_no_index
+
+                # Setting the default value
+                if pname == "x" or pname == "y":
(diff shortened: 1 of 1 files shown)


maximlt:
Thanks @Hoxbro !

PR #720, opened 2022-03-23
Add ability to call interactive on bound functions

philippjfr:
Implements functionality requested in https://github.com/holoviz/hvplot/issues/673

```python
from bokeh import sampledata

ticker = pn.widgets.Select(options=['AAPL', 'IBM', 'GOOG', 'MSFT'], name='Ticker')

def stock_df(ticker):
    df = pd.DataFrame(getattr(sampledata.stocks, ticker))
    df['date'] = pd.to_datetime(df.date)
    return df

stock_dfi = hvplot.bind(stock_df, ticker).interactive(width=600)

stock_dfi.head(10)
```

interactive_df_select

- [x] Add tests
- [x] Automatically detect whether to use DataFrameInteractive or XArrayInteractive object

PR #812, opened 2022-08-01
Add more tests to `Interactive` and some comments

maximlt:
This PR adds various tests to `Interactive`. The aim of these tests is two-fold: cover parts that were untested and better track the underlying logic of `Interactive`. More comments in particular are meant to be added.

Interestingly I wrote these tests first based on the `0.7.3` version of hvPlot, before the support for functions as input was added. The logic at the time was slightly easier to comprehend. When writing the tests I added some utility to inspect calls to the `_clone` method at it seemed central to the way `Interactive` operates. When I finally ran the tests against *master* I had to update these tests, `_clone` is called more than it used too. The logic of `Interactive` has become more difficult to track.

### DIFF
--- a/hvplot/interactive.py
+++ b/hvplot/interactive.py
@@ -23,21 +23,25 @@ from .util import _flatten, is_tabular, is_xarray, is_xarray_dataarray
 
 def _find_widgets(op):
     widgets = []
-    op_args = list(op['args'])+list(op['kwargs'].values())
+    op_args = list(op['args']) + list(op['kwargs'].values())
     op_args = _flatten(op_args)
     for op_arg in op_args:
+        # Find widgets introduced as `widget` in an expression
         if 'panel' in sys.modules:
             if isinstance(op_arg, Widget) and op_arg not in widgets:
                 widgets.append(op_arg)
+        # TODO: Find how to execute this path?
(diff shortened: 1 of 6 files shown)


maximlt:
Ah I realized that part of the PR that implemented the support for functions as input to `Interactive` some other improvements and fixes were made (https://github.com/holoviz/hvplot/pull/720#issuecomment-1084863950), in particular:

> Accessors now always return a new interactive object, previously it would act by changing the object inplace resulting in a lot of oddness when you chained that object

Which I believe is the one that led to an increase of calls to `_clone`, but also obviously allowed to make `Interactive` more robust!

maximlt:
This PR doesn't touch any code of `interactive.py`, it simply adds comments and tests so I will merge as soon as the CI goes green. The tests really on some rather brittle features, it's spying the `_clone` method to track what `Interactive` does and it's checking that the dim transforms are correct by using their `repr`, which isn't guaranteed to be stable and I'm pretty sure I've spotted a few bugs.

Anyway this is better than before, and it has allowed me to better understand how `Interactive` works, to spot a few obvious bugs I'll fix in following PRs, and to be a little bit more confident when modifying `Interactive` in the future.

PR #815, opened 2022-08-03
Interactive: propagate max_rows

maximlt:
Interactive has a `max_rows` parameter that should be passed down to the `pn.pane.DataFrame` object created when the output of a pipeline is a dataframe. This parameter was not passed down the pipeline, which is fixed by this PR.

### DIFF
--- a/hvplot/interactive.py
+++ b/hvplot/interactive.py
@@ -280,12 +280,13 @@ class Interactive:
         return evaluate
 
     def _clone(self, transform=None, plot=None, loc=None, center=None,
-               dmap=None, copy=False, **kwargs):
+               dmap=None, copy=False, max_rows=None, **kwargs):
         plot = self._plot or plot
         transform = transform or self._transform
         loc = self._loc if loc is None else loc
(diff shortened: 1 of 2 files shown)


Hoxbro:
I would expect this to work: 

``` python
import pandas as pd
import hvplot.pandas
import panel as pn

df = pd.DataFrame(range(1000), columns=["A"])
w = pn.widgets.IntSlider(end=1000, step=10)
dfi = df.interactive(max_rows=10)
dfi.loc[dfi.index > w]
```

But I get this error:

image

Am I doing something wrong?

Hoxbro (commented):
I would expect this to work: 

``` python
import pandas as pd
import hvplot.pandas
import panel as pn

df = pd.DataFrame(range(1000), columns=["A"])
w = pn.widgets.IntSlider(end=1000, step=10)
dfi = df.interactive(max_rows=10)
dfi.loc[dfi.index > w]
```

But I get this error:

image

Am I doing something wrong?

maximlt:
Ush good catch!

Issue #763, opened 2022-06-14
Bound input function called many times on widget change


maximlt:
I'm preparing the blog post for hvPlot release and running the following example I found out that the load function was executed many times (more than a dozen) on the *Species* widget change.

```python
import panel as pn
import hvplot.pandas  # noqa
from bokeh.sampledata import penguins

pn.extension()

df = penguins.data

w_species = pn.widgets.Select(name='Species', options=list(df['species'].unique()))

w_sex = pn.widgets.MultiSelect(name='Sex', value=['MALE'], options=['MALE', 'FEMALE'])
w_body_mass = pn.widgets.FloatSlider(name='Min body mass', start=df['body_mass_g'].min(), end=df['body_mass_g'].max(), step=50)
w_y_dim = pn.widgets.RadioButtonGroup(options=['bill_depth_mm', 'flipper_length_mm']) 


def load_data(species):
    """Simluate loading data from e.g a database or from a web API."""
    print('species')
    data = df.loc[df['species'] == species]
    # time.sleep(2)
    # print(species)
    return data

dfi = hvplot.bind(load_data, w_species).interactive()

interactive = (
    dfi
    .loc[(dfi['sex'].isin(w_sex)) & (dfi['body_mass_g'] > w_body_mass)]
    .hvplot(kind='scatter', x='bill_length_mm', y=w_y_dim, by='island')
)

pn.Column(
    interactive.widgets(),
    interactive.panel(),
).servable()
```

![image](https://user-images.githubusercontent.com/35924738/173557170-f8c9730e-104e-4011-9f23-570c52204487.png)

PR #818, opened 2022-08-03
Interactive: only register once the function parameters watchers

maximlt:
Alternative implementation (https://github.com/holoviz/hvplot/pull/768) to fix https://github.com/holoviz/hvplot/issues/763

The problem in https://github.com/holoviz/hvplot/issues/763 was two-fold:

1. the callback used to update the original pipeline object (`self._obj`, e.g. a DataFrame) was registered every time a new `Interactive` instance was created. Even a small pipeline creates many instances so that led to many calls being made when one of the parameters controlling the input function changed.
2. only the last callback registered was actually updating the right `self._obj` object. Even if all the instances share the same object, setting a new object in the update callback would not lead to that new object being set on all the instances, just on that particular one. This is why in the previous attempts made to fix this issue, which I've also tried, only setting the callbacks on the root instance (`depth == 0`) didn't work; the Interactive instance being updated isn't the one that drives the display.

To fix 2., I've added a small abstraction layer using a list as the data holder and properties to access/edit the object held, so that the underlying object can be shared across all the Interactive instances created in a pipeline. This means that any change made to `_obj` by any Interactive instance will be visible to all the other instances.
To fix 1. I've used `depth == 0` to only register the update watchers on the root instance, which now works thanks to the fix for 2.

### DIFF
--- a/hvplot/interactive.py
+++ b/hvplot/interactive.py
@@ -157,27 +157,25 @@ class Interactive:
             if subcls.applies(obj):
                 clss = subcls
         inst = super(Interactive, cls).__new__(clss)
-        inst._obj = obj
+        inst._shared_obj = kwargs.get('_shared_obj', [obj])
         inst._fn = fn
         return inst
 
     @classmethod
     def applies(cls, obj):
         """
-        Subclasses must implement applies and return a boolean to indicates
+        Subclasses must implement applies and return a boolean to indicate
         wheter the subclass should apply or not to the obj.
         """
         return True
 
     def __init__(self, obj, transform=None, fn=None, plot=False, depth=0,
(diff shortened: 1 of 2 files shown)


philippjfr:
I can follow this approach better than the other one, it doesn't monkey-patch the obj, the tests are passing and this PR doesn't have merge conflicts to I'm +1 on merging this PR. Updating a shared list is a little bit ugly but I don't know of a better construct in Python and a class to hold the shared reference seems overkill. Feel free to merge.

maximlt:
> a class to hold the shared reference seems overkill

Indeed I started with a class but it felt overkill, maybe a dataclass would be a good fit but that's not available in Python 3.6 that is still supported by hvPlot. Thanks merging.

PR #822, opened 2022-08-05
Adding docstring to interactive

Hoxbro:
An attempt at adding the docstring for interactive. I have kept the class docstring in the code, but because it is heavily developer-focused, I have removed it when using `?` in a notebook.

### DIFF
--- a/hvplot/interactive.py
+++ b/hvplot/interactive.py
@@ -49,90 +49,144 @@ def _find_widgets(op):
     return widgets
 
 
+"""
+How Interactive works
+---------------------
+
+`Interactive` is a wrapper around a Python object that lets users create
+interactive pipelines by calling existing APIs on an object with dynamic
+parameters or widgets.
+
+An `Interactive` instance watches what operations are applied to the object.
+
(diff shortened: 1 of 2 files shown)


MarcSkovMadsen:
When I have been making these changes I've normally been doing them on the `Class` because VS Code tooltips use them over `__init__` if I am not actively typing the parameters `MyClass(name=...)`.

How does your approach work in VS Code @Hoxbro ?

hvplot/interactive.py
```diff
@@ -365,6 +365,22 @@ def get_ax():
         return get_ax
 
     def __call__(self, *args, **kwargs):
+        """
```
MarcSkovMadsen, line 368: See comment in post about whether the docstring should be on the Class, `__init__` or `__call__` level.

maximlt:
Thanks a lot @Hoxbro , I've just re-arranged the docstrings, which mostly consisted in moving the *How it works* section out of the Interactive docstrings. Both the Interactive class and its `__call__` method have a docstring, so that it will appear in an IDE like VSCode in both situations.

PR #824, opened 2022-08-07
add script.* to .gitignore

MarcSkovMadsen:
I very often have `script.py` or `script.ipynb` in my root folder to explore something while developing or to come examples from discourse into while helping users. I've sometimes added these to PRs by mistake. Would like to avoid.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -124,3 +124,5 @@ hvplot/.version
 
 # MacOS
 .DS_Store
+
+script.*


MarcSkovMadsen:
## Example Use Case

Here I need to see if the documentation I'm adding is working. I use a `script.py` file for that.

image

maximlt:
This is very much tailored to your workflow, and opens the door to other contributors adding their own script name to the gitignore, making it not maintainable. There are many solutions to cope with this. Mine is the following. I have in each repo I work on a folder called `mymess` that is ignored globally by git. All my mess goes in there and new files don't show up in VS Code.

Issue #825, opened 2022-08-07
Document hvplot.bind on website and with docstring


MarcSkovMadsen:
I cannot find anything on website https://hvplot.holoviz.org/search.html?q=bind

![image](https://user-images.githubusercontent.com/42288570/183304108-431cb4f3-7e6c-4dc5-b42a-1410fe5ee47f.png)

And it has no docstring either.

maximlt:
I am surprised the search did not return a better result. But `hvplot.bind` **is** documented here: https://hvplot.holoviz.org/user_guide/Interactive.html#functions-as-inputs

MarcSkovMadsen:
I will close this issue. But someone should one day open an issue about the bad search results on the HoloViz sites. The search returns surprisingly few or irrelevant results.

PR #826, opened 2022-08-07
Add docstring to hv.bind

MarcSkovMadsen:
Partially addresses #825

### DIFF
--- a/hvplot/__init__.py
+++ b/hvplot/__init__.py
@@ -217,8 +217,23 @@ def bind(function, *args, **kwargs):
     
     Reference: https://hvplot.holoviz.org/user_guide/Interactive.html#functions-as-inputs
 
-    Example
+    Arguments
+    ---------
+    function : callable
+        The function to bind constant or dynamic args and kwargs to.
+    args : object, param.Parameter, panel.widget.Widget, or ipywidget
+        Positional arguments to bind to the function.
+    kwargs : object, param.Parameter, panel.widget.Widget, or ipywidget
+        Keyword arguments to bind to the function.
+
+    Returns
     -------
+    Returns a new function with the args and kwargs bound to it and
+    annotated with all dependencies. This function has an `interactive`
+    attribute that can be called to instantiate an `Interactive` pipeline.
+
+    Examples
+    --------
 
     Develop your **algorithm** or data extraction method with the tools you know and love.
 
@@ -247,21 +262,6 @@ def bind(function, *args, **kwargs):
     In a notebook or data app you can now select the appropriate `alpha` and `top` values via
     widgets and see the `top` results of the algorithm in a table depending on the value of `alpha`
     selected.
-
-    Arguments
-    ---------
-    function : callable
-        The function to bind constant or dynamic args and kwargs to.
-    args : object, param.Parameter, panel.widget.Widget, or ipywidget
-        Positional arguments to bind to the function.
-    kwargs : object, param.Parameter, panel.widget.Widget, or ipywidget
-        Keyword arguments to bind to the function.
-
-    Returns
-    -------
-    Returns a new function with the args and kwargs bound to it and
-    annotated with all dependencies. This function has an `interactive`
-    attribute that can be called to instantiate an `Interactive` pipeline.
     """
     bound = _pn.bind(function, *args, **kwargs)
     bound.interactive = lambda **kwargs: Interactive(bound, **kwargs)
--- a/hvplot/__init__.py
+++ b/hvplot/__init__.py
@@ -3,7 +3,7 @@ hvPlot makes data analysis and visualization simple
 ===================================================
 
 hvPlot provides a familiar, high-level API for interactive data exploration and visualization,
-based on the 🐼 Pandas `.plot` API and the innovative `.interactive` API.
+based on the Pandas `.plot` API and the innovative `.interactive` API.
 
 hvPlot
 
@@ -260,7 +260,8 @@ def bind(function, *args, **kwargs):
     Returns
     -------
     Returns a new function with the args and kwargs bound to it and
-    annotated with all dependencies.
+    annotated with all dependencies. This function has an `interactive`
+    attribute that can be called to instantiate an `Interactive` pipeline.
     """
     bound = _pn.bind(function, *args, **kwargs)
     bound.interactive = lambda **kwargs: Interactive(bound, **kwargs)
--- a/hvplot/__init__.py
+++ b/hvplot/__init__.py
@@ -235,7 +235,7 @@ def bind(function, *args, **kwargs):
     ...     # Loading data for the given experiment ...
     ...     return pd.DataFrame({"x": list(range(0,experiment))})
 
-    >>> hvplot.bind(get_data,experiment=experiment).interactive().head(n=rows)
+    >>> hvplot.bind(get_data, experiment=experiment).interactive().head(n=rows)
 
     In a notebook or data app you can now use the `experiment` and `rows` sliders to visualize
     your data interactively as a line plot.
--- a/hvplot/__init__.py
+++ b/hvplot/__init__.py
@@ -207,41 +207,30 @@ Store._backend_switch_hooks.append(_hook_patch_docstrings)
 
 extension = hvplot_extension
 
-@_wraps(_pn.bind)
 def bind(function, *args, **kwargs):
     """
-    Returns a *reactive* function that *depends* on widgets, parameters or Python objects.
-    This means that the "reactive" function can be automatically invoked whenever the underlying
-    widget or parameter values change. The widgets can be Panel or IpyWidgets.
+    Returns a *reactive* function that can be used to start your `.interactive` pipeline by running
+    a model or loading data depending on inputs from widgets, parameters or python objects.
+    The widgets can be Panel or IpyWidgets.
     
-    `hv.bind` can be **used to start your `.interactive` pipeline** by loading data depending
-    on widgets or pameters values.
-    
-    Reference: https://github.com/holoviz/hvplot/issues/825
+    Reference: https://hvplot.holoviz.org/user_guide/Interactive.html#functions-as-inputs
 
-    :Example:
+    Example
+    -------
 
     >>> import pandas as pd
     >>> import panel as pn
-
     >>> import hvplot
-
-    >>> pn.extension()
-
-    >>> experiment = pn.widgets.IntSlider(value=5, start=0, end=10, step=1, name="Experiment")
-    >>> rows = pn.widgets.IntSlider(value=5, start=0, end=10, step=1, name="Rows")
-
+    >>> import numpy as np
+    >>> alpha = pn.widgets.FloatSlider(value=0.5, start=0, end=1.0, step=0.1, name="Alpha")
+    >>> top = pn.widgets.RadioButtonGroup(value=10, options=[5, 10, 25], name="Top")
     >>> def algorithm(alpha):
-    ...         # Running an algorithm that uses alpha...
-    ...         return pd.DataFrame({"output": np.array([0, 3, 5, 2]) ** alpha})
+    ...     # Running an algorithm that uses alpha...
+    ...     return pd.DataFrame({"output": (np.array(range(0,100)) ** alpha)*50})
+    >>> table = hvplot.bind(algorithm, alpha=alpha).interactive().head(n=top)
 
-    >>> hvplot.bind(get_data, experiment=experiment).interactive().head(n=rows)
-
-    In a notebook or data app you can now use the `experiment` and `rows` sliders to visualize
-    your data interactively as a line plot.
-
-    This function is the same as `panel.bind`, but extended by adding the `.interactive` method to
-    the *reactive* function returned.
+    In a notebook or data app you can now see the `alpha` and `top` widgets as well as the `top`
+    results of the algorithm when run with the `alpha` value as input. 
 
     Arguments
     ---------
--- a/hvplot/__init__.py
+++ b/hvplot/__init__.py
@@ -211,7 +211,7 @@ def bind(function, *args, **kwargs):
     """
     Returns a *reactive* function that can be used to start your `.interactive` pipeline by running
     a model or loading data depending on inputs from widgets, parameters or python objects.
-    The widgets can be Panel or IpyWidgets.
+    The widgets can be Panel or ipywidgets.
     
     Reference: https://hvplot.holoviz.org/user_guide/Interactive.html#functions-as-inputs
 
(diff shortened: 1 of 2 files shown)


hvplot/__init__.py
```diff
@@ -209,6 +209,54 @@ def _hook_patch_docstrings(backend):
 
 @_wraps(_pn.bind)
 def bind(function, *args, **kwargs):
+    """
+    Returns a *reactive* function that *depends* on widgets, parameters or Python objects.
+    This means that the "reactive" function can be automatically invoked whenever the underlying
+    widget or parameter values change. The widgets can be Panel or IpyWidgets.
```
maximlt, line 215: `hvplot.bind` is documented as a way to provide a *function as input* to an interactive pipeline, this doesn't really reflect that.

hvplot/__init__.py
```diff
@@ -209,6 +209,54 @@ def _hook_patch_docstrings(backend):
...
+    
+    `hv.bind` can be **used to start your `.interactive` pipeline** by loading data depending
+    on widgets or pameters values.
+    
+    Reference: https://github.com/holoviz/hvplot/issues/825
+
+    :Example:
```
maximlt, line 222: I've seen you sometimes use `:Example"` and sometimes `Example:`. What's the difference?

hvplot/__init__.py
```diff
@@ -209,6 +209,54 @@ def _hook_patch_docstrings(backend):
...
 def bind(function, *args, **kwargs):
+    """
+    Returns a *reactive* function that *depends* on widgets, parameters or Python objects.
+    This means that the "reactive" function can be automatically invoked whenever the underlying
+    widget or parameter values change. The widgets can be Panel or IpyWidgets.
+    
+    `hv.bind` can be **used to start your `.interactive` pipeline** by loading data depending
+    on widgets or pameters values.
+    
+    Reference: https://github.com/holoviz/hvplot/issues/825
```
maximlt, line 220: https://hvplot.holoviz.org/user_guide/Interactive.html#functions-as-inputs

hvplot/__init__.py
```diff
@@ -209,6 +209,54 @@ def _hook_patch_docstrings(backend):
...
+    ...     # Loading data for the given experiment ...
+    ...     return pd.DataFrame({"x": list(range(0,experiment))})
+
+    >>> hvplot.bind(get_data, experiment=experiment).interactive().head(n=rows)
+
+    In a notebook or data app you can now use the `experiment` and `rows` sliders to visualize
+    your data interactively as a line plot.
+
+    This function is the same as `panel.bind`, but extended by adding the `.interactive` method to
+    the *reactive* function returned.
```
maximlt, line 244: The fact that the returned object has an `interactive` "method" is the whole point of using `hvplot.bind` instead of `pn.bind`. This should be explained before.

hvplot/__init__.py
```diff
@@ -209,6 +209,54 @@ def _hook_patch_docstrings(backend):
...
+    >>> rows = pn.widgets.IntSlider(value=5, start=0, end=10, step=1, name="Rows")
+
+    >>> def get_data(experiment):
+    ...     # Loading data for the given experiment ...
+    ...     return pd.DataFrame({"x": list(range(0,experiment))})
+
+    >>> hvplot.bind(get_data, experiment=experiment).interactive().head(n=rows)
+
+    In a notebook or data app you can now use the `experiment` and `rows` sliders to visualize
+    your data interactively as a line plot.
```
maximlt, line 241: The example will display a dataframe, not a line plot.

hvplot/__init__.py
```diff
@@ -209,6 +209,54 @@ def _hook_patch_docstrings(backend):
...
+    >>> import hvplot
+
+    >>> pn.extension()
+
+    >>> experiment = pn.widgets.IntSlider(value=5, start=0, end=10, step=1, name="Experiment")
+    >>> rows = pn.widgets.IntSlider(value=5, start=0, end=10, step=1, name="Rows")
+
+    >>> def get_data(experiment):
+    ...     # Loading data for the given experiment ...
+    ...     return pd.DataFrame({"x": list(range(0,experiment))})
```
maximlt, line 236: I understand that you would like the example to be executable, yet I find it not very clear, it's returning a dataframe with one column and whose length depends on `experiment`, that seems unusual. Here's my suggestion, that requires numpy to be imported.

```suggestion
    >>> def algorithm(alpha):
    ...         # Running an algorithm that uses alpha...
    ...         return pd.DataFrame({"output": np.array([0, 3, 5, 2]) ** alpha})
```

This points to an important point: the docstring should tell why one would need to use `hvplot.bind`, i.e. as the data you want to explorer in an interactive pipeline can be the result of some other preprocessing pipeline (running a simulation, getting data from a database, etc.).

hvplot/__init__.py
```diff
@@ -209,6 +209,54 @@ def _hook_patch_docstrings(backend):
...
+    :Example:
+
+    >>> import pandas as pd
+    >>> import panel as pn
+
+    >>> import hvplot
+
+    >>> pn.extension()
+
+    >>> experiment = pn.widgets.IntSlider(value=5, start=0, end=10, step=1, name="Experiment")
```
maximlt, line 231: Depending on the function used as input, we'll have to make sure the right widget is used. In this case I'd say a `Select` would have been more appropriate.

hvplot/__init__.py
```diff
@@ -209,6 +209,54 @@ def _hook_patch_docstrings(backend):
...
+    >>> rows = pn.widgets.IntSlider(value=5, start=0, end=10, step=1, name="Rows")
```
maximlt, line 232: `step=1` isn't required.

hvplot/__init__.py
maximlt: Is that related in any way to the Numpy docstring convention?

hvplot/__init__.py
```diff
@@ -209,6 +209,54 @@ def _hook_patch_docstrings(backend):
...
+        The function to bind constant or dynamic args and kwargs to.
+    args: object, param.Parameter, panel.widget.Widget, or ipywidget
+        Positional arguments to bind to the function.
+    kwargs: object, param.Parameter, panel.widget.Widget, or ipywidget
+        Keyword arguments to bind to the function.
+
+    Returns
+    -------
+    Returns a new function with the args and kwargs bound to it and
+        annotated with all dependencies.
```
maximlt, line 258: Extra space

hvplot/__init__.py
```diff
@@ -209,6 +209,54 @@ def _hook_patch_docstrings(backend):
...
+    Reference: https://github.com/holoviz/hvplot/issues/825
+
+    :Example:
+
+    >>> import pandas as pd
+    >>> import panel as pn
+
+    >>> import hvplot
+
+    >>> pn.extension()
```
maximlt, line 229: Is there a reason for the spaces between these lines?

hvplot/__init__.py
```diff
@@ -209,6 +209,54 @@ def _hook_patch_docstrings(backend):
 
 @_wraps(_pn.bind)
```
maximlt, line 210: Doesn't adding a docstring mean that the usage of `wraps` becomes useless? If it's not removed, does it mean that `bind` defined here is still going to inherit from the docstring of `pn.bind`? I'm not quite sure as how this behaves.

hvplot/__init__.py
MarcSkovMadsen: I run the example to test it works. I always use `black` to autoformat. I did not think about that this creates spaces in the examples when I copy-paste them back.

I will remove the space.

hvplot/__init__.py
MarcSkovMadsen: I have not given it a lot of thought previously. I've seen different ways of doing it in different places.

I would like to update to 

```rest
Example
---------
```

Because that is in line with the Parameters and Returns section and looks really well organized as a tooltip in VS Code.

hvplot/__init__.py
MarcSkovMadsen: Will update to this link. Thanks

hvplot/__init__.py
MarcSkovMadsen: Good catch

hvplot/__init__.py
MarcSkovMadsen: good point

hvplot/__init__.py
MarcSkovMadsen: of course. will change.

hvplot/__init__.py
MarcSkovMadsen: Thanks. I did not think it mattered. But I think you are right.

hvplot/__init__.py
MarcSkovMadsen: will remove

hvplot/__init__.py
MarcSkovMadsen: will remove

hvplot/__init__.py
MarcSkovMadsen: I think you are right, but will test.

Wraps does more than adding a docstring dynamically.

Pyright and VS Code has a hack for static typing to also use the docstring and function signature.

hvplot/__init__.py
MarcSkovMadsen: As experiment was changed to alpha, I kept the IntSlider for alpha. But changed the number of rows select to a RadioButtonGroup. That makes sense to me.

hvplot/__init__.py
MarcSkovMadsen: I'm not sure I understand. Could you be more specific?

hvplot/__init__.py
MarcSkovMadsen: After having looked at the code I restructured it a bit. I tried imagining being a user studying the example or me trying to to explain the example to a colleague. And for that I would like a bit of space to separate imports and distinct sections of the code.

Hope you like it. Otherwise let me know.

hvplot/__init__.py
MarcSkovMadsen: Ahh. I thought return docstrings also had indentations (spaces) similar to Parameters.

hvplot/__init__.py
```diff
@@ -2,15 +2,18 @@
 hvPlot makes data analysis and visualization simple
 ===================================================
 
-hvPlot provides an easy to use high-level API for visualization, based on the 🐼 Pandas .plot() API
-that works across a wide range of data sources and plotting backends.
+hvPlot provides a familiar, high-level API for interactive data exploration and visualization,
+based on the 🐼 Pandas `.plot` API and the innovative `.interactive` API.
```
maximlt, line 6:
```suggestion
based on the Pandas `.plot` API and the innovative `.interactive` API.
```

hvplot/__init__.py
```diff
@@ -207,8 +208,60 @@ def _hook_patch_docstrings(backend):
...
+        The function to bind constant or dynamic args and kwargs to.
+    args : object, param.Parameter, panel.widget.Widget, or ipywidget
+        Positional arguments to bind to the function.
+    kwargs : object, param.Parameter, panel.widget.Widget, or ipywidget
+        Keyword arguments to bind to the function.
+
+    Returns
+    -------
+    Returns a new function with the args and kwargs bound to it and
+    annotated with all dependencies.
```
maximlt, line 263:
```suggestion
    annotated with all dependencies. This function has an `interactive`
    attribute that can be called to instantiate an `Interactive` pipeline.
```

maximlt (commented):
Thanks for taking into account my comments! Really appreciate. I've just made a couple of suggestions I'll approve directly as they're minor.

I think my last question would be, I'm used to see the `Examples` section being at the bottom of the docstring. This is what the Numpy docstring guide seems to recommend: https://numpydoc.readthedocs.io/en/latest/format.html#examples.

Also as I just realize that, you've named the parameter section `Arguments` while the Numpy guide recommends `Parameters`.

Is there a reason why you're not following it? I feel that following the Numpy styling guide could only help. It's not that we're very good at following style guides, but you know, we can only get better :D

MarcSkovMadsen:
> Thanks for taking into account my comments! Really appreciate. I've just made a couple of suggestions I'll approve directly as they're minor.
> 
> I think my last question would be, I'm used to see the `Examples` section being at the bottom of the docstring. This is what the Numpy docstring guide seems to recommend: https://numpydoc.readthedocs.io/en/latest/format.html#examples.
> 
> Also as I just realize that, you've named the parameter section `Arguments` while the Numpy guide recommends `Parameters`.
> 
> Is there a reason why you're not following it? I feel that following the Numpy styling guide could only help. It's not that we're very good at following style guides, but you know, we can only get better :D

I'm learning here :-). 

- I will rename Parameters to Arguments. I think I copied it from some existing hvPlot code?
- Personally I prefer to meet examples before documentation as a new user. People tend to want to see how easy it is, start from a working example. Similar to preferring you tube videos over written reference manuals. But I can see that in a situation where the docstrings end up as reference documentation on a web site, then it makes sense to have the Parameters and Returns first. Should i change @maximlt ?

MarcSkovMadsen (commented):
😲. My 🐼 emoji is gone 😊

maximlt:
> 😲. My 🐼 emoji is gone. 😊

Even a cute little 🐼 can become a 🔥  topic!  I've preferred removing it to avoid getting ⚡ ! 😄 🙏 

> Personally I prefer to meet examples before api documentation as a new user. People tend to want to see how easy it is to start from a working example. Similar to preferring youtube videos over written reference manuals. But I can see that in a situation where the docstrings end up as reference documentation on a web site, then it makes sense to have the Parameters and Returns first.

Well I think what is important in what you wrote is *as a new user*. Docstrings aren't meant to be read only by new users, I think. For instance, if I take quite an extreme case but that I think is fine for comparison as it also follows the flat API offered by hvPlot, I often read the docstring of `pandas.read_csv` even if I'm definitively not using it for the first time and I exactly know what it does. But I have to do so as it has a long signature and I don't necessarily remember all its parameters, what types they take, etc. I think we're in a similar situation for hvPlot's docstrings in general. I imagine people that want to make a line plot know what a line plot is.

maximlt:
Thanks a lot @MarcSkovMadsen for this PR and the discussion around it! I've replaced `Arguments` by `Parameters` and moved the `Examples` section to the bottom, conventions make everyone's life easier.

MarcSkovMadsen:
I think most users have a basic idea of what a line plot is. But they might not have an idea what what hvPlot and a hvPlot `line` plot is and how it works. Code is being read more often than written. And in my context by lots of users that may not use hvPlot themselves but they have to work with code a colleague wrote. For example to quickly fix bugs or extend the functionality.

Thus it is just very important that documentation meets beginners. If you listen to the latest podcast by "Talk Python to me" they are talk about how a big part of Pydantics and FastApis success is that they started with good documentation over good features or good implementations. And how Sebastian is so good a meeting beginning users that might not even know Python very well.

There are so many more potential new users to meet than existing super users. Its the newbees you loose. The super users know how to find their way around.

PR #827, opened 2022-08-07
Adds docstrings to the `core` module

MarcSkovMadsen:
This PR aims to add better docstring for users. It does this by providing more descriptive text, examples and reference links similar to the docstrings recently added to Panel.

### DIFF
--- a/hvplot/plotting/core.py
+++ b/hvplot/plotting/core.py
@@ -598,6 +598,42 @@ class hvPlotTabular(hvPlotBase):
         return self(x, y, text=text, kind='labels', **kwds)
 
 class hvPlot(hvPlotTabular):
+    """hvPlot provides methods like `.line`, `step`, `scatter` etc. to plot your tabular
+    data source. These methods works just like the familiar Pandas `.plot` function.
+    
+    `hvPlot` supports a wide range of data sources and plotting backends.
+
+    `hvPlot` is available as `.hvplot` on your data source after you have imported the appropriate
+    backend.
+    
+    Example: `.hvplot`   
+
+    >>> import pandas as pd
+    >>> import hvplot.pandas
+
+    >>> x=y=list(range(0, 10))
+    >>> df = pd.DataFrame({"observation": x, "value": y})
+
+    >>> df.hvplot.scatter(x="observation", y="value")
+    >>> # or
+    >>> df.hvplot(x="observation", y="value", kind="scatter")
+    
+    Example: `hvPlot`
+
+    >>> import pandas as pd
+    >>> from hvplot import hvPlot
+
+    >>> x=y=list(range(0, 10))
+    >>> df = pd.DataFrame({"observation": x, "value": y})
+
+    >>> hvPlot(df).scatter(x="observation", y="value")
+
+    We recommend using 
+    
+    - `.hvplot` in a notebook environment as its very familiar to DataFrame users.
+    - `hvPlot` in editors like VS Code because it provides better tooltips and TAB completion.
+
+    """
     __all__ = [
         'line',
         'step',
@@ -632,17 +668,25 @@ class hvPlot(hvPlotTabular):
         """
         Image plot
 
+        You can very often use `image` to display an x, y grid, An example could be geographic
+        data with `latitude` and `longitude` coordinates.
+
+        Reference: https://hvplot.holoviz.org/reference/xarray/image.html
+
+        :example:
+
         Parameters
         ----------
-        x, y : string, optional
-            The coordinate variable along the x- and y-axis
+        x : string, optional
+            The coordinate variable along the x-axis
+        y : string, optional
+            The coordinate variable along the y-axis
         z : string, optional
             The data variable to plot
         colorbar: boolean
             Whether to display a colorbar
-        **kwds : optional
-            Keyword arguments to pass on to
-            :py:meth:`hvplot.converter.HoloViewsConverter`.
+        kwds : optional
+            To see all the keyword arguments available, run `hvplot.help('image')`.
         Returns
         -------
         obj : HoloViews object


MarcSkovMadsen:
Would you be willing to accept a larger PR in the style of what I have started here @maximlt or @philippjfr ?

hvplot/plotting/core.py
```diff
@@ -632,17 +668,30 @@ def image(self, x=None, y=None, z=None, colorbar=True, **kwds):
...
+        Example:
+
+        >>> import hvplot.xarray
+        >>> import xarray as xr
+        >>> ds = xr.tutorial.open_dataset('air_temperature')
+        >>> ds.hvplot.image(x='lon', y='lat', z='air', groupby='time', cmap='kbc_r')
+
         Parameters
         ----------
-        x, y : string, optional
```
MarcSkovMadsen, line 637: x,y does not display nicely in VS Code. It expects one argument. Not a comma separated list.

hvplot/plotting/core.py
```diff
@@ -632,17 +668,30 @@ def image(self, x=None, y=None, z=None, colorbar=True, **kwds):
...
-            The coordinate variable along the x- and y-axis
+        x : string, optional
```
MarcSkovMadsen, line 685: Do we know these have to be strings. What if it is a `DataFrame` with `int` column names?

MarcSkovMadsen: Would you be happy if I added type annotations? 

`def image(self, x: str | None=None, y: str | None=None, z: str | None=None, colorbar: bool=True, **kwds):`

maximlt: That is a question that deserves its own issue :)

maximlt:
>     We recommend using 
>   
>    - `.hvplot` in a notebook environment as its very familiar to DataFrame users.
 >   - `hvPlot` in editors like VS Code because it provides better tooltips and TAB completion.

Recommending different APIs given what editors people use make me sad :( I think we'll have to discuss this more before committing to this.

hvplot/plotting/core.py
philippjfr: Definitely a separate discussion, issue and PR. I won't object to typing hvPlot though.

philippjfr:
>Recommending different APIs given what editors people use make me sad :( I think we'll have to discuss this more before committing to this.

Strongly agree, I can see the issue with tab-completion but on balance I'm a strong -1 on recommending two separate APIs based on your editor.

MarcSkovMadsen:
For this PR I will then 

- let out the recommendation of using `hvPlot` api in VS Code. I will use it my self though because it makes a world of a difference for my productivity and for the ability of my colleagues to follow my code.
- not add type annotations

Would you be willing to look positively at reviewing and accepting a larger PR then?

philippjfr:
Sounds good and to be clear I won't object to adding some discussion about different APIs and usage patterns in the docs. Just not excited about recommending different APIs directly in the docstrings.

> Would you be willing to look positively at reviewing and accepting a larger PR then?

Absolutely.

MarcSkovMadsen:
Just for the record. Besides better intellisense in VS Code using `hvPlot` over `.hvplot` will also avoid linters complaining about *missing imports* from `import hvplot.pandas`.

MarcSkovMadsen:
I've added a *reference example* to the top, opening post and pushed the code. Feel free to comment on the code and guide me. It will be a lot of work to do and if you don't like it @maximlt and @philippjfr I would have wasted a lot of time. Thanks.

maximlt:
I'm on a holidays today, I'll have a closer look at it later this week. What I can say from a quick glance is that it looks like a good start.

A very important point imo is that you seem to be heading towards writing docstrings that are long enough to be part, or even entirely constitute, the reference gallery page of the documented plot. There shouldn't be two separate systems of documentation (docstring and reference gallery) to maintain at the same time, otherwise I wouldn't consider that as an extra burden on a documentation we already struggle to maintain.
Your example docstring looks like the docstrings of Pandas plotting methods (e.g. https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.plot.line.html), which render pretty well in a website I think. The difference with hvPlot's site is that they're part of the API reference, while on hvPlot they're part of the reference gallery.
Therefore I feel that at this stage it would be wise to try to see if we could build the reference gallery page of the `line` method from the docstring you've added, there must be a way to do so but someone will have to find how! It doesn't have to be you though, that could be me but I can't commit to that right now (focusing on releasing a bug fix version of hvPlot) so that will put your work on hold. Thoughts on that @philippjfr ?

MarcSkovMadsen:
Thanks for your comments. I would say that Documentation in different places serves different purposes and is really, really needed for `hvPlot` to be useful for users. 

The Reference Gallery should have lots of different examples in my opinion. You cannot really outsource to the user guide how to use some of the fundamental powers of `by`, `groupby` and ability to sprinkle in widgets that make hvplot so powerful. New users might end up in the reference Gallery via a Google Search and never understand the power of hvPlot. They just look at it and think this is like Pandas `.plot` but with less documentation and not my favorite plotting library. Why should I use it?

I am not so worried about the docstrings in code. After all there are not so many methods to document and keep updated. And the API is pretty stable.

MarcSkovMadsen:
> Which part of Pandas' documentation gives you this impression exactly? That page https://pandas.pydata.org/pandas-docs/stable/user_guide/visualization.html#chart-visualization?

This https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.plot.line.html and similarly the other links I've referenced for line plots.

  - Bokeh: https://docs.bokeh.org/en/latest/docs/reference/models/glyphs/line.html
  - HoloViews: https://holoviews.org/reference/elements/bokeh/Curve.html
  - Pandas: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.plot.line.html
  - Plotly: https://plotly.com/python/line-charts/
  - Matplotlib: https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.plot.html
  - Wiki: https://en.wikipedia.org/wiki/Line_chart

The Pandas link is actually populated by their docstring.

If you are new to plotting then you wont have a chance with `hvPlot` as the documentation is so sparse. You really have to have all the the basic knowledge and intuition from Matplotlib and Pandas .plot first before you start to look at hvPlot currently.

hvplot/plotting/core.py
```diff
@@ -12,62 +14,95 @@
...
     __all__ = []
 
     def __init__(self, data, custom_plots={}, **metadata):
-        if 'query' in metadata:
-            data = data.query(metadata.pop('query'))
-        if 'sel' in metadata:
-            data = data.sel(**metadata.pop('sel'))
-        if 'isel' in metadata:
-            data = data.isel(**metadata.pop('isel'))
+        "Hello"
```
maximlt, line 71: Hello!

hvplot/plotting/core.py
```diff
@@ -12,62 +14,95 @@
...
+        if "query" in metadata:
+            data = data.query(metadata.pop("query"))
+        if "sel" in metadata:
+            data = data.sel(**metadata.pop("sel"))
+        if "isel" in metadata:
+            data = data.isel(**metadata.pop("isel"))
```
maximlt, line 77: Oh be careful it seems your autoformatter was busy doing some extra work!

hvplot/plotting/core.py
```diff
@@ -329,26 +414,36 @@ def scatter(self, x=None, y=None, **kwds):
...
+
+        References
+        ----------
+
+        - Bokeh: http://docs.bokeh.org/en/latest/docs/user_guide/plotting.html#scatter-markers
+        - HoloViews: http://holoviews.org/reference/elements/matplotlib/Scatter.html
+        - Pandas: https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.plot.scatter.html
+        - Plotly: https://plotly.com/python/line-and-scatter/
+        - Matplotlib:  https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.scatter.html
+        - Wiki: https://en.wikipedia.org/wiki/Scatter_plot
```
maximlt, line 444: Do you know if there's a way to automatically check that these links don't 404?

hvplot/plotting/core.py
MarcSkovMadsen: removed

hvplot/plotting/core.py
MarcSkovMadsen: Yes. I got a new way of working (cloud instead of laptop) and setup VS Code to autoformat on save using `black`. Will turn it off again.

MarcSkovMadsen: But could we start using `isort` and `black` soon on these projects. Just makes your life as a developer so much easier.

hvplot/plotting/core.py
MarcSkovMadsen: Yes. Python and pytest. I will add a test.

MarcSkovMadsen: Please note that for me these links are the only way to understand hvPlot. They are a life saver. The current documentation is so sparse that I have no idea on how it should work, what options are available or how things can be combined. The only way to understand hvPlot is by getting the perspective from the underlying frameworks.

But finding the alternative information takes a long time because things are not called the same across. So you have to spend a considerable amount of googling.

hvplot/plotting/core.py
```diff
@@ -357,57 +452,104 @@ def area(self, x=None, y=None, y2=None, stacked=True, **kwds):
...
+
+        Example
+        -------
+
+        >>> import hvplot.pandas
+        >>> from bokeh.sampledata.degrees import data
+        >>> data.hvplot.area(x='Year', y='Computer Science',
+        ...     label='% of Computer Science Degrees Earned by Women',
+        ...     ylim=(0, 100), width=500, height=400
+        ... ).opts(bgcolor='goldenrod')
```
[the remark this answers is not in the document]
MarcSkovMadsen: I found the example in the reference documentation with `.opts`. I searched for a background color argument to `area` but could not find it.

I'm in doubt.  `hvPlot` does not talk a lot about `.opts`. But you will end up using them eventually I would claim unless you are just exploring.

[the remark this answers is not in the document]
MarcSkovMadsen: I'm also thinking more and more that I should go over the examples once more and remove any dependency on external bokeh sample data. Its hard to understand the available fields of some data that is just imported. Its easier to understand if its explicitly defined.

Probably I should also fine tune the examples to use nice colors etc. I.e. give the user something that can really be used.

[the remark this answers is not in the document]
MarcSkovMadsen: I'm actually not sure I like the example. Its not how you would normally start out using hvPlot. Its more like a trick how to get a stacked area chart that sums to 100%.

MarcSkovMadsen:
Hi @maximlt 

I've

- Added an `urls_are_valid` pytest. 
- Reviewed the current contribution.

Let me know if you think this could get into the next release of hvPlot. Then I will continue the docstring work in a new PR.

MarcSkovMadsen:
One note on signatures and docstrings. I can see some key parameters are not in the signature or docstring.

Unless somewhere tells me otherwise (@maximlt , @philippjfr ) the guiding principles for me are

- Special and very important parameters should be in the signature and docstring. Around 7 parameters for a signature is fine. A bit more for docstrings is ok. But for now I will focus on adding them to the docstring as changing the function signature is more risky and needs more careful review.

I say this because I can see for example the special `_kind_options` are often neither documented or in the signature.

image

maximlt:
Thanks @MarcSkovMadsen ! Excellent improvements to the docstrings!

Merging as is. I think there are still some changes that are due to your auto-formatter, please the next time make sure you don't commit such changes, they're difficult to revert.

The test file you have added will actually not run. All the test files are named "testthing.py" and not "test_thing.py", the former being what is picked up by pytest. I keep it anyway as ultimately I'd like to turn it into a script that would run during a doc build, I don't think it's necessary to run it part of the unit tests (and this single test takes 30 secs to run).

MarcSkovMadsen:
Thanks @maximlt 

The docs are a tough process. There is a lot I have to learn and a lot we have to align on. I'm not done yet at all.

But I am super motivated by the fact that I can get improvements into the next release and start to get feedback and see how it works out in practice. Thanks.

PR #838, opened 2022-08-18
Rework the landing page and add a Getting started guide

maximlt:
The landing page was basically a getting started guide which was a little difficult to extend to include `.interactive` and `explorer`. I've reworked it to (hopefully) make it look nicer and to include these two features. I've also added a Getting started guide that has 4 pages: Installation, `.hvplot`, `.interactive` and `explorer`. The page on `.hvplot` is basically the former landing page. The content on these pages could be improved (in a future iteration) but the scaffolding is now in place for this next step to be easier. Interestingly, `.hvplot()`, `.interactive()` and `explorer` are documented separately but I believe they could be combined in one story around data exploration. For this to happen some improvements to the Explorer would be required so that it supports more data structures.

I've added many screenshots and GIFs to support the landing page. I've tried to make them as light as possible, the total size is 3.9 MB. If it's too big, they will have to be hosted somewhere else. The advantage of having them in the repo is that the README could re-use them. Also I guess this would make it easier to set up docs versioning.

I'm going to build the docs based on this branch to see how this looks.


MarcSkovMadsen:
If you want lighter videos use .mp4. There are options to auto run them.

But test on mobile+laptop and different browsers to find an exact set of settings that work in all places.

MarcSkovMadsen:
Hi @maximlt 

Is it possible to see the rendered result somewhere? That would make the review easier.

MarcSkovMadsen (commented):
The diagram.svg file  could be simplified a bit one day. I think better alignment of objects and less arrows would make the figure easier to digest.

MarcSkovMadsen:
The networkx colors also need an overhaul one day. The red and green dont go very well together. Makes it look very old and outdated.

maximlt:
Yep you can find a link to the dev docs on the README of every page of the HoloViz projects. This is the second badge on the third line.

image

Agreeing with both of your suggestions.

MarcSkovMadsen:
Regarding the link to the "dev" docs. I guess they would be of the current master? I.e. not this branch before it was merged.

I.e. I could not use that link to help me review the PR @maximlt ?

maximlt:
Are you not clicking on the right link? When I access this URL I do see the landing page added by this PR: https://pyviz-dev.github.io/hvplot/

The docs are built and deployed in two ways:
- on release, i.e. as soon as a new tag is pushed to the repo
- thanks to Github workflow_dispatch events, that offer a small UI to parameterize and launch workflows. This is what I've used to build the docs from this PR (selection another branch than *master*)
image

Issue #823, opened 2022-08-07
WARNING: Calling the .opts method with options broken down by options group ... is deprecated


MarcSkovMadsen:
hvplot=0.8.0, holoviews=1.15.0

I'm trying to help a user here https://discourse.holoviz.org/t/interactive-and-dynamically-created-dict-of-widgets-to-xarray-sel/4000/5

When running the below code example I see the WARNING

```bash
WARNING:param.main: Calling the .opts method with options broken down by options group (i.e. separate plot, style and norm groups) is deprecated. Use the .options method converting to the simplified format instead or use hv.opts.apply_groups for backward compatibility.
2022-08-07 15:06:51,458 Calling the .opts method with options broken down by options group (i.e. separate plot, style and norm groups) is deprecated. Use the .options method converting to the simplified format instead or use hv.opts.apply_groups for backward compatibility.
```

```python
import collections
from pathlib import Path

import cartopy.crs as ccrs
import gdown
import hvplot
import hvplot.xarray
import numpy as np
import panel as pn
import panel.widgets as pnw
import xarray as xr
import pathlib

collections.Callable = collections.abc.Callable

CONFIG = {
    "2022070700_231_example.nc": "1vI5xH-kmA4OkoqJJz9keze-vN3cYCIf4",
    "2022070700_237_example.nc": "1RnWmfiW3U7VYS4Qoqo7_uFdeAIXAajye",
}

for file, file_id in CONFIG.items():
    if not pathlib.Path(file).exists():
        gdown.download(id=file_id, output=file, quiet=False)

file_list = [pathlib.Path(file) for file in CONFIG]

def open_dataset(file):
    key = f"file {file}"
    if not key in pn.state.cache:
        pn.state.cache[key]=xr.open_dataset(file)
    return pn.state.cache[key]

def get_sub_dataset(file, field):
    xds = open_dataset(file)
    return xds[field]

def get_options(file, field):
    options = {}
    xds_sub = get_sub_dataset(file, field)
    for (_, d) in enumerate(xds_sub.dims):
        for (_, c) in enumerate(xds_sub.coords):
            if d == c and (d != "lon" and d != "lat"):
                if xds_sub.coords[c].dtype == np.dtype("<M8[ns]"):
                    time_list = []
                    for _, el in np.ndenumerate(
                        xds_sub.coords["time"].values
                    ):
                        time_list.append(np.datetime_as_string(el, timezone="UTC"))
                    options[d] = time_list
                else:
[...]

PR #844, opened 2022-08-25
Fix HoloViews .opts warning by not calling GeoViews features

maximlt:
Fixes https://github.com/holoviz/hvplot/issues/823

GeoViews `Feature` objects are already instantiated in Geoviews. These objects are callable, this can be used to set up their options. hvPlot was calling them but not passing any options, making it a no-op. This PR was motivated by a HoloViews issue that led to a warning being emitted on a no-op `.opts` call (https://github.com/holoviz/holoviews/issues/5407)

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -1334,7 +1334,7 @@ class HoloViewsConverter:
     def _apply_layers(self, obj):
         if self.coastline:
             import geoviews as gv
-            coastline = gv.feature.coastline()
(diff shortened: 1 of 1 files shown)


philippjfr:
Hmm, sorry should have reviewed this earlier. This change is not safe I think. The reason why we kept the call in most cases is because it creates a copy of the feature, which means subsequent `.opts` calls won't modify the singleton feature. Instead of call we should probably use `.clone` here.

PR #850, opened 2022-08-25
Clone GeoViews' features

maximlt:
That's a safer approach to avoid modifying the singleton feature exposed by GeoViews.


philippjfr:
Looks good thanks!

PR #863, opened 2022-08-26
Update core docstrings

MarcSkovMadsen:
Improves the `core` docstrings


hvplot/plotting/core.py
```diff
@@ -45,31 +45,34 @@ class hvPlotBase:
...
-    ...     x="Year",
-    ...     y=["Art and Performance", "Business", "Biology", "Education", "Computer Science"],
-    ...     value_label="% of Degrees Earned by Women",
-    ...     legend="top",
-    ...     height=500,
-    ...     width=620,
-    ...     kind="line",
-    ... )
-    >>> line
+    .. code-block::
```
MarcSkovMadsen, line 48: using .. code-block:: will enable better sphinx documentation and it renders just as well in VS Code. I changed this systematically because the code is easier to copy and use without the `>>>`. Less friction.

maximlt:
@MarcSkovMadsen is this ready to be reviewed?

MarcSkovMadsen:
Yes. Thanks for asking. I think even more could be done, but it's like I can get it to 67% of perfect. The next contributor with a fresh perspective probably to 80%. And so on.

I really want the examples to be readable, copy-pastable and useful in the way that the data provided can easily to used to illustrate how the `kind` of plot works with datetime data instead of numeric data etc.

It also makes it much easier to have examples that quickly can be customized for example to address discourse questions or create reproducible examples for github.

maximlt:
Thanks a lot @MarcSkovMadsen for this, that's a massive amount of work that is going to change the experience of using hvPlot for a lot of users 👍 Despite my number of additional commits, each one of them were just making some very minor changes. This is going to be in a next release, hopefully tomorrow.

MarcSkovMadsen:
For its also a big step. It gets the docs to 51% percent to perfect. Then there is 49% to go. I need more energy. Someone else needs to do an iteration too. And then it needs to be updated based on user feedback.

PR #867, opened 2022-08-27
Disable the urls test

maximlt:
I thought they would not be picked up as the file name is `test_links.py` while the test files in hvPlot are formatted as `testthing.py`, I was wrong, and it appeared this test isn't very robust and fails from time to time.

### DIFF
--- a/hvplot/tests/test_links.py
+++ b/hvplot/tests/test_links.py
@@ -47,11 +47,11 @@ def _verify_urls(urls):
     return True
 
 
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
When its made robust maybe it does not need to run for every combination of OS and Python version?

maximlt:
I think it should only run part of the docs build. I wouldn't like volunteers to find out their tests CI runs fail because of these tests (which will happen sooner or later) as it would be totally unrelated to their changes.

Issue #862, opened 2022-08-26
Use https instead of http in links


MarcSkovMadsen:
If I search the hvplot code I see lots of `http://...` links. It is better practice to change to `https://` for security reasons.

![image](https://user-images.githubusercontent.com/42288570/186827035-742d48c8-e12c-4d3c-91d4-5a8a0723e81b.png)

jbednar:
Yes. PR gratefully accepted!

Issue #871, opened 2022-08-28
alpha option not found for kde plot with plotly


MarcSkovMadsen:
I've tried to run the Plotting_with_Plotly.ipynb notebook on the latest master branch.

I see the warning WARNING:param.main: alpha option not found for kde plot with plotly; similar options include: [] all over the place.

I believe this is confusing to (new) users and should removed. Either by just removing the `alpha` argument. Or adding support for `alpha`.

![image](https://user-images.githubusercontent.com/42288570/187060548-d1c13476-23cf-4084-8658-e6a047b2187c.png)

MarcSkovMadsen:
I also experience the warning when using the explorer with Plotly as the backend

![image](https://user-images.githubusercontent.com/42288570/187126421-335ccd00-1ac5-43fb-9c15-f72c3ef02a6b.png)

adivekar-utexas:
Same, I can also reproduce this.

Issue #870, opened 2022-08-28
Invalid url: http://www.ucrdatatool.gov/Search/Crime/State/StatebyState.cfm


MarcSkovMadsen:
The url http://www.ucrdatatool.gov/Search/Crime/State/StatebyState.cfm in the datasets.yaml file is not valid. Neither is the https:// version.

![image](https://user-images.githubusercontent.com/42288570/187059949-55bee23b-9f2e-4708-bc96-0e3ada2e8bfb.png)

I've tried to find a replacement but was not able to.

Hoxbro:
A quick search seems that the page has been moved to https://crime-data-explorer.app.cloud.gov.

The URL is only used for metadata and the data itself is located here: 'hvplot/examples/data/crime.csv'

Hoxbro:
I will close this issue and let the discussion continue in #873.

PR #873, opened 2022-08-28
Change to https

MarcSkovMadsen:
Fixes https://github.com/holoviz/hvplot/issues/862

Notes

- I found that changing to https in SVGs makes them invalid. So I did not do it.
- I found some invalid urls and fixed them
- There is one invalid url left that I cannot fix, which is https://github.com/holoviz/hvplot/issues/870
- I updated the `test_links.py` file to test all files and urls. But some domain refuse to respond to request requests so I exempted those.
- I manually opened and ran all the notebooks to verify they still worked. I discoved some unrelated issues
  - https://github.com/holoviz/holoviews/issues/5412
  - https://github.com/holoviz/hvplot/issues/871


MarcSkovMadsen:
The tests fail due to the obsolete url mentioned in https://github.com/holoviz/hvplot/issues/870. I can't find the url.

Let me know if there is more I should do. Thanks.

Hoxbro:
I will just add the invalid URL to `SKIP_URLS`. It is "only" metadata, and I feel it is more correct to have the site on which the data was downloaded even though it does not exist anymore.

Alternatively, you could update the URL to the new site I mentioned in issue #870, but I will at least verify that it is actually possible to get the data on the new site. 

So there are two ways forward: one is easy, and one is hard. I will let it be up to you which one to choose.

MarcSkovMadsen:
I would prefer to have a valid url from the new site. 

I tried to find it without luck. As I don't have the original context I would not be able to know if I found the right url. I don't even know how Intake works or what the purpose of the meta data is. And for now I don't need to learn.

Can someone else help find the correct url?

philippjfr:
Think this is probably the best link: https://ucr.fbi.gov/crime-in-the-u.s/2018/crime-in-the-u.s.-2019

philippjfr:
Alternatively there's also: https://crime-data-explorer.fr.cloud.gov/pages/explorer/crime/crime-trend

hvplot/tests/test_links.py
```diff
@@ -1,32 +1,70 @@
...
 PACKAGE_ROOT = ROOT.parent
 MAX_WORKERS = 10
+POST_FIXES = [".py", ".ipynb", ".md", ".yaml"]
+SKIP_URLS = [
+    "https://anaconda.org/anaconda/hvplot",
+    "https://anaconda.org/conda-forge/hvplot",
+    "https://anaconda.org/pyviz/hvplot",
+    "https://creativecommons.org/publicdomain/zero/1.0/",
+    "https://github.com/rasterio/rasterio",
+    "https://www.dask.org",
```
maximlt, line 29: Why are these URLs skipped?

Hoxbro:
I would change the test_links.py to the following. Even though it is nice with the fixtures, I think we should run all URLs in one go. This can both be good if there are duplicate URLs in different files and avoid overhead when starting up the ThreadPoolExecutor. If a URL fail it should be easy to find the file by searching for the URL.  

I changed also changed it to a simpler regex and using `requests.head` to only get the header of the site. Though [anaconda.org](https://anaconda.org) give status code 400 when using `head` and 200 when using `get`.
 

``` python
"""Urls in docstrings etc. should be valid and secure, i.e.

- exist, i.e. provide a 200 response
- use https:// instead of http:// unless
    - https:// is not supported by the web site
    - https:// cannot be used. For example in SVGs.
"""

import pathlib
import re
from concurrent.futures import ThreadPoolExecutor
from functools import partial

import requests

URL_REGEX = re.compile(r"(https?:\/\/?[\w/\-?=%.]+\.[\w/\-&?=%.]+)")
ROOT = pathlib.Path(__file__).parents[2]
POST_FIXES = ["*.py", "*.ipynb", "*.md", "*.yaml", "*.yml"]
SKIP_URLS = [
    "https://www.ucrdatatool.gov/Search/Crime/State/StatebyState.cfm",  # Not available anymore
    "http://weegee.vision.ucmerced.edu/datasets/landuse.html",  # Not available anymore
    "https://github.com/holoviz/hvplot",  # Alot of issues/PRs
    "http://octopart.com",  # Located in Notebook output, not relevant
    "https://jsperf.com",  # Located in Notebook output, not relevant
    "https://web.archive.org",  # Can be slow and does not work with the regex.
    "https://anaconda.org/anaconda/hvplot",  # Give a status code of 400?
    "https://anaconda.org/conda-forge/hvplot",  # Give a status code of 400?
    "https://anaconda.org/pyviz/hvplot",  # Give a status code of 400?
    "https://bugs.chromium.org/p/chromium/issues",  # Give status code 405 (needs login)
    "https://mybinder.org",  # Give status code 405
]
SKIP_FILES = [
    ".ipynb_checkpoints",  # Can give problems when running test local
]


def _get_files_to_check():
    for post_fix in POST_FIXES:
        for file in ROOT.rglob(post_fix):
            if any(skip in str(file) for skip in SKIP_FILES):
                continue
            yield file


def _find_urls(text):
    urls = set()
    for url in set(re.findall(URL_REGEX, text)):
        if url.endswith("."):
            # Regex will keep a dot at the end
            urls.add(url[:-1])
            continue
        if any(url.startswith(skip) for skip in SKIP_URLS):
            continue
        urls.add(url)

    return urls


def _verify_urls(urls):
    func = partial(requests.head, timeout=5)

    with ThreadPoolExecutor() as executor:
        futures = executor.map(func, urls)

    for url, response in zip(urls, futures):
        if not response.ok:
            raise ValueError(
                f"The url {url} responded with status {response.status_code}."
            )


def test_urls():
    urls = set()
    for file in _get_files_to_check():
        text = file.read_text()
        urls |= _find_urls(text)

    urls = sorted(urls)
    _verify_urls(urls)
```

maximlt:
I replaced the broken link by an internet archive link. I'd be happy if one day we could replace this crime dataset by another one, I find it a little creepy. Thanks @MarcSkovMadsen !!!

Issue #849, opened 2022-08-25
Missing A in docs


MarcSkovMadsen:
https://pyviz-dev.github.io/hvplot/

I believe it should be "A familiar and high-level ..."

![image](https://user-images.githubusercontent.com/42288570/186656423-b4e50c57-01ff-42b7-8a0b-85e79934179f.png)

jbednar:
Agreed! A PR would be very helpful, preferably combined with other docs fixes.

PR #874, opened 2022-08-28
Add missing A

MarcSkovMadsen:
Fixes https://github.com/holoviz/hvplot/issues/849


MarcSkovMadsen:
The test that fails times out. So I believe this is ready and I don't expect to do more here.

PR #903, opened 2022-09-06
Add bokeh tab

sophiamyang:
Hi, in the backend section, I thought we should also show bokeh since people might not read the document in order.

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -130,7 +130,7 @@ align: center
 
 ::::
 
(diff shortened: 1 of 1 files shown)


MarcSkovMadsen:
I have a PR with that too. Same for data backend. There should be a Pandas example.

Also because the home page is not great for communicating like the video you are creating, so it's important everything works also if things are not shown too to bottom.

philippjfr:
I'll merge this one and then rebase yours @MarcSkovMadsen.

PR #905, opened 2022-09-14
Add support for rioxarray rasterio and update tests

Hoxbro:
Fixes https://github.com/holoviz/hvplot/issues/929

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -639,8 +639,13 @@ class HoloViewsConverter:
     def _process_crs(self, data, crs):
         """Given crs as proj4 string, data.attr, or cartopy.crs return cartopy.crs
         """
-        # get the proj string: either the value of data.attrs[crs] or crs itself
-        _crs = getattr(data, 'attrs', {}).get(crs or 'crs', crs)
+        if hasattr(data, "spatial_ref"):
+            # if data is a rioxarray
+            _crs = data.spatial_ref.attrs.get('crs_wkt', crs)
+        else:
+            # get the proj string: either the value of data.attrs[crs] or crs itself
+            _crs = getattr(data, 'attrs', {}).get(crs or 'crs', crs)
+
         try:
             return process_crs(_crs)
         except ValueError:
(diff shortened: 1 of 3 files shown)


hvplot/converter.py
```diff
@@ -639,8 +639,13 @@ def _process_symmetric(self, symmetric, clim, check_symmetric_max):
     def _process_crs(self, data, crs):
         """Given crs as proj4 string, data.attr, or cartopy.crs return cartopy.crs
         """
-        # get the proj string: either the value of data.attrs[crs] or crs itself
-        _crs = getattr(data, 'attrs', {}).get(crs or 'crs', crs)
+        if hasattr(data, "spatial_ref"):
+            # if data is a rioxarray
+            _crs = data.spatial_ref.attrs.get('crs_wkt', crs)
```
ianthomas23, line 644:
```suggestion
        if hasattr(data, 'rio') and data.rio.crs is not None:
            # if data is a rioxarray with a valid CRS
            _crs = data.rio.crs.to_wkt()
```

ianthomas23:
I think the use of `data.rio.crs` is preferred as it checks a number of possible locations in the `xarray` for the CRS. It does however return an instance of `rasterio.crs.CRS` which we cannot pass to `cartopy` directly, so here I am converting to WKT first. This may not be ideal as it might give us various warnings, let's see what CI gives us.

An alternative to the `to_wkt()` is to add specific support in our CRS to cartopy for a `rasterio.crs.CRS`.

setup.py
```diff
@@ -83,6 +83,7 @@ def get_setup_version(reponame):
         'pooch',
         'scipy',
         'ipywidgets',
+        'rioxarray',
```
maximlt, line 86: @Hoxbro could you move `rioxarray` from the `tests` requirements to the `examples` requirements? Previously the tests you touched in this PR were skipped on Python 3.6 as the test suite only installs `-o tests` in this case, and these dependencies don't include any IO packages. I guess we could argue what dependencies should be part of the `tests` dependencies, but I guess for now we could just keep the status quo, at least to unblock this PR.

maximlt:
Great! Thanks a lot @Hoxbro !

PR #930, opened 2022-10-11
Ensure that xarray dataset metadata propagates to NdOverlays

### LINKED ISSUE #816: Coordinate label of a multi-element chart from an xarray Dataset lacks long name and units
<details>
<summary>Relevant software versions are collapsed here.</summary>

```
# Name                    Version                   Build  Channel
bokeh                     2.4.3            py39haa95532_0
holoviews                 1.15.0           py39haa95532_0
hvplot                    0.8.0            py39haa95532_0
python                    3.9.12               h6244533_0
xarray                    2022.6.0           pyhd8ed1ab_1    conda-forge
```

</details>

Given an `xarray.Dataset` that contains multiple data variables and a coordinate with "long_name" and/or "units" attributes, a multi-element plot created from that dataset labels the coordinate with only the coordinate's name instead of its long name and units:

``` python
In [1]: import holoviews, hvplot.xarray, xarray

In [2]: dset = xarray.Dataset(
   ...:     {"u": ("t", [1, 3]), "v": ("t", [4, 2])},
   ...:     coords={"t": ("t", [0, 1], {"long_name": "time", "units": "s"})},
   ...: )

In [3]: ndoverlay = dset.hvplot.line()

In [4]: ndoverlay.ddims[0]
Out[4]: Dimension('t')

In [5]: holoviews.render(ndoverlay).xaxis.axis_label
Out[5]: 't'
```

The same is true for an `NdLayout` returned by `dset.hvplot.line(subplots=True)`.

In contrast, a single-element plot created from one data variable has coordinates labeled with the long name and units:

``` python
In [6]: curve = dset["u"].hvplot.line()

In [7]: curve.kdims[0]
Out[7]: Dimension('t', label='time', unit='s')

In [8]: holoviews.render(curve).xaxis.axis_label
Out[8]: 'time (s)'
```

Similarly, a single-element bar plot `dset.hvplot.bar()` of both variables in the dataset honors the long name and units:

``` python
In [9]: dset.hvplot.bar().kdims[0]
Out[9]: Dimension('t', label='time', unit='s')
```

A possible cause is that, whereas the `HoloViewsConverter.single_chart` and other plot methods apply the long name and units stored in `self._redim` by calling the `redim(**self._redim)` method on the chart [[code](https://github.com/holoviz/hvplot/blob/04f6dae12c449987abf69601be47c1e292e1ab3a/hvplot/converter.py#L1504)], the `HoloViewsConverter.chart` method lacks such a call for the multi-element case [[code](https://github.com/holoviz/hvplot/blob/04f6dae12c449987abf69601be47c1e292e1ab3a/hvplot/converter.py#L1604-L1610)].  Could the call be included in the `HoloViewsConverter.chart` method?

To work around the difference, the user may re-dimension the returned overlay, although converting the attributes from the dataset is a bit verbose:

``` python
In [10]: labeled_ndoverlay = ndoverlay.redim(**{"t": {
    ...:     hname: dset["t"].attrs[xname]
    ...:     for hname, xname in [("label", "long_name"), ("unit", "units")]
    ...: }})

In [11]: labeled_ndoverlay.ddims[0]
Out[11]: Dimension('t', label='time', unit='s')
```

Hoxbro:
Fixes #816

With this change, the xarray dataset metadata propagates to `NdOverlay` figures. This will give more descriptive information like the xlabel in the example below, where it before only showed `t`. 

image

The added `.redim` could be placed on the line above or below, but I can´t think of a reason why one of them is better than the other.

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -1622,7 +1622,7 @@ class HoloViewsConverter:
         for c in y:
             kdims, vdims = self._get_dimensions([x], [c])
             chart = element(data, kdims, vdims).redim(**{c: self.value_label})
-            charts.append((c, chart.relabel(**self._relabel)))
+            charts.append((c, chart.relabel(**self._relabel).redim(**self._redim)))
         return (self._by_type(charts, self.group_label, sort=False)
                 .opts(cur_opts, backend='bokeh')
                 .opts(compat_opts, backend=self._backend_compat))
--- a/hvplot/tests/testcharts.py
+++ b/hvplot/tests/testcharts.py
@@ -78,7 +78,7 @@ class TestChart2D(ComparisonTestCase):
     def test_xarray_dataset_with_attrs(self):
         try:
             import xarray as xr
-            import hvplot.xarray
+            import hvplot.xarray  # noqa
         except ImportError:
             raise SkipTest('xarray not available')
 
--- a/hvplot/tests/testcharts.py
+++ b/hvplot/tests/testcharts.py
@@ -3,7 +3,7 @@ import numpy as np
 from unittest import SkipTest, expectedFailure
 from parameterized import parameterized
 
-from holoviews import NdOverlay, Store, dim
+from holoviews import NdOverlay, Store, dim, render
 from holoviews.element import Curve, Area, Scatter, Points, Path, HeatMap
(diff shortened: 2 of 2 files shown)


philippjfr:
Looks good!

PR #944, opened 2022-10-14
add intended rolling_window param into pipeline

droumis:
The ROLLING_WINDOW param was not actually being used in the pipeline, although that was the intention. 

I also changed it to default '1D' to make it consistent with the later visualization. Also because what it was before ('1W') throws an error `ValueError: <Week: weekday=6> is a non-fixed frequency`

### DIFF
--- a/examples/getting_started/interactive.ipynb
+++ b/examples/getting_started/interactive.ipynb
@@ -61,7 +61,7 @@
    "outputs": [],
    "source": [
     "LATITUDE = 30.\n",
-    "ROLLING_WINDOW = '1W'\n",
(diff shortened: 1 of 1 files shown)


droumis:
hmmm I see that some metadata about my notebook session made it into the commit.. What is the best practice here for scrubbing that info?

droumis:
the notebook cleaner is described [here](https://github.com/holoviz/holoviz/blob/master/doc/contributing.md#:~:text=Jupyter%20Notebooks%20(.ipynb,handled%20in%20notebooks.)

I'll do that now

droumis:
ready to merge now

PR #967, opened 2022-11-16
Add pre-commit

Hoxbro:
Adding pre-commit. The pre-commit itself is inspired by Holoviews, though I tried adding `ruff` instead of flake8, this can of course be reversed if needed. 

I can also add `isort` if needed.


hvplot/interactive.py
```diff
@@ -1,64 +1,3 @@
...
-                    "Holoviews 1.15.1 or greater installed."
-                )
-            nested_op = {"args": [op_arg.start, op_arg.stop, op_arg.step], "kwargs": {}}
-            for widget in _find_widgets(nested_op):
-                if widget not in widgets:
-                    widgets.append(widget)
-    return widgets
-
-
 """
```
maximlt, line 1: This comment in a docstring was not meant to become the module docstring :/

maximlt (changes requested):
Please let's not use `ruff` in hvPlot. It's:

- too new
- more useful for large code bases
- an incomplete re-implementation of flake8 (why not making flake8 faster instead?)

`doit test_flakes` should also be updated.

Issue #970, opened 2022-11-17
hvplot incompatible with bokeh 3.0.1, 3.0.2


Timothy-W-Hilton:
Thanks for all your excellent work with holoviz.

Seems there's an incompatibility between hvplot and bokeh 3.0.  The below shows results for bokeh 3.0.2; bokeh 3.0.1 gives me the same result. bokeh 2.4.3 works fine.

#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc)

hvplot                    0.8.1
bokeh                     3.0.2
python                    3.10.6
Ubuntu 20.04.5 LTS
conda 4.1.2, condaforge channel (both via miniforge)

#### Description of expected behavior and the observed behavior

installing hvplot in a minimal conda environment results in an ImportError on `import hvplot`:
> ImportError: cannot import name 'Box' from 'bokeh.models' (/home/timh/mambaforge/envs/venv-test/lib/python3.10/site-packages/bokeh/models/__init__.py)

#### Complete, minimal, self-contained example code that reproduces the issue

```
mamba create -n venv-test python=3.10 bokeh hvplot
conda activate venv-test
python
>>> import hvplot
```

If I specify bokeh v2.4.3 it works fine:

```
mamba create -n venv-test python=3.10 bokeh=2.4.3 hvplot
conda activate venv-test
python
>>> import hvplot
```

#### Stack traceback and/or browser JavaScript console output
>>> import hvplot
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/timh/mambaforge/envs/venv-test/lib/python3.10/site-packages/hvplot/__init__.py", line 65, in <module>
    import panel as _pn
  File "/home/timh/mambaforge/envs/venv-test/lib/python3.10/site-packages/panel/__init__.py", line 3, in <module>
    from . import layout # noqa
  File "/home/timh/mambaforge/envs/venv-test/lib/python3.10/site-packages/panel/layout/__init__.py", line 1, in <module>
    from .accordion import Accordion # noqa
  File "/home/timh/mambaforge/envs/venv-test/lib/python3.10/site-packages/panel/layout/accordion.py", line 5, in <module>
    from .base import NamedListPanel
  File "/home/timh/mambaforge/envs/venv-test/lib/python3.10/site-packages/panel/layout/base.py", line 13, in <module>
    from ..io.model import hold
[...]

PR #974, opened 2022-11-21
Add panel pin to ensure Bokeh <3

maximlt:
FIxes https://github.com/holoviz/hvplot/issues/970

With the release of Bokeh 3 users started to get Bokeh 3 when installing hvPlot, while it's not compatible with it.

This PR adds Panel as a runtime dependency, as it is actually required by the explorer. It also pins it to `>=0.11.0` as this version of Panel is the first one that added an upper pin to Bokeh.

Alternatively we could have pinned HoloViews to `>= 1.15.0` as it pinned `"panel >=0.13.1"`, however it's just been released 6 months ago. Previous releases of HoloViews pinned `"panel >=0.9.5"` which didn't have the upper pin for Bokeh.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,6 +34,7 @@ install_requires = [
     'pandas',
     'numpy>=1.15',
     'packaging',
+    'panel >=0.11.0',
 ]
 
 _examples = [


maximlt:
Pinning Panel instead of pinning more HoloViews allows us to preserve support of Python 3.6, at least for the time being.

PR #980, opened 2022-11-24
Add 0.8.2 release notes

maximlt:


jlstevens:
Just wondering, why are you deleting the CHANGELOG?

doc/releases.md
```diff
@@ -283,3 +335,15 @@ This release includes a number of major improvements to the documentation and co
...
+
+## Version 0.2.1
+
+**July 8, 2018**
+
+## Version 0.2.0
+
+**July 7, 2018**
+
+## Version 0.1.1
```
jlstevens, line 347: I think this will look wrong when built. I propose one of three options:

1. Fill out some entries for these old releases
2. Comment out these entries for filling in later (if we think we will do that)
3. Collapse them into one sentence about the earliest releases, keeping only the dates.

doc/releases.md
```diff
@@ -1,44 +1,76 @@
...
 - Interactive:
-  - Set interactive function to lazy to avoid calling it twice ([772](https://github.com/holoviz/hvplot/pull/772))
-  - Add support for hvplot kind accessor ([781](https://github.com/holoviz/hvplot/pull/781), [858](https://github.com/holoviz/hvplot/pull/858))
-  - Add more tests to `Interactive` and some comments ([812](https://github.com/holoviz/hvplot/pull/812))
-  - Fixes to the operators implemented by Interactive ([814](https://github.com/holoviz/hvplot/pull/814))
-  - Propagate `max_rows` ([815](https://github.com/holoviz/hvplot/pull/815))
-  - Only register the function parameters watchers once ([818](https://github.com/holoviz/hvplot/pull/818))
-  - Do not re-execute transform on copied clones ([833](https://github.com/holoviz/hvplot/pull/833))
-  - Internal callback made dependent only when parameters are found ([857](https://github.com/holoviz/hvplot/pull/857))
+  - Set interactive function to lazy to avoid calling it twice ([#772](https://github.com/holoviz/hvplot/pull/772))
```
jlstevens, line 42: Took me a while to see the changes here! (Missing #s)

doc/releases.md
```diff
@@ -1,44 +1,76 @@
 # Releases
 
+See [the HoloViz blog](https://blog.holoviz.org/tag/hvplot.html) for a visual summary of the major features added in a selection of releases.
+
+## Version 0.8.2
+
+**November 24, 2022**
+
+The 0.8.2 release is a bug fix release. It improves compatibility with HoloViews 1.15.2 that emits now a warning when parsing DataFrame with non-string columns. The HoloViz tools are not yet compatible with the recent Bokeh 3, this hvPlot release pins Panel in a way that hvPlot users getting this version will no longer see Bokeh 3 being installed.  The documentation has seen some improvements too; @MarcSkovMadsen has kept improving the docstrings and we congratulate @sophiamyang for her first contribution that made the landing page nicer! Many thanks to @droumis, @Hoxbro, @maximlt, @philippjfr and @MarcSkovMadsen for contributing!
```
jlstevens, line 9: I would make this sentence into the top entry in the notes (there must be a PR with this pin, right?):

> The HoloViz tools are not yet compatible with the recent Bokeh 3, this hvPlot release pins Panel in a way that hvPlot users getting this version will no longer see Bokeh 3 being installed.

Maybe something like:

```
- Dependencies
   - Added Panel pin to ensure Bokeh <3 until HoloViz ecosystem updated to support Bokeh 3
      < PR link>   
```

maximlt:
> Just wondering, why are you deleting the CHANGELOG?

To only have one place that has the changelog. For example, I had forgotten to update CHANGELOG.md in the previous release. It's also because hvPlot doesn't distribute its CHANGELOG.md, it's not in the manifest file.

doc/releases.md
maximlt: Ok I'll opt for 3.

doc/releases.md
maximlt: Yes that's the packaging entry:

```
- Packaging:
  - Add Panel lower pin to ensure Bokeh 3 is not installed ([#974](https://github.com/holoviz/hvplot/pull/974))
```

I'll move it up.

jlstevens:
>  It's also because hvPlot doesn't distribute its CHANGELOG.md, it's not in the manifest file.

I would argue I don't expect users to see the CHANGELOG and that it is for developers viewing/cloning the github repo.

I suppose it makes more sense to remove here as hvplot is already using `.md` for the docs unlike holoviews which still uses `.rst`. Why not keep the CHANGELOG but with a single line?

> For the release notes, please consult /doc/releases.md

Then there is still a CHANGELOG but you never need to update it.

doc/releases.md
jlstevens: Ok, then I would move this entry to the top and remove the sentence in the overall description.

doc/releases.md
```diff
@@ -1,44 +1,76 @@
...
+  - New and updated docstrings ([#892](https://github.com/holoviz/hvplot/pull/892), [#863](https://github.com/holoviz/hvplot/pull/863))
+  - Improve the landing page ([#874](https://github.com/holoviz/hvplot/pull/874), [#903](https://github.com/holoviz/hvplot/pull/903), [#876](https://github.com/holoviz/hvplot/pull/876))
+  - Change *http* to *https* ([#873](https://github.com/holoviz/hvplot/pull/873))
+  - Add intended `rolling_window` param into pipeline ([#944](https://github.com/holoviz/hvplot/pull/944))
+  - Updates the README ([#817](https://github.com/holoviz/hvplot/pull/817))
+  - Miscellaneous documentation improvements ([#866](https://github.com/holoviz/hvplot/pull/866), [#887](https://github.com/holoviz/hvplot/pull/887), [#910](https://github.com/holoviz/hvplot/pull/910))
+- Packaging:
+  - Add Panel lower pin to ensure Bokeh 3 is not installed ([#974](https://github.com/holoviz/hvplot/pull/974))
+- Development:
+  - Add pre-commit ([#967](https://github.com/holoviz/hvplot/pull/967))
```
jlstevens, line 31: Add 'pre-commit' what exactly?

doc/releases.md
maximlt: I think we should keep the sentence there as I think users can still end up with an environment that first has Bokeh 3 and then pulls an old version of hvPlot/HoloViews/Panel ... Holoviz not yet supporting Bokeh 3 is at least disclaimed more a bit more explicitly with this sentence.

jlstevens: In that case, the sentence should be shorter and clearer. Maybe simply:

> Please note that the HoloViz ecosystem is not yet compatible with Bokeh 3.

maximlt:
Thanks Jean-Luc, I applied your suggestions.

I think distributing the CHANGELOG is kind of nice for repackagers if they want to have a look at it without having to find the package website. But contrary to e.g. HoloViews, hvPlot wasn't distributing it so it's OK replacing it with the small sentence you suggested.

I think the improvement would be to have the CHANGELOG in the root dir of the repo, the doc building pulling it from there so that there's only one source. It must be possible with sphinx now, I've seen it done in some places. That's for another PR though!

PR #1002, opened 2022-12-03
improve the hist reference guide

### LINKED ISSUE #998: Make it easier to plot distributions of categorical data
Trying to contribute to pythonplot.com during the PyData Global 2022 I realized that it might be difficult for users to figure out how to plot a distribution of a categorical variable.

hvPlot (and Pandas) follows the definition of *histogram*: A histogram is an approximate representation of the [distribution](https://en.wikipedia.org/wiki/Frequency_distribution) of **numerical** data.

image

Plotly on the other hand does not care about that. They just make it easy to plot for their users 

image

We should also make this easier for our users. We could

- Point out in the histogram reference guide that a histogram is only for numerical data.
- Provide an example of a distribution plot for categorical variable
- Document this in docstrings
- Improve the error message if the user provides a categorical variable. It is really, really hard to understand the error message right now.

Or maybe just support categorical parameters in `.hvplot.hist`.

MarcSkovMadsen:
Closes #998. 

Please note https://github.com/holoviz/holoviews/pull/5540 closes the issue about the non-understandable error message when trying to `hist` categorical data.

In addition, I've added more examples to illustrate what is possible.

Comparing to the plotly documentation there is still a long, long way to go https://plotly.com/python/histograms/.

### DIFF
--- a/examples/reference/pandas/hist.ipynb
+++ b/examples/reference/pandas/hist.ipynb
@@ -88,10 +88,11 @@
    ]
   },
   {
+   "attachments": {},
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "If you want to plot the distribution of a categorical column you can calculate the distribution using `value_counts` and plot it using `.hvplot.bar`."
+    "If you want to plot the distribution of a categorical column you can calculate the distribution using Pandas' method `value_counts` and plot it using `.hvplot.bar`."
    ]
   },
   {
--- a/examples/reference/pandas/hist.ipynb
+++ b/examples/reference/pandas/hist.ipynb
@@ -70,7 +70,7 @@
     "import pandas as pd\n",
     "from bokeh.sampledata.commits import data as commits\n",
     "\n",
-    "commits=commits.reset_index().sort_values(\"datetime\")\n",
+    "commits = commits.reset_index().sort_values(\"datetime\")\n",
     "commits.head(3)"
    ]
   },
--- a/examples/reference/pandas/hist.ipynb
+++ b/examples/reference/pandas/hist.ipynb
@@ -105,22 +105,9 @@
   }
  ],
  "metadata": {
-  "kernelspec": {
-   "display_name": "Python 3 (ipykernel)",
-   "language": "python",
-   "name": "python3"
-  },
   "language_info": {
-   "codemirror_mode": {
-    "name": "ipython",
-    "version": 3
-   },
-   "file_extension": ".py",
-   "mimetype": "text/x-python",
    "name": "python",
-   "nbconvert_exporter": "python",
-   "pygments_lexer": "ipython3",
-   "version": "3.9.7"
+   "pygments_lexer": "ipython3"
   }
  },
  "nbformat": 4,
--- a/examples/reference/pandas/hist.ipynb
+++ b/examples/reference/pandas/hist.ipynb
@@ -6,14 +6,16 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import hvplot.pandas  # noqa"
+    "import hvplot.pandas  # noqa\n",
+    "\n",
+    "# hvplot.extension(\"matplotlib\")"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "`hist` is often a good way to start looking at data to get a sense of the distribution. Similar methods include [`kde`](kde.ipny) (also available as `density`)."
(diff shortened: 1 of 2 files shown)


examples/reference/pandas/hist.ipynb
```diff
@@ -6,14 +6,16 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import hvplot.pandas  # noqa"
+    "import hvplot.pandas  # noqa\n",
+    "\n",
+    "# hvplot.extension(\"matplotlib\")"
```
maximlt, line 11: Why do you use matplotlib?

hvplot/plotting/core.py
```diff
@@ -1295,6 +1296,20 @@ def hist(self, y=None, by=None, **kwds):
...
+        .. code-block::
+
+            import hvplot.pandas
+            import pandas as pd
+
+            data = pd.DataFrame({
+                "library": ["bokeh", "plotly", "matplotlib", "bokeh", "matplotlib", "matplotlib"]
+            })
+
+            data["library"].value_counts().hvplot.bar(invert=True, flip_yaxis=True)
```
maximlt, line 1311: It seems using `invert` and `flip_yaxis` is confusing here, they seem to be required while as far as I understand they're just there for stylistic reasons.

examples/reference/pandas/hist.ipynb
```diff
@@ -40,16 +51,78 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas as pd\n",
+    "from bokeh.sampledata.commits import data as commits\n",
+    "\n",
+    "commits=commits.reset_index().sort_values(\"datetime\")\n",
```
maximlt, line 73:
```suggestion
    "commits = commits.reset_index().sort_values(\"datetime\")\n",
```

examples/reference/pandas/hist.ipynb
MarcSkovMadsen: I don't use it. But I indicate how a user could change the backend.

hvplot/plotting/core.py
MarcSkovMadsen: What do you mean by *seams confusing* ?

For context I put them in because otherwise you will not get readable and useful plots.

examples/reference/pandas/hist.ipynb
MarcSkovMadsen: Its done to help the user. If you are not a super user of hvplot you 1. might not know that its possible or 2. remember how to do this.

The intent is to help the user and enable her to quickly use the notebook to explore and try out different options including changing the plotting backend.

maximlt: Yes but it's also a good idea to explain things in relevant sections. Following your argument, we could document anything anywhere, it seems to me this is a recipe for spaghetti documentation.

There are already explanations on how to get Matplotlib plots:
- on the home page
- in the getting started guide
- in a dedicated guide to plotting with Matplotlib
- in a guide that explains how to switch between plotting backends

It seems to me that it's enough.

At some point I'd like the reference gallery to also show Matplotlib and Plotly plots. This is what HoloViews does, except that to do that it has to duplicate the notebooks which makes it pretty annoying to make any sort of modification, since you have to modify three notebooks. I also makes it too easy for things to diverge. I wouldn't like to have this in hvPlot now, I prefer improving the default Bokeh reference pages first.

MarcSkovMadsen: I'f you where not tied to notebooks, you could develop a live app that could enable the user to select the data backed (pandas, xarray, duckdb etc), the plot, the plot backend.

This would also give you a lot of automated tests to help make hvplot and holoviews more robust.

hvplot/plotting/core.py
maximlt: This is the plot I get without `invert=True, flip_yaxis=True`, which doesn't look that bad and is closer to what I expect from a *distribution* plot, which is the intent of this added example.

image

> I also think its really nice to demonstrate a lot of the parameters

Totally agree with that, but not in the docstrings, as there are a very large number of examples we could put in there, and without any visual reference they are not going to help "potential buyers" much.

```suggestion
            data["library"].value_counts().hvplot.bar()
```

maximlt:
Thanks @MarcSkovMadsen !

PR #1005, opened 2022-12-07
change GeoViews to geoviews (import name)

raybellwaves:


maximlt:
Thanks! Why did you remove `pyproj`?

raybellwaves:
> Thanks! Why did you remove `pyproj`?

I put it in the commit message but show have explained here: it's installed with cartopy

maximlt:
Indeed I missed that. I actually added pyproj recently, the fact that cartopy depends on pyproj is an implementation detail. Sure this is not meant to change any time soon, but well, I prefer to be explicit in these cases.

If that's fine for you I'll add it back?

raybellwaves:
np. just added it back in

maximlt:
Great thanks!

PR #1015, opened 2023-01-18
Python 3.11 support

### LINKED ISSUE #1013: Support Python 3.11
Python 3.11 provides major performance improvement that I would like to take advantage of when using the HoloViz ecosystem.

Panel and HoloViews already support python 3.11 according to Pypi.

Please support python 3.11. Thanks.

Hoxbro:
Resolves #1013

Updated the test suite to not fail and added Python 3.11 to it. Also added python 3.11 as a classifier. 

No code change has been done.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -52,61 +52,71 @@ jobs:
     timeout-minutes: 90
     defaults:
       run:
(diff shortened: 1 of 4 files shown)


maximlt:
Thanks!

PR #1026, opened 2023-03-03
Let ``process_crs`` try a few approaches before raising error

jsignell:
This PR makes the `process_crs` more generous in what it accepts and adds tests for the function.

### DIFF
--- a/hvplot/tests/testutil.py
+++ b/hvplot/tests/testutil.py
@@ -295,7 +295,7 @@ def test_process_crs(input):
 
 def test_process_crs_raises_error():
     pytest.importorskip("pyproj")
-    ccrs = pytest.importorskip("cartopy.crs")
+    pytest.importorskip("cartopy.crs")
     with pytest.raises(ValueError, match="must be defined as a EPSG code, proj4 string"):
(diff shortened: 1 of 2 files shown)


jsignell:
I'm not 100% sure why the python 3.7 jobs are failing, but I would expect this new test to fail on main in all cases, so I think it would be reasonable to skip the new test for python 3.7 and call it a day. It is more likely tied to the pyproj version though. It looks like pyproj dropped support for python 3.7 a while back so that is probably related.

Hoxbro:
I'm fine with skipping the failing test on Python 3.7.

maximlt:
Thanks @jsignell ! Looks good to me.

Note for myself, `crs` accepts:
- `'EPSG:<epsg_code>'` as `str`
- `'epsg:<epsg_code>'` as `str`
- `<epsg_code>` as `int`
- PROJ.4 string
- `pyproj.Proj`
- `cartopy.crs.CRS`
- and probably more :)

Also the utils that parse the projection like `proj_to_cartopy` and `check_crs` are from *salem*, need to check if they've been updated.

PR #1029, opened 2023-03-06
Ensure interactive pipeline is not re-executed needlessly

philippjfr:
- Fixes `Interactive.eval()` ensuring it returns the current value (previously it would only ever return the initial value)
- Ensures that unless parameters change we don't re-evaluate the pipeline each time the value is requested by a callback.

### DIFF
--- a/hvplot/interactive.py
+++ b/hvplot/interactive.py
@@ -98,7 +98,7 @@ def _find_widgets(op):
 # consist of multiple chains, such as `dfi[dfi.A > 1]`, as the `Interactive`
 # instance is referenced twice in the expression. As a consequence `_depth`
 # is not the total count of `Interactive` instance creations of a pipeline,
-# it is the count of instances created in outer chain. In the example, that
+# it is the count of instances created in the outer chain. In the example, that
 # would be `dfi[]`. `Interactive` instances don't have references about
 # the instances that created them or that they create, they just know their
 # current location in a chain thanks to `_depth`. However, as some parameters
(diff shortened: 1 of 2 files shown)


maximlt (commented):
Just left a comment in the code. Besides that I've tested the PR locally on a few examples and things seemed to work fine. So it's all good to me, merge it when you make the suggested change, if you think it makes sense.

> Fixes Interactive.eval() ensuring it returns the current value (previously it would only ever return the initial value)

Indeed `eval` was broken, don't know we missed that until now!

> Ensures that unless parameters change we don't re-evaluate the pipeline each time the value is requested by a callback.

That's the part I struggle to understand. Could you please explain what was going on before, with an example maybe?

One last comment, `_setup_invalidation` sets up the `_invalidate_current` callback on every Parameter of the pipeline, and it does so everytime `Interactive.__init__` runs. So `_invalidate_current` in registered many times, that number grows quickly with the complexity and length of the pipeline. It is a pretty cheap function so I don't think it has any performance penalty. I just thought it was worth noting it as it might come as a bit unexpected and, the way interactive pipelines are built, iteratively for sure, it's likely users are going to register a lot of "orphan" callbacks, hopefully without any bad effects.

philippjfr:
>That's the part I struggle to understand. Could you please explain what was going on before, with an example maybe?

To be fully clear I'll preface this with the fact that the caching is still all-or-nothing, i.e. we do not in fact cache intermediate steps in a chain and then only evaluate the parts of the pipeline that have to be re-evaluated. Overall one might argue that the caching of intermediates may be too expensive anyway particularly for very long pipelines on large datasets. This is somewhat different from HoloViews where chaining generally results in a small number of steps and most operations are intended to reduce the size of the dataset.

Basically what happened before is that interactive held the original object and the series of transforms that have to be applied to arrive at the final object. Whenever the function returned by `_callback` was invoked it would then re-evaluate the whole pipeline (while eval just used the `_current` value). Since we never updated `_current` the `eval` method didn't work and because `_callback` would simply always re-evaluate the pipeline everything that required the output of the function would have to re-run the entire pipeline to get the current value.

What we do now is to internally invalidate the `_current` state so that when it is next required by `.eval()` or by a `_callback` we re-compute it and use that cached value until it is again invalidated.

philippjfr:
I've now cleaned up the invalidation code and updated the docstrings to explain the approach.

philippjfr:
Python 3.6 tests seem to be failing due to the solver consistently. Will ignore and merge.

PR #1031, opened 2023-03-08
Only use long_name if it is a string

Hoxbro:
Fixes the hvplot part of https://github.com/holoviz/holoviews/issues/5615

This change makes it so that it only uses the long name if it is a string.


maximlt:
LGTM, thanks!

PR #1032, opened 2023-03-11
Improve bar documentation especially for color

MarcSkovMadsen:
Addresses a users comment about limited `bar` documentation and no documentation of `color`. See https://discourse.holoviz.org/t/is-there-a-way-to-add-a-specific-color-to-each-group-in-a-grouped-bar-plot-like-hue-in-seaborns-barplot/5066

Also fixes a few docstring bugs identified when rendering the output of `hvplot.help('bar')` on Discourse.

https://user-images.githubusercontent.com/42288570/224508335-3a1d31b4-64ee-4943-a0ee-591886aa3529.mp4

## Issues identified while working on this

The main issue is that lots of things does not work for Matplotlib and Plotly backends as they do for Bokeh backend

- #1034 
- #1035 
- #1036 
- #1037 
- #1038 
- #1039

### DIFF
--- a/examples/reference/pandas/bar.ipynb
+++ b/examples/reference/pandas/bar.ipynb
@@ -18,9 +18,7 @@
     "\n",
     "A `bar` plot represents **categorical data** with rectangular bars with heights proportional to the **numerical values** that they represent.\n",
     "The x-axis represents the categories and the y axis represents the numerical value scale.\n",
-    "The bars are of equal width which allows for instant comparison of data.\n",
-    "\n",
-    "Write `hvplot.extension('bokeh')`, `hvplot.extension('matplotlib')` or `hvplot.extension('plotly')` to change the plotting backend"
+    "The bars are of equal width which allows for instant comparison of data."
    ]
   },
   {
@@ -32,7 +30,7 @@
     "pd.DataFrame({\n",
     "    \"framework\": [\"hvPlot\", \"HoloViews\", \"Panel\"], \n",
     "    \"stars\": [700, 2400, 2600]\n",
-    "}).hvplot.bar(x=\"framework\", y=\"stars\", color=\"gold\", title=\"Bar Plot of Github Stars as of March 2023\", ylabel=\"⭐\")"
+    "}).hvplot.bar(x=\"framework\", y=\"stars\", color=\"gold\", title=\"Bar Plot of Github Stars\", ylabel=\"⭐\")"
    ]
   },
   {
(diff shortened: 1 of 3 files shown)


examples/reference/pandas/bar.ipynb
```diff
@@ -18,9 +18,7 @@
     "\n",
     "A `bar` plot represents **categorical data** with rectangular bars with heights proportional to the **numerical values** that they represent.\n",
     "The x-axis represents the categories and the y axis represents the numerical value scale.\n",
-    "The bars are of equal width which allows for instant comparison of data.\n",
-    "\n",
-    "Write `hvplot.extension('bokeh')`, `hvplot.extension('matplotlib')` or `hvplot.extension('plotly')` to change the plotting backend"
+    "The bars are of equal width which allows for instant comparison of data."
```
maximlt, line 21: Removed the mention to the other extensions. If we do it, it's something that should happen all at once to make sure we do it consistently.

examples/reference/pandas/bar.ipynb
```diff
@@ -32,7 +30,7 @@
     "pd.DataFrame({\n",
     "    \"framework\": [\"hvPlot\", \"HoloViews\", \"Panel\"], \n",
     "    \"stars\": [700, 2400, 2600]\n",
-    "}).hvplot.bar(x=\"framework\", y=\"stars\", color=\"gold\", title=\"Bar Plot of Github Stars as of March 2023\", ylabel=\"⭐\")"
+    "}).hvplot.bar(x=\"framework\", y=\"stars\", color=\"gold\", title=\"Bar Plot of Github Stars\", ylabel=\"⭐\")"
```
maximlt, line 33: Removed the date, just to avoid having to update it in the future.

examples/reference/pandas/bar.ipynb
```diff
@@ -55,13 +53,6 @@
     "autompg.head()"
    ]
   },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "Note that the columns `yr`, `origin`, `name` and `mfr` are categorical.Note that the columns `yr`, `origin`, `name` and `mfr` are categorical."
```
maximlt: Duplicate sentences, and almost all the data in this dataset is actually categorical, I think

maximlt:
Thanks a lot!

examples/reference/pandas/bar.ipynb
MarcSkovMadsen: Its ok with me. I actually put it in though such that people Boes this is older. Now i would expect the data to be up to date 😂

maximlt: Yep it's difficult to have both ^^

PR #718, opened 2022-03-23
Use nbval to execute notebooks and precommit with nbqa for linting notebooks

maximlt:
For some reason the CI would not even start running on https://github.com/holoviz/hvplot/pull/705.

PR #1044, opened 2023-03-15
Run examples with nbval

maximlt:
Let's see if we're more lucky than a year ago with https://github.com/holoviz/hvplot/pull/718


maximlt:
Wo it seems like this time it actually worked :) Merging immediately!

Issue #1051, opened 2023-03-28
When hvplot checks `is_intake`, remove pathname requirement


kthyng:
#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc)

```
hvplot                    0.8.3                      py_0    pyviz
intake                    0.6.6                    pypi_0    pypi
bokeh                     2.4.3              pyhd8ed1ab_3    conda-forge
```

#### Description of expected behavior and the observed behavior

I'm making a plot with intake through an intake catalog using my own `DataFrameTransform`. When `hvplot` checks the incoming `data` in `_process_data()` in converter.py, it uses `utils.py` `is_intake()` which runs
```
    if not check_library(data, 'intake'):
        return False
```
and an intake Transform that is not a fully-qualified path in the intake namespace (as mine isn't because it is just defined in my project) won't pass this test. But, I don't think it should need to since it is subsequently checked with `isinstance(data, DataSource)`. See discussion on this https://github.com/intake/intake/issues/724.

#### Complete, minimal, self-contained example code that reproduces the issue

Save to a local file "test_transform.py":
```
from intake.source.derived import GenericTransform

class DataFrameTransform(GenericTransform):
    """Transform where the input and output are both Dask-compatible dataframes

    This derives from GenericTransform, and you mus supply ``transform`` and
    any ``transform_kwargs``.
    """

    input_container = "dataframe"
    container = "dataframe"
    optional_params = {}
    _df = None

    def to_dask(self):
        if self._df is None:
            self._pick()
            self._df = self._transform(self._source.to_dask().compute(),
                                       **self._params["transform_kwargs"])
        return self._df

    def _get_schema(self):
        """load metadata only if needed"""
        self.to_dask()
        return Schema(dtype=self._df.dtypes,
                      shape=(None, len(self._df.columns)),
                      npartitions=self._df.npartitions,
                      metadata=self.metadata)

    def read(self):
        # return self.read()
        return self.to_dask()


def DF(df):
    return df
```

```
import pandas as pd
import intake

# make data
d = {"a": [0,0.1], "b": [1,1.1]}
pd.DataFrame(d).to_csv("test.csv")

# make catalog
cat = \
'''
name: ctd
description: 

sources:
  ctd_source_base:
    description: CTD Source base
    driver: csv
    args:
      urlpath: test.csv

  ctd_source:
    description: CTD Source
    driver: test_transform.DataFrameTransform
    args:
[...]

PR #1052, opened 2023-03-28
tiny fix so hvplot can be used with intake

kthyng:
...when an intake Transform is local instead of formally installed

PR for https://github.com/holoviz/hvplot/issues/1051

### DIFF
--- a/hvplot/util.py
+++ b/hvplot/util.py
@@ -320,7 +320,10 @@ def is_dask(data):
     return isinstance(data, (dd.DataFrame, dd.Series))
 
 def is_intake(data):
-    return any(check_library(cls, 'intake') for cls in type(data).mro())
+    if "intake" not in sys.modules:
+        return False
+    from intake.source.base import DataSource
+    return isinstance(data, DataSource)
 
 def is_ibis(data):
     if not check_library(data, 'ibis'):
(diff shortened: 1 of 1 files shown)


hvplot/util.py
```diff
@@ -320,10 +320,7 @@ def is_dask(data):
     return isinstance(data, (dd.DataFrame, dd.Series))
 
 def is_intake(data):
-    if not check_library(data, 'intake'):
-        return False
-    from intake.source.base import DataSource
-    return isinstance(data, DataSource)
+    return any(check_library(cls, 'intake') for cls in type(data).mro())
```
Hoxbro, line 323:
```suggestion
    if "intake" not in sys.modules:
        return False
    from intake.source.base import DataSource
    return isinstance(data, DataSource)
```

I think we need to have the DataSource check, for an edge cases like this will return True:
image

martindurant: You could still have both, then - replace the current library check with the modified one, and then import as before.

martindurant: No, it is currently failing for subclasses of DataSource which are defined outside of the intake package.

kthyng:
Is that one test suite randomly failing?

philippjfr:
Sorry, just getting back from vacation. Will merge when tests pass.

Hoxbro:
The failing test is unrelated, so I will merge.

PR #1054, opened 2023-04-05
separate import tests to see what's missing

raybellwaves:
This allows to see which package(s) out of cartopy, geoview and pyproj are missing.

### DIFF
--- a/hvplot/util.py
+++ b/hvplot/util.py
@@ -230,10 +230,16 @@ def process_crs(crs):
     """
     try:
         import cartopy.crs as ccrs
+    except ImportError:
+        raise ImportError('Geographic projection support requires cartopy.')
+    try:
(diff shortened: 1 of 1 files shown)


maximlt:
GeoViews depends on Cartopy and it seems like Cartopy depends on pyproj. So maybe it all comes down of just telling users to install GeoViews? Or did you observe cases where this was not true and needed the more fine grained error report you're implementing in this PR?

Hoxbro:
I think this will make it a bit annoying if you have to install more than one of the packages. 

Maybe check all and then print out the packages which are missing?

maximlt:
Thanks @raybellwaves, I took the liberty to update your PR by grouping the checks so that users get a single list of the missing dependencies.

PR #1056, opened 2023-04-06
For geopandas, when passing a positional arg it is interpretted as `color`

### LINKED ISSUE #778: Positional arguments to `.polygons`
I am in the pysal tutorial and noticed that the in matplotlib the positional arg on `gdf.plot` maps to the `c` arg. In hvplot _I think_ it maps to `x` but it seems like it gets dropped. 

It seems to me that the only positional argument should be `c` OR positional arguments should not be allowed at all. 

image

jsignell:
Closes #778 

```python
import geopandas as gpd
import hvplot.pandas

countries = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))
countries.hvplot("continent")
```

image

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -706,6 +706,10 @@ class HoloViewsConverter:
                     kind = 'polygons'
                 elif geom_type in ('LineString', 'LineRing', 'Line'):
                     kind = 'paths'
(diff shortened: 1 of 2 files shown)


maximlt:
This makes sense to me, thanks!

PR #1077, opened 2023-06-02
Drop Python 3.6 and 3.7 and fix test-suite

maximlt:
As per the title.


setup.py
```diff
@@ -86,24 +86,16 @@ def get_setup_version(reponame):
...
     'matplotlib',
     'plotly',
     'pygraphviz',
-    'ipykernel <6.18.0',  # temporary
-    'numpy < 1.24', # temporary, for a numba error
     'ipywidgets',
+    'numba >=0.51.0',
+    'datashader >=0.6.5',
+    'spatialpandas >=0.4.3',
+    'hdf5 !=1.14.1',  # Gives coredump in test suite on Linux and Mac
```
maximlt, line 96: I believe `hdf5` is only available for conda and not PyPI.

Hoxbro: Thank you. I did not know that `hdf5` was conda only. Do you have any suggestions on how to handle it? 

I would make a new item in `extras_require`, which is called in the CI test. But do you have a better way to do it?

maximlt:
Perfect thanks!

Issue #1080, opened 2023-06-05
xarray points plot give key error with geo=True


rsignell-usgs:
I have an xarray dataset with lon,lat, and would like to plot as points with `geo=True`. 

 It doesn't work, giving a key error on lon.     If I convert the dataset to a dataframe, it works.

Here's a simple reproducer:
```
import numpy as np
import xarray as xr
import hvplot.xarray
import hvplot.pandas
 
ds = xr.Dataset({
    'lat': ('trajectory', np.linspace(30, 35, 15)),
    'lon': ('trajectory', np.linspace(16, 20, 15))
})
```
This doesn't work, giving:  `KeyError: 'lon'`
```
ds.hvplot.points(x='lon', y='lat', geo=True, tiles='OSM')  
```
But this *does* work: 
```
ds.to_dataframe().hvplot.points(x='lon', y='lat', geo=True, tiles='OSM')
```

keewis:
you can use `set_coords` to avoid that particular issue, but that will still fail because it breaks the assumption in `holoviews` that `lat` and `lon` are dimensions.

PR #1082, opened 2023-06-06
Work with non-gridded xarray dataset

Hoxbro:
Attempt at supporting request in https://github.com/holoviz/hvplot/issues/1080.

This is very much still work-in-progress. 

Needs: https://github.com/holoviz/holoviews/pull/5744


rsignell-usgs:
@Hoxbro is there something I should be doing here?  Just want to make sure the ball isn't in my court.

Hoxbro:
It is definitely not in your court. 

xarray support in Holoviews was designed to force gridded behavior, and we want it not always to be the case, so I have to think a bit about how to implement it (without breaking something else) :slightly_smiling_face:

PR #1084, opened 2023-06-08
Various minor docs fixes

maximlt:
- Remove defaults to pull everything from conda-forge and pyviz/label/dev
- Fix rendering the Plotly plot on the extension guide


maximlt:
Ok so the only way I found to display the plotly plot is to load `require.js` via sphinx.

Issue #1097, opened 2023-06-25
Create a Fugue integration similar to ipyvizzu


ahuang11:
I was talking to Fugue maintainers (@goodwanghan) and they suggested we could do something like:
https://github.com/vizzuhq/ipyvizzu/blob/main/src/ipyvizzu/integrations/fugue.py

so users can do something like:
```
LOAD "....csv"(header=TRUE,infer_schema=TRUE)
SELECT * FROM df WHERE name == "KATRINA" AND season == 2005
OUTPUT USING hvplot:points(x="LON",
        y="LAT",
        color="USA_WIND",
        hover_cols=["NAME", "LAT", "LON", "ISO_TIME", "USA_WIND",     "USA_PRES"],
        tiles=TRUE,
        geo=TRUE,
        cmap="Reds")
```

I imagine a separate module named `fugue.py` under https://github.com/holoviz/hvplot/tree/main/hvplot

Looks fairly straightforward and happy to implement if accepted

goodwanghan:
With this integration, hvplot will be connected with spark sql, bigquery, trino, and soon snowflake (when we release the snowflake integration)

Here is an example on bigquery: https://www.linkedin.com/search/results/content/?keywords=ipyvizzu%20bigquery&sid=s8r&update=urn%3Ali%3Afs_updateV2%3A(urn%3Ali%3Aactivity%3A7039309444190523392%2CBLENDED_SEARCH_FEED%2CEMPTY%2CDEFAULT%2Cfalse)

PR #1102, opened 2023-07-13
Adds fugue integration

ahuang11:
Closes https://github.com/holoviz/hvplot/issues/1097

Surprisingly easy to set up! cc: @goodwanghan

```python
import xarray as xr
import hvplot.pandas
import fugue.api as fa
from hvplot.fugue import _parse_hvplot

ds = xr.tutorial.open_dataset("air_temperature")
df = ds.groupby("time.season").mean().mean("lon").to_dataframe().reset_index()

fa.fugue_sql(
    """
    OUTPUT df USING hvplot:points(
        x="lat",
        y="air",
        by="season",
        size=100,
        opts={"width": 500, "height": 500}
    )
    """
)
```

image

Also, not sure if we should change this:
```
        assert_or_throw(len(dfs) == 1, FugueWorkflowError("not single input"))
```

hvplot + panel can probably dump extra dfs in tabs, but I just followed this example
https://github.com/vizzuhq/ipyvizzu/blob/main/src/ipyvizzu/integrations/fugue.py#L54


mltmess/issue_fugue.py
```diff
@@ -0,0 +1,24 @@
+import pandas as pd
```
ahuang11, line 1: What's mltmess

maximlt: Yek that's my mess!

maximlt: Removed in https://github.com/holoviz/hvplot/pull/1102/commits/7167264f7cc5469fe507703ea439ffa487f7c8a2

maximlt (approved):

Issue #1101, opened 2023-07-07
Add `threshold` kwarg; show individual points when zoomed; else datashade


ahuang11:
Can there be a `threshold` kwarg;
```python
df.hvplot("x", "y", rasterize=True, threshold=5000)
```

That does this:
```python
import holoviews as hv

from holoviews.operation.datashader import rasterize
from holoviews.operation import apply_when

points = hv.Points(np.random.randn(100000, 2))

apply_when(points, operation=rasterize, predicate=lambda x: len(x) > 5000)
```

https://discourse.holoviz.org/t/show-individual-points-when-zoomed-else-datashade/2204/5?u=ahuang11

jbednar:
Yes, see https://github.com/holoviz/holoviews/issues/1076 . I think it would need to be `rasterize_threshold` or `count_limit` or `data_size_limit`, since `threshold` could mean just about anything.

ahuang11:
To be clear, this would be in HoloViews or hvplot, or both?

jbednar:
The keyword as discussed here would be in hvplot. I'd like the same functionality available as a HoloViews option, but don't know how other HoloViews devs feel about that.

PR #1103, opened 2023-07-14
Adds resample_when; the limit before automatically applying datashade/rasterize/downsample

ahuang11:
Closes https://github.com/holoviz/hvplot/issues/1101

        The threshold before applying the operation (datashade / rasterize);
        if the number of points is below this value, the plot will
        not be rasterized or datashaded.

Needs more testing on various types of charts, but this is such a cool feature!!

https://github.com/holoviz/hvplot/assets/15331990/8da98b3b-00ec-4cb9-ab56-462276bfce3d


Hoxbro:
Could a case be made for having `rasterize`/`datashade` also accept integers as inputs, and if it was an integer it will use `apply_when`?

Also is your example running the current code? I can't see why `op_threshold` is enabled in it.

ahuang11:
Oops the video I set a default op_threshold of 5000 but I decided not to modify the default. 

Having rasterize and datashade accept integers could work. What are others preferences?

ahuang11:
I personally prefer it separate as it's more explicit.

For example, this reads better:
```python
df.hvplot.points('0', '1', rasterize=True, op_threshold=1000)
```

Than this in my opinion:
```python
df.hvplot.points('0', '1', rasterize=1000)
```

Hoxbro:
Then I would suggest going in the other direction and having two variables, `rasterize_threshold` and `datashade_threshold`, to be more explicit. Mainly because `op` does not mean anything to an average user. 

An alternative could be another prefix.

ahuang11:
I do not like having both keywords like that.

Perhaps `operation_threshold` or `rasterization_threshold` then? It'll be listed under https://hvplot.holoviz.org/user_guide/Customization.html#datashading-options

maximlt:
Generally when it's about API design I ask @jbednar, @philippjfr or @jlstevens what they think about the suggestions. They probably already thought about the suggested API a number of years ago, I also like the idea that it can help ensuring consistency with the existing APIs (name, style, etc.) across HoloViz as they've designed most of that.

jbednar:
I like aggregation_threshold=1e8, or   rasterize_threshold=1e8, or aggregate_over=1e8 or rasterize_if_over=1e8. I agree it should be one argument to cover all aggregation operations. I think there should also be a way to configure this as a default at the holoviews level, BTW, not just hvplot.

ahuang11:
Having a default will cause it to always return an overlay of the original chart element + Image (rasterized) so I'm hesitant on that.

hvplot/converter.py
```diff
@@ -1371,6 +1379,23 @@ def method_wrapper(ds, x, y):
         layers = _transfer_opts_cur_backend(layers)
         return layers
 
+    def _aggregate_obj(self, operation, obj, opts):
+        def exceeds_aggregation_threshold(plot):
+            vdim = plot.vdims[0].name
+            data = plot.data[vdim]
+            data_size = np.size(data)
```
maximlt, line 1386: Is this going to work on all data types supported by hvPlot? I haven't checked myself, asking you first to know whether you thought about it in which case you can just confirm and resolve this comment!

hvplot/converter.py
```diff
@@ -212,6 +212,12 @@ class HoloViewsConverter:
         Whether to apply rasterization using the Datashader library,
         returning an aggregated Image (to be colormapped by the
         plotting backend) instead of individual points
+    aggregation_threshold (default=None):
+        The threshold before toggling the operation (datashade / rasterize);
+        if the number of individual points exceeds this value, the plot will
+        be rasterized or datashaded; else the plot with the original points
+        will be returned instead. If this is unset, the operation
+        will always be applied.
```
maximlt, line 220: Could you clarify what you mean with:

> If this is unset, the operation will always be applied.

hvplot/converter.py
```diff
@@ -286,7 +292,7 @@ class HoloViewsConverter:
 
     _op_options = [
         'datashade', 'rasterize', 'x_sampling', 'y_sampling',
-        'aggregator'
+        'aggregator', "aggregation_threshold"
```
maximlt, line 295: Nit pick, it's probably best to keep the quotes consitant and in this case use single quotes.

Additional question: do you think `downsample` (recently added option) should be added in this list? It could be an oversight.

hvplot/converter.py
```diff
@@ -466,9 +472,12 @@ def __init__(
                     ylim = (y0, y1)
 
         # Operations
+        self.aggregation_threshold = aggregation_threshold
         self.datashade = datashade
         self.rasterize = rasterize
         self.downsample = downsample
+        if self.aggregation_threshold and not (self.datashade or self.rasterize):
```
maximlt, line 479: TBH I prefer explicit over implicit and have users specify what operation they want, do you have a strong argument for the operation be set automatically?

maximlt: If we keep the implemented behavior, the docstring of `aggregator_threshold` will have to be updated to reflect that.

maximlt (commented):
@ahuang11 do you think `aggregator_threshold` could also be used to apply the downsample operation (`lttb` by default now but will certainly be extended to `nth` for some tabular plot types and `regrid` for gridded types)? If so, should it be renamed to reflect that it just doesn't apply to aggregator operations like rasterize and datashade (assuming for instance that lttb isn't an aggregating operation, which I think is correct to say?).

We're seriously lacking a user guide that demonstrates how to plot large data and where we could demonstrate that feature and how other operations work, not your fault though!

hvplot/converter.py
ahuang11: Yes I believe it always returns a numpy array with `plot.data[vdim]`

maximlt:
After a discussion with Andrew we're suggesting `resample_when` instead of `aggregator_threshold`, `resample` is feels more understandable and general and `when` is more inline with `apply_when`. I also liked `resample_above` but thought that maybe it's too specific and might limit the type of value it could accept in the future (e.g. some predicate function), `when` is more future-proof. Interested in feedback from others on this name, naming is hard!

@ahuang11 the tests are now failing, were they passing before? I would also suggest making `resample_when` apply to the `downsample` op and not just `rasterize/datashade`.

ahuang11:
Something seems incorrect with downsample=True; the ranges keep reverting on zoom.
```python
import hvplot.pandas
import pandas as pd
import numpy as np
import xarray as xr
import panel as pn

df = pd.DataFrame(
    np.random.multivariate_normal((0, 0), [[0.1, 0.1], [0.1, 1.0]], (100000,))
)
df.columns = ["A", "B"]
dmap = df.hvplot.scatter("A", "B", downsample=True)

pn.panel(dmap).show()
```

https://github.com/holoviz/hvplot/assets/15331990/6c90498a-a2c1-41cd-a67d-d3c40d900e7a

maximlt:
> Added support for downsample

Thanks a lot for that! I did a little bit of refactoring, in particular I found out that calling `df.hvplot.points(resample_when=100, rasterize=True)` would lead to errors as the generated plot element has no vdims but two kdims `x` and `y`, so `plot.vdims[0]` raised an error. So I decided to leverage the fact that elements implement `__len__` and `len(plot)` to obtain the size. I've also reworked the docstring as I just couldn't wrap my head around it.

I'm going to merge now, don't hesitate opening new PRs to refine the functionality if you thing it needs that.

PR #1113, opened 2023-07-21
Test that no projection is set when tiles is used without geo=True

maximlt:
Simply add some tests for https://github.com/holoviz/hvplot/pull/1083/files


hvplot/tests/testgeo.py
```diff
@@ -174,6 +183,45 @@ def test_plot_with_specific_gv_tile_obj(self):
...
+    def setup_class(cls):
+        import hvplot.pandas  # noqa
+
+    def test_plot_tiles_doesnt_set_geo(self, simple_df):
+        plot = simple_df.hvplot.points('x', 'y', tiles=True)
+        assert len(plot) == 2
+        assert isinstance(plot.get(0), hv.Tiles)
+        assert 'openstreetmap' in plot.get(0).data
+        bk_plot = bk_renderer.get_plot(plot)
+        assert bk_plot.projection == 'mercator'
```
maximlt, line 198: This is true if the test is executed on its own. When the whole module runs, GeoViews is imported and this comes with some side-effects, `bk_plot.projection` becomes Cartopy CRS (Mercator). I'm not sure this side-effect is desired.

Issue #1115, opened 2023-07-21
hvplot.show() does not return the expected threading.Thread | Server object


dogbunny:
#### ALL software version info
hvplot 0.8.4

#### Description of expected behavior and the observed behavior
calling hvplot.show() returns the threading.Thread or Server object (depending on whether threading is true or false)

The show method likely just needs to have a return added: https://github.com/holoviz/hvplot/blob/main/hvplot/utilities.py#L29

#### Complete, minimal, self-contained example code that reproduces the issue
```
import hvplot
import hvplot.pandas
import pandas as pd
import numpy as np

ts = pd.Series(np.random.randn(1000), index=pd.date_range('1/1/2000', periods=1000))
ts = ts.cumsum()
plot = ts.hvplot()
s = hvplot.show(plot, **{'threaded': True})
s.stop()
```

#### Stack traceback and/or browser JavaScript console output
In [12]: s.stop()
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[12], line 1
----> 1 s.stop()

AttributeError: 'NoneType' object has no attribute 'stop'

PR #1121, opened 2023-07-26
Update show method to return the configured Server or StoppableThread

dogbunny:
Fix to https://github.com/holoviz/hvplot/issues/1115

Note:  There do not appear to be any existing tests for this method or file.

### DIFF
--- a/hvplot/utilities.py
+++ b/hvplot/utilities.py
@@ -10,7 +10,7 @@ output = _hv.output
 
 save = _hv.save
 
-def show(obj, title=None, port=0, **kwargs) -> _pn.io.server.Server | _pn.io.server.StoppableThread:
+def show(obj, title=None, port=0, **kwargs):
     """
     Displays HoloViews objects in and outside the notebook
 
--- a/hvplot/utilities.py
+++ b/hvplot/utilities.py
@@ -1,5 +1,3 @@
-from __future__ import annotations
-
 import panel as _pn
 import param
 import holoviews as _hv
--- a/hvplot/utilities.py
+++ b/hvplot/utilities.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import panel as _pn
 import param
 import holoviews as _hv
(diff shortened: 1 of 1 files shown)


hvplot/utilities.py
```diff
@@ -22,11 +22,14 @@ def show(obj, title=None, port=0, **kwargs):
         Allows specifying a specific port
     **kwargs: dict
         Additional keyword arguments passed to Panel show method.
+    Returns
+    -------
+    a Server | threading.Thread (if threaded=true)
```
Hoxbro, line 27: Are you sure it is returning a thread if `threaded=True`?

dogbunny: If I'm following the code correctly, utility.show() is calling `[panel].pane.HoloViews(obj).show(title, port, **kwargs)` which then calls `serve` from panel.io.server.py.  Both of those have return values of `-> threading.Thread | Server:`
I've tested locally and it does work as expected there.

Hoxbro: Running this
``` python
import hvplot
import hvplot.pandas
import numpy as np
import pandas as pd

ts = pd.Series(np.random.randn(1000), index=pd.date_range("1/1/2000", periods=1000))
ts = ts.cumsum()
plot = ts.hvplot()
s = hvplot.show(plot, threaded=True)
print(type(s))
s.stop()
```

Outputs:
``` python
<class 'panel.io.server.StoppableThread'>
Launching server at http://localhost:35483
```

I think it should be `panel.io.server.StoppableThread` instead of `threading.Thread` to be correct.

(I also think the same is the case for Panel docstring, so if you want, you are welcome to open a PR on Panel)

dogbunny: You're right it's `panel.io.server.StoppableThread | panel.io.server.Server`.  I'll update this PR and submit one for panel to fix its return value as well.

hvplot/utilities.py
```diff
@@ -8,7 +10,7 @@
 
 save = _hv.save
 
-def show(obj, title=None, port=0, **kwargs):
+def show(obj, title=None, port=0, **kwargs) -> _pn.io.server.Server | _pn.io.server.StoppableThread:
```
Hoxbro, line 13:
```suggestion
def show(obj, title=None, port=0, **kwargs):
```

Even though this is correct, hvplot is not yet ready to use type annotations.

dogbunny: that's unfortunate, will remove those.

Hoxbro:
Thank you for this PR! :1st_place_medal:

Issue #1120, opened 2023-07-24
hvplot setting xlim/ylim should first sel data when rasterize=True/datashade=True


ahuang11:
We can make an optimization in hvplot; namely, if ylim is provided like...

16 seconds
```
t = time.perf_counter()
display(ds.hvplot("lon", "lat", rasterize=True, ylim=(26, 30)))
elapsed = time.perf_counter() - t
print(f"Elapsed time: {elapsed:0.4f} seconds")
```

To something similar:

3 seconds
```
t = time.perf_counter()
display(ds.sel(lat=slice(30, 26)).hvplot("lon", "lat", rasterize=True))
elapsed = time.perf_counter() - t
print(f"Elapsed time: {elapsed:0.4f} seconds")
```

https://discourse.holoviz.org/t/fast-visualization-of-large-datasets-without-saving-to-disk-using-hvplot-kerchunk-and-datashader/5751/2?u=ahuang11

philippjfr:
I agree with this, in hvPlot we have the benefit that data and plotting operations can be combined and thereby optimize the rendering experience. The only concern here is that xlim/ylim only determine the initial extent and when you zoom/pan you expect the rest of the data to be revealed. However I think this could be solved by providing xlim/ylim as inputs to x_range/y_range on the rasterize or other resampling operation.

PR #1126, opened 2023-08-09
Use x/ylims as datashader operations' x/yranges

ahuang11:
Closes https://github.com/holoviz/hvplot/issues/1120


jbednar (approved):
Looks right to me, but I'm surprised we weren't always doing this!

ahuang11:
Actually I think we might already be doing this here so this PR is not necessary.
https://github.com/holoviz/hvplot/blob/main/hvplot/converter.py#L1303-L1306
cc @philippjfr

PR #1127, opened 2023-08-23
Expose timeseries downsampling as an option

philippjfr:
Exposes `downsample1d` as an option like `rasterize` and `datashade`.


maximlt:
- [ ] Need to add docs

PR #1130, opened 2023-08-29
Suggest to install datashader and geoviews from defaults

droumis:
Old:

`ImportError: In order to use datashading features the Datashader library must be available. It can be installed with:
  conda install -c pyviz datashader`

New:

`ImportError: In order to use datashading features the Datashader library must be available. It can be installed with:
  conda install datashader`

Same for GeoViews


maximlt:
Perfect thanks!

PR #1131, opened 2023-09-07
Add Dask DataFrame plot test

Hoxbro:
Continue the work done in #1129 by adding simple plotting tests of `dask.DataFrame` and `dask.Series`. 

I discovered a small bug that `kde` and `density` did error out because `pd.melt` was used.

Which means that something like this errors out on main:
``` python
import dask.dataframe as dd
import hvplot.pandas
import pandas as pd

y = ["A", "B", "C", "D"]
kind = "density"
df = dd.from_pandas(pd._testing.makeDataFrame(), npartitions=2)
df.hvplot(y=y, kind=kind)
```


maximlt:
LGTM! I've added `dask` to the core test dependencies, its support should definitely be tested as often as possible.

It took me a while before I figured out this PR wasn't targeting *main*, once the branch it targets is merged I will point it to *main* so the commit history is as expected.

Hoxbro:
> LGTM! I've added dask to the core test dependencies, its support should definitely be tested as often as possible.

Fine by me. 

> It took me a while before I figured out this PR wasn't targeting main, once the branch it targets is merged I will point it to main so the commit history is as expected.

Yeah, didn't want the merge conflict, so based in on the Polars PR. This should automatic be re-targeted when  Polars Pr are merged. Though, something like this could be needed `git merge main --strategy-option ours`.

maximlt (approved):

PR #1132, opened 2023-09-09
Support plots that use `by` with `rasterize` with hv.ImageStack

ahuang11:
Alongside https://github.com/holoviz/holoviews/pull/5751
```python
import pandas as pd
import hvplot.pandas
import holoviews as hv
hv.extension('bokeh')

df = pd.DataFrame(
    {
        "time": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
        "value": [1, 2, 3, 4, 5, 6, 7, 8, 9, 10],
        "by": ["d", "d", "d", "a", "a", "b", "b", "b", "b", "b"],
    }
)

df.hvplot(
    x="time",
    y="value",
    by="by",
    rasterize=True,
    color_key="RdBu_r"
)

```

image

colorbar is not yet supported in HoloViews.

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -1330,7 +1330,7 @@ class HoloViewsConverter:
                 opts['rescale_discrete_levels'] = self._plot_opts['rescale_discrete_levels']
         else:
             operation = rasterize
-            if hv.__version__ < '1.18.0':
+            if Version(hv.__version__) < Version('1.18.0a1'):
                 eltype = 'Image'
             else:
                 eltype = 'ImageStack' if self.by else 'Image'
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -468,10 +468,7 @@ class HoloViewsConverter:
 
         # By type
         self.subplots = subplots
-        if subplots:
-            self._by_type = NdLayout
-        else:
-            self._by_type = NdOverlay
+        self._by_type = NdLayout if subplots else NdOverlay
 
         self._backend = Store.current_backend
         if hvplot_extension.compatibility is None:
@@ -1338,7 +1335,6 @@ class HoloViewsConverter:
                 style['cmap'] = self._style_opts['cmap']
             if self._dim_ranges.get('c', (None, None)) != (None, None):
                 style['clim'] = self._dim_ranges['c']
-            print(style)
 
         processed = operation(obj, **opts)
 
@@ -1514,12 +1510,10 @@ class HoloViewsConverter:
         cur_opts = {
             element.name: cur_el_opts,
             'NdOverlay': filter_opts('NdOverlay', dict(self._overlay_opts, batched=False), backend='bokeh'),
-            'ImageStack': filter_opts('ImageStack', dict(self._overlay_opts, batched=False), backend='bokeh'),
         }
         compat_opts = {
             element.name: compat_el_opts,
             'NdOverlay': filter_opts('NdOverlay', dict(self._overlay_opts), backend=self._backend_compat),
-            'ImageStack': filter_opts('ImageStack', dict(self._overlay_opts), backend=self._backend_compat),
         }
 
         ys = [y]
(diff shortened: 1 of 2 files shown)


ahuang11:
I think it boils down to this not working as expected:
```python
import holoviews as hv
import xarray as xr
import numpy as np
import panel as pn
from holoviews.operation.datashader import rasterize

hv.extension("bokeh")


da = xr.DataArray(coords={"x": [1, 2, 3], "y": [6, 7, 8]}, dims=["x", "y"])

da_1 = da.copy()
da_1.name = "a"
da_1.values = [[np.nan, np.nan, 1], [np.nan] * 3, [np.nan] * 3]

da_2 = da.copy()
da_2.name = "b"
da_2.values = [[np.nan] * 3, [1, 1, np.nan], [np.nan] * 3]

da_3 = da.copy()
da_3.name = "c"
da_3.values = [[np.nan] * 3, [np.nan] * 3, [1, 1, 1]]

ds = xr.merge([da_1, da_2, da_3])

hv.ImageStack(ds.to_array().fillna(0), kdims=["x", "y"])
```

hvplot/converter.py
```diff
@@ -1330,7 +1330,7 @@ def method_wrapper(ds, x, y):
                 opts['rescale_discrete_levels'] = self._plot_opts['rescale_discrete_levels']
         else:
             operation = rasterize
-            eltype = 'Image'
+            eltype = 'ImageStack' if self.by else 'Image'
```
Hoxbro, line 1333: We should safeguard this so it only happens with a holoviews version which support ImageStack (1.18)

Hoxbro (commented):
When running the code (with https://github.com/holoviz/holoviews/pull/5751), I get the following plot. I suspect some custom handling of the hover tool is needed. 

image

maximlt:
Just noting, we need to open an issue to record that this needs to be documented when HoloViews is released with the new feature this PR depends on.

ahuang11:
I am confused, what needs to be documented?

maximlt:
Sorry for the lack of context. `rasterize` together with `by` should certainly be documented, Simon said he already has a notebook in preparation. I'll open an issue.

ahuang11:
> When running the code (with [holoviz/holoviews#5751](https://github.com/holoviz/holoviews/pull/5751)), I get the following plot. I suspect some custom handling of the hover tool is needed.
> 
> image

Seems like this needs to be fixed in holoviews
image

Hoxbro:
I have left a comment in the holoviews issue about the hover tool. I'm also a bit surprised that the colorbar stops at 0.8.

hvplot/converter.py
```diff
@@ -1330,7 +1330,10 @@ def method_wrapper(ds, x, y):
                 opts['rescale_discrete_levels'] = self._plot_opts['rescale_discrete_levels']
         else:
             operation = rasterize
-            eltype = 'ImageStack' if self.by else 'Image'
+            if hv.__version__ < '1.18.0':
```
Hoxbro, line 1333: This does not work always work. `"1.18.0" < "1.2.2"`.

```suggestion
            if Version(hv.__version__) < Version('1.18.0a1'):
```

where Version is `from packaging.version import Version`. `a1` is because this should work with the a1 release.

hvplot/tests/testoperations.py
```diff
@@ -194,6 +195,11 @@ def test_datashade_rescale_discrete_levels_default_True(self):
         actual = plot.callback.inputs[0].callback.operation.p['rescale_discrete_levels']
         assert actual is expected
 
+    def test_rasterize_by(self):
```
Hoxbro, line 198: This should also have a version check, but let us ignore it if all tests pass.

Hoxbro (approved):
If this is ready, you can merge it.

hvplot/tests/testoperations.py
maximlt: Yep for us it works well but sometimes distros that re-package our tools complain about this kind of stuff (and probably just patch things).

ahuang11: Ok I'll add a check in another PR

ahuang11: https://github.com/holoviz/hvplot/pull/1144

Issue #1095, opened 2023-06-22
Allow string formatting by column name for labels' text


ahuang11:
I think this would be really cool
`df.hvplot.labels("Longitude", "Latitude", "{City} ({Latitude}N, {Longitude}E)", by="City")`

```python
import pandas as pd

df = pd.DataFrame(
    {'City': ['Buenos Aires', 'Brasilia', 'Santiago', 'Bogota', 'Caracas'],
     'Country': ['Argentina', 'Brazil', 'Chile', 'Colombia', 'Venezuela'],
     'Latitude': [-34.58, -15.78, -33.45, 4.60, 10.48],
     'Longitude': [-58.66, -47.91, -70.66, -74.08, -66.86]})
df.hvplot.points("Longitude", "Latitude", by="City") * df.hvplot.labels("Longitude", "Latitude", "{City} ({Latitude}, {Longitude})", by="City")
```

A workaround:
```
import pandas as pd

df = pd.DataFrame(
    {'City': ['Buenos Aires', 'Brasilia', 'Santiago', 'Bogota', 'Caracas'],
     'Country': ['Argentina', 'Brazil', 'Chile', 'Colombia', 'Venezuela'],
     'Latitude': [-34.58, -15.78, -33.45, 4.60, 10.48],
     'Longitude': [-58.66, -47.91, -70.66, -74.08, -66.86]})

df["Label"] = [f"{row['City']} ({row['Latitude']}N, {row['Longitude']}E)" for _, row in df.iterrows()]
df.hvplot.points("Longitude", "Latitude", by="City") * df.hvplot.labels("Longitude", "Latitude", "Label", by="City")
```
<img width="835" alt="image" src="https://github.com/holoviz/hvplot/assets/15331990/8ef06cfe-6fcb-4550-a615-6e381ccafb35">

PR #1140, opened 2023-09-15
Allow string formatting for labels

ahuang11:
Closes https://github.com/holoviz/hvplot/issues/1095

I thought about putting this in holoviews, but passing format strings to vdims seems a tad strange...
`hv.Labels(kdims=["x", "y"], vdims=["{x} {y}"])`, but passing it in hvplot feels natural

```python
import pandas as pd
import hvplot.pandas

df = pd.DataFrame(
    {
        "City": ["Buenos Aires", "Brasilia", "Santiago", "Bogota", "Caracas"],
        "Country": ["Argentina", "Brazil", "Chile", "Colombia", "Venezuela"],
        "Latitude": [-34.58, -15.78, -33.45, 4.60, 10.48],
        "Longitude": [-58.66, -47.91, -70.66, -74.08, -66.86],
    }
)
df.hvplot.points("Longitude", "Latitude", by="City") * df.hvplot.labels(
    "Longitude", "Latitude", "{City} ({Latitude}, {Longitude})", by="City"
)
```

image


hvplot/tests/testcharts.py
```diff
@@ -347,6 +347,11 @@ def test_errorbars_no_hover(self):
         bkplot = Store.renderers['bokeh'].get_plot(plot)
         assert not bkplot.tools
 
+    def test_labels_format(self):
+        plot = self.df.hvplot("x", "y", text="({x}, {y})", kind="labels")
+        print(plot)
```
maximlt, line 352: Is this still WIP?

ahuang11: Heh I was testing and I forgot to uncomment.

ahuang11: Fixed now.

examples/reference/pandas/labels.ipynb
```diff
@@ -42,6 +42,23 @@
     "df.hvplot.labels(x='Longitude', y='Latitude', text='City', text_baseline='bottom', hover=False) * \\\n",
     "df.hvplot.labels(x='Longitude', y='Latitude', text='Country', text_baseline='top', hover=False)"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "It's also possible to provide a template string containing the names of the columns and reduce the font size."
```
maximlt, line 50: Thanks for updating the docs! The docstring of `labels` in `core.py` should also be adapted.

hvplot/converter.py
```diff
@@ -2041,7 +2042,20 @@ def labels(self, x=None, y=None, data=None):
         self.use_index = False
         data, x, y = self._process_chart_args(data, x, y, single_y=True)
 
-        text = self.kwds.get('text', [c for c in data.columns if c not in (x, y)][0])
+        text = self.kwds.get('text')
+        if not text:
+            text = [c for c in data.columns if c not in (x, y)][0]
```
maximlt, line 2047: Edge-case but if the first column that is not x/y contains a string with `'{foo}'` this might cause some problems?

maximlt (changes requested):
This seems like a useful feature to me (even if I have rarely if ever used `labels`!).

I would have expected to be able to use formatting options like `{Density:.2f}` except that it raised an error when I tried. We could merge this feature without support for that, it'd be far less useful though. If you try to implement it, I'd suggest leveraging:
- `'{a} {b}'.format(a=1)` raises a `KeyError`, so you don't need to check whether some columns are wrongly typed/missing, it does it for you
- `'{a} {b}'.format(a=1, b=2, c=3)` doesn't raise an error so you can pass all the columns data to `.format` and it'll happily use only what it needs.

hvplot/converter.py
ahuang11: I'm not sure I follow; can you clarify this point?

maximlt: If I call `df.hvplot.labels()` without passing `text=...` it will default to the first column that is not x or y, if that column name is e.g. `'Volume {m3}` (edge-case-y, I know!), the code will break as it'll assume `m3` is a column name and find it missing.

ahuang11: Okay I added support for this edge case: if there's only one missing col, and there's only one {} in the col, it uses the legacy behavior.

It does not support Volume {m3} {m4} though.

ahuang11:
I added formatting capabilities.

hvplot/converter.py
maximlt: I believe this could be approached differently by just skipping the whole formatting code path if `text` is not provided.

hvplot/converter.py
```diff
@@ -2041,7 +2042,22 @@ def labels(self, x=None, y=None, data=None):
...
+        text_cols = re.findall(r"\{([^{}]+)\}", text)
```
maximlt, line 2048: You didn't like my suggestions to leverage `.format()` ? :D I find a bit weird having to handle ourself parsing format strings with `re` and then splitting on `:`, could it be we're missing cases? `!r` comes to mind, even if it doesn't look useful in this case someone might still want to try that.

ahuang11: Thanks for the insight and insisting; I didn't understand how it was relevant initially, but now it's much more elegant!

maximlt: Perfect! Thanks a lot for bearing with me :D

maximlt (approved):

PR #1142, opened 2023-09-18
Silence GDAL exceptions warning and filter out warnings from the test suite

### LINKED ISSUE #1104: GDAL 4.0 warning from osgeo
Running this:

```
import hvplot.xarray
import rioxarray as rxr

# Downloaded from: https://oin-hotosm.s3.amazonaws.com/5d7dad0becaf880008a9bc88/0/5d7dad0becaf880008a9bc89.tif
cog_url = "/home/shh/Downloads/5d7dad0becaf880008a9bc89.tif"
da = rxr.open_rasterio(cog_url)
plot = da.hvplot.rgb(x="x", y="y", rasterize=True, projection="Robinson")
plot
```

I'm getting this warning:
`/home/shh/miniconda3/envs/holoviz/lib/python3.11/site-packages/osgeo/osr.py:385: FutureWarning: Neither osr.UseExceptions() nor osr.DontUseExceptions() has been explicitly called. In GDAL 4.0, exceptions will be enabled by default.`

This should be looked into, and check which setting we should use.

Hoxbro:
Trying to handle some of the warnings seen in CI.

Have ignored geo-related warnings for now. 

Fixes #1104


hvplot/util.py
```diff
@@ -121,6 +121,7 @@ def proj_to_cartopy(proj):
     import cartopy.crs as ccrs
     try:
         from osgeo import osr
+        getattr(osr, "UseExceptions", lambda: None)()
```
maximlt, line 124: Do you think there's a risk this breaks users code if they use GDAL on their side together with hvPlot? I'm not sure what it means to "use exceptions"?

maximlt (approved):

PR #1146, opened 2023-09-20
Param 2: don't call param.objects() before `super().__init__` in the explorer

maximlt:
The explorer accepts in its constructor kwargs that are dispatched to the various controls (which wouldn't work well if two controls share the same attribute name but that's another issue) when they are instantiated. As such it needs to prune the kwargs before calling `super().__init__(**params)`. This was done by gathering all the controls classes calling `.param.objects()` and doing some filtering. Param 2 will raise an error (or a warning) when `param.objects()` is accessed before `super().__init__` is called. In this PR I arrange the code to call `param.objects()` after `super().__init__`.


Hoxbro:
Tests are green, so I will merge this.

PR #1155, opened 2023-09-26
Use dynspread=True by default for explorer if datashade/rasterize

ahuang11:
I think it's more user friendly if dynspread is True, or else can barely see anything:
image

With dynspread=True:
image

### DIFF
--- a/hvplot/ui.py
+++ b/hvplot/ui.py
@@ -288,7 +288,7 @@ class Operations(Controls):
     aggregator = param.Selector(default=None, objects=AGGREGATORS, doc="""
         Aggregator to use when applying rasterize or datashade operation.""")
 
-    dynspread = param.Boolean(default=False, doc="""
+    dynspread = param.Boolean(default=True, doc="""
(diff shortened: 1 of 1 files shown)


maximlt:
I think `dynspread` has docs, except that Bokeh doesn't support tooltips for all the widgets?

maximlt:
Alternatively, the explorer could warn the user when they set rasterize/datashader on a dataset that has too little data.

ahuang11:
Is there a downside to setting dynspread=True in any case though?

maximlt:
Maybe, if so that would be the default in hvPlot?

ianthomas23:
You wouldn't want to use any form of spreading if you are rendering antialiased lines in datashader. But I don't know if antialiased lines have reached hvplot yet.

jbednar:
BTW, I think Explorer should also be enabling either LTTB or rasterize automatically when the data size is too large. Maybe LTTB by default for line data >100000 points, and rasterize by default for point data > 100000 points?

ahuang11:
What's LTTB?

PR #1156, opened 2023-09-26
See Explorer output code within a tab

ahuang11:
image


hvplot/ui.py
```diff
@@ -503,8 +504,8 @@ def _plot(self, *events):
         if kwargs.get("geo"):
             if "crs" not in kwargs:
                 xmax = np.max(np.abs(self.xlim()))
-                self.crs = "PlateCarree" if xmax <= 360 else "GOOGLE_MERCATOR"
-                kwargs["crs"] = self.crs
+                self.geographic.crs = "PlateCarree" if xmax <= 360 else "GOOGLE_MERCATOR"
```
maximlt, line 507: Why is this part of the code touched?

maximlt (changes requested):
Oops failing tests 🙃 

Nice feature! Could you please add some tests and documentation?

In your screenshot is it normal there's no `kind` in the call to `hvplot()`?


Related work/thoughts

- there's an issue somewhere to turn `plot_code` into a Parameter (that'd maybe just `code`/`plot_repr`/`code_repr`/... to avoid breaking current users of `plot_code`)
- `plot_code` doesn't infer the variable name, it defaults to `df` as shown in your example. I found this cool trick that leverages f-strings to obtain the name of a variable, might be worth considering.
```python
my_df = pd.DataFrame()
var_name = f'{my_df=}'.split('=')[0]
```

maximlt:
Also @ahuang11 you could please add to the 0.9.0 milestone the PRs you *need* in that release?

hvplot/ui.py
```diff
@@ -503,8 +509,8 @@ def _plot(self, *events):
...
+                kwargs["crs"] = self.geographic.crs
```
ahuang11, line 513:
```suggestion
                self.crs = "PlateCarree" if xmax <= 360 else "GOOGLE_MERCATOR"
                kwargs["crs"] = self.crs
```

hvplot/ui.py
```diff
@@ -515,6 +521,12 @@ def _plot(self, *events):
 
             feature_scale = kwargs.pop("feature_scale", None)
             kwargs['features'] = {feature: feature_scale for feature in kwargs.pop("features", [])}
+        else:
+            # Always remove these intermediate keys from kwargs
+            kwargs.pop('geo')
+            kwargs.pop('crs_kwargs', {})
+            kwargs.pop('projection_kwargs', {})
+            kwargs.pop('feature_scale', None)
```
ahuang11, line 529:
```suggestion
```

hvplot/ui.py
```diff
@@ -704,6 +736,14 @@ def __init__(self, ds, **params):
             params["kind"] = "image"
         super().__init__(ds, **params)
 
+    @property
+    def _var_name(self):
+        return "ds"
```
Hoxbro, line 741: This could also be a `xr.DataArray`?

ahuang11: Yes, I figured, it's probably more accurate to reproduce what I'm doing internally to get the explorer.

Hoxbro: I misread the code. You did what I expected. :upside_down_face:

hvplot/ui.py
```diff
@@ -551,9 +560,25 @@ def _plot(self, *events):
...
+    @property
+    def _var_name(self):
+        return "data"
+
+    @property
+    def _backend(self):
+        return "pandas"
+
+    def _update_refresh_plot(self, event):
+        self.refresh_plot = event.new
```
ahuang11, line 581:
```suggestion
```

hvplot/ui.py
```diff
@@ -551,9 +560,22 @@ def _plot(self, *events):
         finally:
             self._layout.loading = False
 
+    def _code(self):
+        self.code = self._build_code_snippet()
+        self._code_pane.object = f"""```python\nimport hvplot.{self._backend}\n\n{self.code}\n```"""
```
maximlt, line 565: I would not include the import, it doesn't bring much in my opinion.

hvplot/ui.py
```diff
@@ -289,10 +293,8 @@ class Geographic(Controls):
     @param.depends('geo',  watch=True, on_init=True)
     def _update_crs_projection(self):
         enabled = bool(self.geo or self.project)
-        self.param.crs.constant = not enabled
-        self.param.crs_kwargs.constant = not enabled
-        self.param.projection.constant = not enabled
-        self.param.projection_kwargs.constant = not enabled
+        for key in GEO_KEYS:
+            getattr(self.param, key).constant = not enabled
```
maximlt, line 297:
```suggestion
            self.param[key]constant = not enabled
```

maximlt (commented):
I just gave a quick review before @Hoxbro's one.

I'm the one who suggested adding the `code` Parameter to the explorer. IIRC, this actually was a suggestion made by @MarcSkovMadsen a while back, I'd be happy to hear from you Marc whether this is what you expected at the time?

If it's alright, I believe the new `code` Parameter should be documented. I'm also wondering whether `plot_code()` should be deprecated, the only advantage it has is that you can provide a name for the data variable. So now we have `.plot_code(var_name='df')`, `.code` and `_build_code_snippet()`. We could have just `code` so not allow users to provide a variable name? The minimum we should do is merge `plot_code` and `_build_code_snippet`, if we want to expose such a method as public API.

hvplot/ui.py
ahuang11: I disagree; it reduces the friction for the user to immediately get started--else, the user needs to manually type `import hvplot.pandas`

If they do not want the import, they can simply copy up to the line below.

ahuang11:
Initially I wanted to merge `_build_code_snippet` and `plot_code`, but I believe you suggested:
`there's an issue somewhere to turn plot_code into a Parameter (that'd maybe just code/plot_repr/code_repr/... to avoid breaking current users of plot_code)`

hvplot/ui.py
maximlt: IIRC you have to run the import first in a notebook for the explorer to work. What use case do you have in mind?

maximlt:
Sorry I wasn't clear enough, I meant that we couldn't just remove `plot_code` or turn it into a Parameter. I believe though it's fine changing its code, I doubt there's someone doing any kind of parsing on its output.

hvplot/ui.py
```diff
@@ -289,10 +293,8 @@ class Geographic(Controls):
     @param.depends('geo',  watch=True, on_init=True)
     def _update_crs_projection(self):
         enabled = bool(self.geo or self.project)
-        self.param.crs.constant = not enabled
-        self.param.crs_kwargs.constant = not enabled
-        self.param.projection.constant = not enabled
-        self.param.projection_kwargs.constant = not enabled
+        for key in GEO_KEYS:
+            self.param[key]constant = not enabled
```
maximlt, line 297:
```suggestion
            self.param[key].constant = not enabled
```

hvplot/ui.py
ahuang11: This doesn't necessarily have to run in notebook; they can copy it to a script or in the future, if they launch the explorer from CLI.

ahuang11:
Okay I refactored this so that _build_code_snippet is now merged into plot_code and also cleaned up unnecessary redundancy with _code.

ahuang11:
Tests failing seem unrelated:
```python

File /usr/share/miniconda3/envs/test-environment/lib/python3.8/site-packages/geoviews/data/geom_dict.py:327, in geom_from_dict(geom, xdim, ydim, single_type, multi_type)
    325 split_holes = geom.pop('holes', None)
    326 if split_holes is not None and len(split_holes) != len(split_geoms):
--> 327     raise DataError('Polygons with holes containing multi-geometries '
    328                     'must declare a list of holes for each geometry.')
    330 if single_type is Point:
    331     if len(splits) > 1 or any(len(g) > 1 for g in split_geoms):

DataError: Polygons with holes containing multi-geometries must declare a list of holes for each geometry.
```

maximlt:
- I merged main into this branch
- I decided against having `import.<backend>` displayed in the code
  - I believe it's not always going to be very useful
  - We've had issues (resolved now I *think*) with handling multiple imports
  - So, I wasn't sure about this feature, and as it's adding is easier than removing, I prefer postponing adding this feature in a next release, if we decide it's the right thing to do

PR #1157, opened 2023-09-28
Add tests for `downsample`

maximlt:
Follow-up of https://github.com/holoviz/hvplot/pull/1127 with some tests and indicating that downsampling is available only if HoloViews >= 1.16 is installed.

### DIFF
--- a/hvplot/tests/testoperations.py
+++ b/hvplot/tests/testoperations.py
@@ -235,3 +235,24 @@ class TestChart2D(ComparisonTestCase):
                               x_sampling=5, y_sampling=2)
         assert all(plot.data.x.diff('x').round(0) == 5)
         assert all(plot.data.y.diff('y').round(0) == 2)
+
+
+class TestDownsample(ComparisonTestCase):
+    def setUp(self):
+        import hvplot.pandas # noqa
+        self.df = pd.DataFrame(np.random.random(100))
+
+    def test_downsample_default(self):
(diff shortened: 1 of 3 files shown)


hvplot/tests/testoperations.py
```diff
@@ -235,3 +235,24 @@ def test_plot_resolution_with_rasterize(self, kind, element):
...
+    def setUp(self):
+        import hvplot.pandas # noqa
+        self.df = pd.DataFrame(np.random.random(100))
+
+    def test_downsample_default(self):
+        from holoviews.operation.downsample import downsample1d
+
+        plot = self.df.hvplot.line(downsample=True)
+
+        assert isinstance(plot.callback.operation, downsample1d)
```
Hoxbro, line 250:
```suggestion
        assert isinstance(plot.callback.operation, downsample1d)
        assert plot.callback.operation.algorithm == "lttb"
```

Just to be safe.

maximlt: Yep good check! By the way, do you know why `nth` wasn't exposed in hvPlot?

Hoxbro: Other than it being a suboptimal algorithm, I can't think of a reason. Implementing it would be straightforward, but there is no need to expose it if we don't want people to use it. 

Do you have a better explanation, @philippjfr?

PR #1159, opened 2023-10-01
Add support for opts in explorer

ahuang11:
As title says, allows using opts; very crude at the moment simply using dict, but it's still useful as an escape hatch.

image


maximlt:
I'm not sure to be convinced by this functionality, it seems to only cater to advanced users while it is likely a lot of explorer users are beginner hvPlot users (since it's a good way to access hvPlot's plotting API without having to learn dozens of parameters). I can easily imagine there will be a lot of users clicking on this generic `Opts` tab and being puzzled by what they see.

If it's a feature that you believe would have unblocked you while working with the explorer, let's see how it could be adapted so that it's obvious it's for advanced usage. For instance, it's not uncommon for GUIs have advanced options that are made visible only after ticking a checkbox. Some text around the options widget would also help understand what kind of input you can provide, with an example.

ahuang11:
I don't view the explorer as just for beginners. I imagine it's for both beginners and advanced users, with advanced users using it to quickly take a look at their files as others have described [here](https://discourse.pangeo.io/t/do-you-use-panoply-ncview-other-command-line-viz-tool/3693/5). One of the things users need to quickly take a look is adjusting the colorbar [like this](https://atdotpy.substack.com/p/108-want-to-align-colorbar-ticks/comments?utm_source=%2Fsearch%2Fcolor_levels&utm_medium=reader2).

I agree that additional text can be added to help the user out.

hvplot/tests/testui.py
```diff
@@ -269,6 +274,40 @@ def test_explorer_code_gridded_dataarray():
...
+            y='lat'
+        ).opts(
+            color_levels=3
+        )""")
+
+    assert explorer._code_pane.object == dedent("""\
+        ```python
+        import hvplot.xarray
+
+        da.hvplot(
```
maximlt, line 298: Why are the options present twice, in the `hvplot(...)` call and in `hvplot().opts(...)`?

hvplot/ui.py
```diff
@@ -540,16 +549,17 @@ def _plot(self, *events):
...
+
         df = self._data
         if len(df) > MAX_ROWS and not (self.kind in KINDS["stats"] or kwargs.get('rasterize') or kwargs.get('datashade')):
             df = df.sample(n=MAX_ROWS)
         self._layout.loading = True
         try:
             self._hvplot = _hvPlot(df)(
                 kind=self.kind, x=self.x, y=y, by=self.by, groupby=self.groupby, **kwargs
-            )
+            ).opts(**opts_kwargs)
```
maximlt, line 562: Why not the options to the converter directly? hvPlot has some special options handling given the backend configuration https://hvplot.holoviz.org/user_guide/Plotting_Extensions.html#plot-styling-options

maximlt (commented):
For this PR to be merged and released in hvPlot 0.9.0, in addition to my inline comments I'd add that:

- I'd suggest renaming the tab to `Advanced` and for this tab to be the last one.
- I'd suggest adding text above the options input to indicate how to use it, what data it accepts (an example could help).

I've tried to run the explorer but the tabs don't show. Ah I realized just know this PR is not targeting main but another PR, something is broken.

image

hvplot/ui.py
```diff
@@ -709,6 +732,10 @@ def settings(self):
         settings = {k: v for k, v in sorted(list(settings.items()))}
         return settings
 
+    def opts(self):
```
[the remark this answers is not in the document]
ahuang11: For parallelism with settings(), but I suppose it's not necessary

hvplot/tests/testui.py
ahuang11: Unintended; fixed now.

hvplot/ui.py
ahuang11: It's less robust, and not all opts are available.

ahuang11:
To address the following:
> I'd suggest renaming the tab to Advanced and for this tab to be the last one.
> I'd suggest adding text above the options input to indicate how to use it, what data it accepts (an example could help).

Renamed to advanced and add a doc tooltip. I didn't like how adding text above looked.

image

hvplot/ui.py
```diff
@@ -362,6 +362,19 @@ def _update_options(self):
...
 
 
+class Advanced(Controls):
+
+    opts = param.Dict(
+        default={}, doc="""
+        Additional styling options to include.
+        Examples:
+        - Image: {"color_levels": 11}
+        - Line: {"line_style": "dashed"}
```
maximlt, line 372: `line_style` isn't a valid option.

ahuang11, line 372:
```suggestion
        - Line: {"line_dash": "dashed"}
```

ahuang11:
To me, opts is essential to get the final plot desired. Here's another example that I don't think hvplot alone can do:

```python
import hvplot.xarray
import xarray as xr

df = xr.tutorial.open_dataset("air_temperature")
df.hvplot("lon", "lat").opts(colorbar_opts={"title": "Air Temperature", "width": 10, "bar_line_alpha": 0})
```

PR #1135, opened 2023-09-14
Improve x-limit, y-limit, and rasterize support for geo features 

Hoxbro:
Fixes #1093 
Fixes #1090
Fixes  https://discourse.holoviz.org/t/any-smart-approach-for-determining-what-holoviz-package-broke-my-notebook/5944 

If geo is true, we will use GeoViews tile sources, and if false, we will use HoloViews. GeoViews features and tiles are projected to the respective projection. 

---
1093 code:
``` python
import cartopy.crs as ccrs
import geoviews as gv
import hvplot.xarray  # noqa
import numpy as np
import xarray as xr


def sample_data(shape=(20, 30)):
    """
    Return ``(x, y, u, v, crs)`` of some vector data
    computed mathematically. The returned crs will be a rotated
    pole CRS, meaning that the vectors will be unevenly spaced in
    regular PlateCarree space.

    """
    crs = ccrs.RotatedPole(pole_longitude=177.5, pole_latitude=37.5)

    x = np.linspace(311.9, 391.1, shape[1])
    y = np.linspace(-23.6, 24.8, shape[0])

    x2d, y2d = np.meshgrid(x, y)
    u = 10 * (2 * np.cos(2 * np.deg2rad(x2d) + 3 * np.deg2rad(y2d + 30)) ** 2)
    v = 20 * np.cos(6 * np.deg2rad(x2d))

    return x, y, u, v, crs


xs, ys, U, V, crs = sample_data()

mag = np.sqrt(U**2 + V**2)
angle = (np.pi / 2.0) - np.arctan2(U / mag, V / mag)

ds = xr.Dataset(
    {
        "mag": xr.DataArray(mag, dims=("y", "x"), coords={"y": ys, "x": xs}),
        "angle": xr.DataArray(angle, dims=("y", "x"), coords={"y": ys, "x": xs}),
    },
    attrs={"crs": crs},
)
ds.hvplot.vectorfield(
    x="x",
    y="y",
    angle="angle",
    mag="mag",
    hover=False,
    geo=True,
    tiles="CartoLight",
)
```
image

1090 code:
``` python
import hvplot.pandas
import pandas as pd

df = pd.DataFrame({"lon": [0, 10], "lat": [40, 50], "v": [0, 1]})

(
    df.hvplot.points(x="lon", y="lat", s=200, c="y", geo=True, tiles="CartoLight")
    * df.hvplot.points(x="lon", y="lat", c="v", geo=True)
)
```
image

---
[...]

PR #1163, opened 2023-10-04
Revert geo + rasterize changes introduced in 1135 

### LINKED ISSUE #1162: Rasterized and projected RGB plot very slow or hanging
Running the code from https://github.com/holoviz/hvplot/issues/1104 on the main branch is either very very slow or hanging on my laptop:

```python
import hvplot.xarray
import rioxarray as rxr

# Downloaded from: https://oin-hotosm.s3.amazonaws.com/5d7dad0becaf880008a9bc88/0/5d7dad0becaf880008a9bc89.tif
cog_url = "/home/shh/Downloads/5d7dad0becaf880008a9bc89.tif"
da = rxr.open_rasterio(cog_url)
plot = da.hvplot.rgb(x="x", y="y", rasterize=True, projection="Robinson")
plot
```

Hoxbro:
Fixes #1162

The changes made in #1135 were already possible with `project=True`.


Hoxbro:
Okay, reverting most of the changes in #1135 related to `gv.project`, as they were already possible with the setting `project=True`.

Hoxbro:
@ahuang11, can you take a look at this PR? If you see weird behavior with geo + rasterize stuff, try to enable `project=True`.

hvplot/tests/testgeo.py
```diff
@@ -140,6 +140,7 @@ def test_geo_with_rasterize(self):
             rasterize=True,
             dynamic=False,
             aggregator="max",
+            project=True,
```
ahuang11, line 143: This looks good to me; however points is back to being broken without `project=True` which to me seems undesirable.

Can we keep the last PR's changes but with points?:
```python
        if self.geo and self.crs != self.output_projection and self.kind == "points":
            import geoviews as gv
            obj = gv.project(obj, projection=self.output_projection)
```

ahuang11: Actually maybe putting it under the existing clause:
```python
        if self.crs and self.project or (self.kind == "points" and self.rasterize):
            # Apply projection before rasterizing
            import geoviews as gv
            obj = gv.project(obj, projection=self.output_projection)
```

Hoxbro: I would rather not. There is an extra cost associated with `gv.project`. It seems points work for the simple case, but we can't guarantee it will work for all cases. 

The documentation [here](https://hvplot.holoviz.org/user_guide/Geographic_Data.html#declaring-an-output-projection) could have some an [admonition](https://sphinx-book-theme.readthedocs.io/en/stable/reference/kitchen-sink/admonitions.html) that say try `project=True` if you see problem with projection and rasterize.

ahuang11: > It seems points work for the simple case, but we can't guarantee it will work for all cases.

Do you mean if the points X*Y is large, it won't work?

Hoxbro: Or it will slow it significantly down with no way to disable it.

ahuang11: Just tested:
We can test it with this:

With rasterize=True (no project), it takes 24s; with project, 40s
```python
import hvplot.xarray
import rioxarray as rxr

da = rxr.open_rasterio("https://oin-hotosm.s3.amazonaws.com/5d7dad0becaf880008a9bc88/0/5d7dad0becaf880008a9bc89.tif")
da.isel(band=0).hvplot.points(x="x", y="y", projection="Robinson")
```

I'll let you make the final call~

ahuang11 (approved):

PR #1164, opened 2023-10-05
Param 2 and HoloViews 1.18 compatibility

Hoxbro:


hvplot/util.py
```diff
@@ -24,6 +24,7 @@
 hv_version = Version(hv.__version__)
 bokeh_version = Version(bokeh.__version__)
 bokeh3 = bokeh_version >= Version("3.0")
+param2 = Version(param.__version__) >= Version("2.0rc4")
```
Hoxbro, line 27: Just wanted a rc version high enough.  

rc should be removed at some point

Hoxbro:
Matplotlib pin should be removed when https://github.com/holoviz/holoviews/pull/5924 and https://github.com/holoviz/holoviews/pull/5910 are merged, and a dev release it out. 

They will be in the next Holoviews release.

Hoxbro:
Need to dive deeper into where the `dict` problem is located. For this HoloViews dev release, I ignore the failing notebooks, which is not ideal.

PR #1181, opened 2023-10-16
Fix displaying widgets controlled DynamicMaps in the explorer

maximlt:
See https://github.com/holoviz/panel/issues/5628#issuecomment-1763443895


philippjfr (approved):
Looks good, this will always continue working.

PR #1186, opened 2023-10-28
fix typo in the docstring and an explanation for `interactive.panel()` and `interactive.widgets()`

iuryt:
This is not major and should be straightforward.
I hope it is fine to create a PR to such a simple fix.

The explanation in the tutorial originates from this discussion.
https://discourse.holoviz.org/t/why-is-the-multiselect-rendered-twice-and-the-chart-only-once/6286/3

I noticed the typo while looking for extra arguments in `interactive()`

### DIFF
--- a/examples/user_guide/Interactive.ipynb
+++ b/examples/user_guide/Interactive.ipynb
@@ -405,7 +405,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


maximlt:
@iuryt of course this is fine, any kind of contribution is more than welcome!!! Thanks!

PR #1195, opened 2023-11-09
 Rename amin and amax to min and max 

Hoxbro:
The plotting tests began to fail yesterday: https://github.com/holoviz/hvplot/actions/runs/6802887790

This is likely because a new numba version was released on conda-forge, which also updated the numpy version. (red=failing test, green=passing test)

image

@maximlt, you can change the code to your liking. This is mostly to get the ball rolling.

### DIFF
--- a/examples/conftest.py
+++ b/examples/conftest.py
@@ -1,13 +1,1 @@
-import numpy as np
-
-from packaging.version import Version
-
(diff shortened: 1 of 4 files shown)


Hoxbro:
LGTM

Issue #1190, opened 2023-11-02
hvPlot Explorer issues


ahuang11:
These things make the explorer unintuitive to use:
- [ ] Widget_location -> holoviews pane does not support column
- [ ] Two Time(s)
- [ ] By tooltip
- [ ] Null value
- [ ] Groupby should be None instead
- [ ] Infer color label from dataset
- [ ] Rot better label
- [ ] Frame wise / axis wise
- [ ] Height / width doesn't do anything
- [ ] Remove datashader count for gridded
- [ ] Red message until its corrected
- [ ] Clim could be renamed to Colorbar limits
- [ ] Color disabled for gridded
- [ ] Cartopy needs installed or else everything breaks silently
- [ ] Tiles activated without geo
- [ ] Bokeh invalid tooltip

Credits to @jbednar for discovering these.

jbednar:
For reference, I was able to demonstrate each of these issues using this code on a machine without Cartopy installed (i.e., an  environment with hvPlot but without GeoViews):

```python
import hvplot.xarray, xarray as xr, holoviews as hv
hv.extension("bokeh", width=100)

ds = xr.tutorial.open_dataset('air_temperature').load()

hvplot.explorer(ds, x="lon", y="lat")
```
<img width="1161" alt="image" src="https://github.com/holoviz/hvplot/assets/1695496/2ee0720b-bbfe-4807-b643-82272f27e421">

Notice the small size of the plot relative to the page width, which is why the horizontal placement of the HoloMap/DynamicMap widgets is a problem. For different zoom levels and browser widths the issue gets much worse; I think we really need to be reserving much more of the screen real estate for the plot:

<img width="897" alt="image" src="https://github.com/holoviz/hvplot/assets/1695496/dabc92a7-b880-4f9b-bb05-9e3ba72bd4e1">

maximlt:
General comment on naming things, I definitely want the name of the parameters passed to `hvplot.<method>` appear in the explorer as I strongly believe the explorer is a good way to discover and learn hvPlot's (and Pandas' to some extent) API. It's for example the best way to understand the difference between `by` and `groupby`, just by playing around with the options. So their actual names must be shown in the UI.

>  Two time(s) in the fields list: 'time' and "['time']"

I've seen this issue with a dev release of Param, just make sure to reproduce it with Param 2.0.

jbednar:
> So their actual names must be shown in the UI.

[...]

PR #1197, opened 2023-11-11
Fix minor issues in explorer

ahuang11:
Tweaks to labels, placeholders, margins, layout of explorer, based on the minor issues listed here:
https://github.com/holoviz/hvplot/issues/1190

image

### DIFF
--- a/hvplot/ui.py
+++ b/hvplot/ui.py
@@ -110,9 +110,14 @@ class Controls(Viewer):
 
 class Colormapping(Controls):
 
-    clim = param.Range()
+    clim = param.Range(label="Colorbar Limits (clim)", doc="""
+        Upper and lower limits of the colorbar.""")
 
-    cnorm = param.Selector(default='linear', objects=['linear', 'log', 'eq_hist'])
+    cnorm = param.Selector(
+        label="Colorbar Normalization (cnorm)",
+        default='linear',
+        objects=['linear', 'log', 'eq_hist']
+    )
 
     color = param.String(default=None)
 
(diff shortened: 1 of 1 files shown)


maximlt:
Why has the status bar been moved to the top? I'm much more used to having a status bar at the bottom, like in the Perspective viewer or in QGIS. That's where it usually is apparently:
image

ahuang11:
I personally do not think the `Live Update` should be called "Status Bar"

Also, if it's on the bottom, the plot will push it out of view.

image

maximlt:
Yes probably status bar isn't the right word for that, although I've seen status bar used for components that do provide some control (like in [QGIS](https://docs.qgis.org/3.28/en/docs/user_manual/introduction/qgis_gui.html#status-bar)).

It's easy to see though that we can add more of this kind of tools (reset, export, ...) and indicators (mouse position?), and I guess we should try to have them fit all in the same horizontal bar, so pick between top or bottom location. I think I prefer at the bottom as that's not the first thing users should see and click on, however I'm no UI/UX professional so I may just be plain wrong.

> Also, if it's on the bottom, the plot will push it out of view

That's certainly the crux of the problem, in that we don't have very good control over the plot height, and even the Tab layout height that right now is as tall as its tallest item. There's a bit of a tension between allowing the users to set the plot dimensions (aka the explorer as a plot builder/editor/wizard) and giving them the best plot for them to explore their data (aka the explorer as the explorer!).

maximlt:
Ok at the top is fine.

ahuang11:
Can we merge this then or do you need more time to review it?

maximlt:
I'll review it later.

maximlt:
LGTM!

I still think the status/control bar is a bit in a weird place and looks to big for what it is, I don't want to block to PR for too long though as others depend on it and there may be conflicts. Hopefully one day we'll be supported by UI/UX engineers for this sort of design work!

image

PR #1198, opened 2023-11-12
Limit explorer widget width

ahuang11:
Condenses the widget box so there's more space for the plot. Will address the time slider in another PR.

Builds on top of https://github.com/holoviz/hvplot/pull/1197 and partially addresses https://github.com/holoviz/hvplot/issues/1190

Changes based on discovery in https://github.com/holoviz/panel/issues/5856

image

### DIFF
--- a/hvplot/ui.py
+++ b/hvplot/ui.py
@@ -467,7 +467,7 @@ class hvPlotExplorer(Viewer):
             self.statusbar,
             show_name=False,
             default_layout=pn.Row,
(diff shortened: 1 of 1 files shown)


maximlt (commented):
I've refactored this slightly to leverage `__panel__` more as the approach taken was by-passing it, by-passing the auto-throttling of widgets created from Range/Number parameters.

ahuang11:
Beautiful solution!

PR #1199, opened 2023-11-12
Move plot widget to bottom

ahuang11:
Builds on top of https://github.com/holoviz/hvplot/pull/1198 for more screen space for the plot and partially addresses https://github.com/holoviz/hvplot/issues/1190

image

### DIFF
--- a/hvplot/ui.py
+++ b/hvplot/ui.py
@@ -662,10 +662,7 @@ class hvPlotExplorer(Viewer):
                 ('Advanced', self.advanced),
             ]
             if event and event.new not in ('area', 'kde', 'line', 'ohlc', 'rgb', 'step'):
-                colormapping_params = self._create_param(self.colormapping, widgets={
-                    "clim": {"placeholder": "(min, max)"},
-                })
-                tabs.insert(5, ('Colormapping', colormapping_params))
+                tabs.insert(5, ('Colormapping', self.colormapping))
         self._control_tabs[:] = tabs
 
     def _check_y(self, event):
--- a/hvplot/ui.py
+++ b/hvplot/ui.py
@@ -631,12 +631,6 @@ class hvPlotExplorer(Viewer):
     def _groups(self):
         raise NotImplementedError('Must be implemented by subclasses.')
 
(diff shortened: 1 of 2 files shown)


maximlt (commented):
I don't exactly know why but it doesn't fit very well in a notebook output cell. Do you interact with the explorer always as an app? I'd guess the explorer is more used in a notebook context.

image

When served as an app the plot occupies the whole space and the widget box looks a little alone centered down there.

image

What alternative do we have? We could maybe dynamically create a new tab populated with the widget box? Or add it below the groupby multiselect?

ahuang11:
A floating pane or built into Bokeh's toolbar if ever possible

ahuang11:
Oh I suppose we could do `widget_layout = pn.Row` instead of `pn.Column`.

ahuang11:
image

ahuang11:
Another thing we need to fix is the default `legend_loc`.

hvplot/ui.py
```diff
@@ -528,9 +531,10 @@ def __init__(self, df, **params):
                 # Using .layout on the HoloViews pane to display the widgets
                 # https://github.com/holoviz/panel/issues/5628#issuecomment-1763443895
                 pn.Tabs(('Plot', self._hv_pane.layout), ('Code', self._code_pane)),
-                sizing_mode='stretch_width',
+                sizing_mode='stretch_both'
             ),
-            sizing_mode='stretch_both'
+            sizing_mode='stretch_width',
+            height=600
```
ahuang11, line 537: `min_height` unfortunately did not work for me.

maximlt:
> image

The widgets are not well vertically aligned, I don't think we can merge with this issue, let's see if I can fix it.

maximlt:
Came up with a workaround:
image

maximlt (approved):
The explorer will look much better with this change when there are multiple widgets!

ahuang11:
Shall we merge?

maximlt:
Yes, I was waiting for the tests to pass.

PR #1200, opened 2023-11-12
Explorer notify on any exceptions

ahuang11:
Continues addressing https://github.com/holoviz/hvplot/issues/1190

image

### DIFF
--- a/hvplot/ui.py
+++ b/hvplot/ui.py
@@ -1,3 +1,5 @@
+from functools import partial
+
 import holoviews as _hv
 import numpy as np
 import panel as pn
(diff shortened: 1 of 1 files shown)


maximlt (commented):
What is this PR trying to fix from the linked issue? This maybe?

> When showing an error, mention that the plot will not update properly until the error message is corrected

Generally I think the explorer can be embedded into other apps, in which it shouldn't override the exception handler if it's already set.

The `exception_handler` config was added in Panel 0.14 but the min pin for hvPlot currently is 0.11. It's probably fine to bump the pin to 0.14, although I'd do it in another PR to keep track of that change in the release notes.

ahuang11:
This one:
> Cartopy needs installed or else everything breaks silently; need to catch all such errors to avoid locking UI

ahuang11:
image
```python
import hvplot.xarray
import xarray as xr
import panel as pn
pn.extension()
pn.config.exception_handler = lambda exc: print(f"Try this: {exc}")

ds = xr.tutorial.open_dataset("air_temperature")
ds.hvplot.explorer("lon", "lat")
```
image

maximlt:
I'm still unsure that's the right approach, the explorer can wrongly display external exceptions:
image

ahuang11:
How do you suggest I approach it? Write a try/except decorator and wrap every method?

maximlt (changes requested):
Coming back to this PR after some weeks, I'm unsure about the validity of the goal here. I don't consider the explorer being so much different from any other Panel component, like a widget, and I wouldn't like internal widget errors to be displayed in the UI of my apps, at least certainly not by default. Internal errors should be raised so users can report the traceback in an issue.

Could this PR focus only on fixing the cartopy issue? If we want to handle all internal errors in the same way, we should discuss it in an issue.

ahuang11:
Closing this as no longer necessary.

PR #1201, opened 2023-11-12
Add None placeholder

ahuang11:
Continues addressing https://github.com/holoviz/hvplot/issues/1190
image

### DIFF
--- a/hvplot/ui.py
+++ b/hvplot/ui.py
@@ -13,7 +13,7 @@ from .plotting import hvPlot as _hvPlot
 from .util import is_geodataframe, is_xarray, instantiate_crs_str
 
 # Defaults
-NONE_PLACEHOLDER = "<null>"
+NONE_PLACEHOLDER = "<no selection>"
 
 KINDS = {
     # these are for the kind selector
--- a/hvplot/ui.py
+++ b/hvplot/ui.py
@@ -13,6 +13,8 @@ from .plotting import hvPlot as _hvPlot
 from .util import is_geodataframe, is_xarray, instantiate_crs_str
 
 # Defaults
+NONE_PLACEHOLDER = "<null>"
+
 KINDS = {
     # these are for the kind selector
     'dataframe': sorted(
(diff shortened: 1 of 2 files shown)


Hoxbro:
Maybe rename `null` to something else. `null` can be confusing to Python-only developers.

ahuang11:
`<None>`?

Hoxbro:
>  I kind of like 

Don't keep me in suspense :upside_down_face:

ahuang11:
Haha, apparently `<no selection>` is hidden without the backticks!

hvplot/ui.py
```diff
@@ -828,6 +847,7 @@ def _populate(self):
                 elif pname == 'groupby' and len(getattr(self, pname, [])) == 0 and len(p.objects) > 2:
                     setattr(self, pname, p.objects[2:])
 
+                p.objects = [NONE_PLACEHOLDER] + p.objects
```
ahuang11, line 850:
```suggestion                if pname not in ('x', 'y'):
                    p.objects = [NONE_PLACEHOLDER] + p.objects
```

maximlt:
I see the issue says:

> Null value isn't visible in a Multiselect widget, so once a value has been selected, it's not intuitive how to go back to unselected (which is Command-click on Mac)

But is this really true? A select HTML element with multiple options is a very basic element that is used, everywhere?

ahuang11:
I think it's nice to have a <null selection> or empty slot; it's probably more straightforward for the user.

@jbednar was the one who raised this issue so he might be able to provide additional feedback.

maximlt:
I very much like the approach you took when developing the Chat components in Panel, drawing heavily on ChatGPT's design and UX. Let's do the same thing with the explorer? We can pick a champion alternative, or a couple, and gradually adapt the explorer's design to follow theirs. What do you think? Without someone with a UX hat, I find difficult to come to a conclusion on this kind of change.

@jbednar if you think that the MultiSelect widget needs a _null_ option, I suggest a feature request should be opened on Bokeh, as there's nothing specific to the explorer here.

Generally, we need better multi select/choice widgets, this has been in the back of my mind for a while, I'll have to scratch that itch at some point!

philippjfr:
I dislike a null option on a `MultiSelect` widget quite a bit. It means you can have the nonsensical combination of the null value and other values, which is plain weird. The real problem is providing a clear UI for clearing a MultiSelect widget which should indeed happen either in Bokeh or Panel.

philippjfr:
Okay, reading the implementation I see now that you prevent the nonsensical combinations, but having to do this in Python really should not be necessary and is still weird UX. Better multi-select in Bokeh/Panel is the correct way forward, but if you want to merge this as a stopgap I won't object.

ahuang11:
I also vote to add it to Panel/Bokeh, but I think we would need to understand what a better multi-select means--perhaps finding a website that does it really well as Maxime mentioned.

philippjfr:
>To me this is extremely unintuitive behavior, getting the user stuck and unable to back out. I would argue that forcing users to figure out that (on a Mac) they can hold down fn while clicking to unselect is bizarre and hostile UX. I have no idea what's at fault (the browser, Bokeh, Panel, hvPlot??).

Standard behavior of the `<select multiple>` tag: https://www.w3schools.com/tags/tryit.asp?filename=tryhtml_select_multiple

jbednar:
Still hostile UX! :-)

maximlt:
The UX of the standard multi-select HTML element might be considered bad (I don't think so, YMMV), but I don't think adding another behavior on top of it improves it.

---

There are all sorts of flavors of multi-select widgets out there. There's already a feature request opened on Panel for one of them inspired by Excel https://github.com/holoviz/panel/issues/3018.

jbednar:
> The UX of the standard multi-select HTML element might be considered bad (I don't think so, YMMV), but I don't think adding another behavior on top of it improves it.

Hmm. Going back to it, I see that I'm not even able to detect that it's a multi-select at all in casual usage! Totally hostile UX; how would someone be expected to realize that? Clicking around, it only seems like one item can be selected, and for some reason, after selecting anything, there's no obvious way to unselect it. Yet the actual affordance of this widget is not only to allow multiple selections, it allows no selection too, but only if you discover the magic modifier key. Ugh!!! I hate it. So never mind; it's beyond saving.

ahuang11:
Closing this until we find a better resolution.

PR #1203, opened 2023-11-14
Toggle explorer selections

ahuang11:
Builds on https://github.com/holoviz/hvplot/pull/1201 and only shows valid selections.

image


maximlt:
Is this PR in draft mode because it's blocked by https://github.com/holoviz/hvplot/pull/1201 or because it needs more work?

ahuang11:
I think it's still buggy at the current state

Issue #1214, opened 2023-12-12
hvplot.ohlc() doesn't respect the hover_cols param


vanheck:
Hi, when use `hvplot.ohlc()` with user defined hover, the additional columns specified in `hover_cols` params doesn't appear in the result ColumnDataSource.

Source of this problem in stackoverflow: https://stackoverflow.com/questions/77631385/how-to-add-data-into-hover-in-hvplot-bokeh-candlestick-without-missing-date-ga

#### ALL software version info
python 3.12
bokeh 3.3.2
hvplot 0.9.0
holoviews 1.18.1

#### Complete, minimal, self-contained example code that reproduces the issue

```py
import pandas as pd
import hvplot.pandas

data = pd.DataFrame({
        "Open": [100.00, 101.25, 102.75],
        "High": [104.10, 105.50, 110.00],
        "Low": [94.00, 97.10, 99.20],
        "Close": [101.15, 99.70, 109.50],
        "Volume": [10012, 5000, 18000],
    }, index=[pd.Timestamp("2022-08-01"), pd.Timestamp("2022-08-03"), pd.Timestamp("2022-08-04")])

df = pd.DataFrame(data)

# fix hover ------
from bokeh.models import HoverTool

hover = HoverTool(
    tooltips=[
        ('Date', '@Date{%Y-%m-%d}'),
        ('Open', '@Open{0.00}'),
        ('High', '@High{0.00}'),
        ('Low', '@Low{0.00}'),
        ('Close', '@Close{0.00}'),
        ('Volume', '@Volume{0}'),
    ],
    formatters={'@Date': 'datetime'},
    mode='vline'
)
# fix hover ------

ohlc_cols = ["Open", "High", "Low", "Close"]
df.hvplot.ohlc(y=ohlc_cols, hover_cols=["Date", *ohlc_cols, "Volume"], tools=[hover])
```

#### Screenshot
![obrazek](https://github.com/holoviz/hvplot/assets/9953249/fc98717f-4401-489e-a99b-c8d5f2ca5e56)

maximlt:
Fixed by https://github.com/holoviz/hvplot/pull/1216

PR #1216, opened 2023-12-13
Respect hover_cols for OHLC

maximlt:
Addresses https://github.com/holoviz/hvplot/issues/1214

```python
import pandas as pd
import hvplot.pandas

df = pd.DataFrame({
        "Open": [100.00, 101.25, 102.75],
        "High": [104.10, 105.50, 110.00],
        "Low": [94.00, 97.10, 99.20],
        "Close": [101.15, 99.70, 109.50],
        "Volume": [10012, 5000, 18000],
    }, index=[pd.Timestamp("2022-08-01"), pd.Timestamp("2022-08-03"), pd.Timestamp("2022-08-04")])

ohlc_cols = ["Open", "High", "Low", "Close"]
df.hvplot.ohlc(y=ohlc_cols, hover_cols=["Volume"])
```

image

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -2060,7 +2060,7 @@ class HoloViewsConverter:
             o, h, l, c = y
         neg, pos = self.kwds.get('neg_color', 'red'), self.kwds.get('pos_color', 'green')
         color_exp = (dim(o)>dim(c)).categorize({True: neg, False: pos})
-        ds = Dataset(data, [x], [o, h, l, c])
+        ds = Dataset(data, [x], [o, h, l, c] + self.hover_cols)
         if ds.data[x].dtype.kind in 'SUO':
             rects = Rectangles(ds, [x, o, x, c])
         else:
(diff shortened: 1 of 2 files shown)


hvplot/converter.py
```diff
@@ -2060,7 +2060,7 @@ def ohlc(self, x=None, y=None, data=None):
             o, h, l, c = y
         neg, pos = self.kwds.get('neg_color', 'red'), self.kwds.get('pos_color', 'green')
         color_exp = (dim(o)>dim(c)).categorize({True: neg, False: pos})
-        ds = Dataset(data, [x], [o, h, l, c])
+        ds = Dataset(data, [x], [o, h, l, c] + self.hover_cols)
```
Hoxbro, line 2063: Have you checked if this works when `hover_cols="all"`?

hvplot/converter.py
```diff
@@ -2060,7 +2060,11 @@ def ohlc(self, x=None, y=None, data=None):
             o, h, l, c = y
         neg, pos = self.kwds.get('neg_color', 'red'), self.kwds.get('pos_color', 'green')
         color_exp = (dim(o)>dim(c)).categorize({True: neg, False: pos})
-        ds = Dataset(data, [x], [o, h, l, c] + self.hover_cols)
+        ohlc_cols = [o, h, l, c]
+        if x in self.hover_cols:
+            self.hover_cols.remove(x)
```
Hoxbro, line 2065: Can this problem if `x` is `a` or `l` and `self.hover_cols` is all?

maximlt: Why would `x` be equal to `l`?

Hoxbro: Sorry. I just took all the letters of `all` and did not see that it was already part of the `ohlc` columns.

PR #1222, opened 2023-12-21
Polars in hvplot.plot: avoid registering the hvplot accessor and simple test

maximlt:
Follow-up of  to add a simple test and to avoid registering the hvplot accessor, as I don't think this was intended.

### DIFF
--- a/hvplot/tests/testplotting.py
+++ b/hvplot/tests/testplotting.py
@@ -1,13 +1,23 @@
 """
 Tests pandas.options.backend setting
 """
-from packaging.version import Version
 from unittest import TestCase, SkipTest
-import pandas as pd
 
+import holoviews as hv
+import pandas as pd
+import pytest
+
+from packaging.version import Version
 from parameterized import parameterized
 
-from ..converter import HoloViewsConverter
+from hvplot.converter import HoloViewsConverter
+from hvplot.plotting import plot
+
+try:
+    import polars as pl
+    skip_polar = False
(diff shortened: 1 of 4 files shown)


hvplot/tests/testplotting.py
```diff
@@ -58,3 +68,10 @@ def setUp(self):
             raise SkipTest('entrypoints for plotting.backends was added '
                            'in pandas 0.25.1')
         pd.options.plotting.backend = 'hvplot'
+
+
+@pytest.mark.skipif(skip_polar, reason="polars not installed")
+def test_plot_supports_polars():
+    dfp = pl.DataFrame(pd._testing.makeDataFrame())
```
Hoxbro, line 75:
```suggestion
def test_plot_supports_polars():
    pl = pytest.importorskip("polars")
    dfp = pl.DataFrame(pd._testing.makeDataFrame())
```

Alternative way (also needs import removal also)

Hoxbro (approved):
A tiny comment, but nothing that blocks this PR.

hvplot/tests/testplotting.py
maximlt: Made the change, even if I'd say that having the marker above the test definition helps distinguish tests when looking at many of them, while an inline skip is easy to miss.

Hoxbro: Yep, I wouldn't have had a problem if you kept what you had before.

PR #1223, opened 2023-12-21
Explorer: disable the geo controls when geoviews is not installed

maximlt:
Addressing (a tiny) part of https://github.com/holoviz/hvplot/issues/1190 with a different take compared to https://github.com/holoviz/hvplot/pull/1200 whose goal is to cover all kind of exceptions the explorer might raise, while in this PR I focused on finding the best way to let them know they should install GeoViews to enable the geo controls.

I ended up trying to import geoviews on initialization and if that fails disabling all the geo controls:

image

The `Checkbox` widget has no tooltip so instead I used a Toggle with a label that is updated to indicate GeoViews is required.
Trying to instantiate an explorer by passing geo kwargs (e.g. `df.explorer(geo=True)`) will raise an import error.

### DIFF
--- a/hvplot/ui.py
+++ b/hvplot/ui.py
@@ -350,14 +350,17 @@ class Geographic(Controls):
 
         self.param.crs.objects = crs_list
         self.param.projection.objects = crs_list
+        updates = {}
         if self.projection is None:
-            self.projection = crs_list[0]
+            updates['projection'] = crs_list[0]
 
         if self.global_extent is None:
-            self.global_extent = True
+            updates['global_extent'] = True
 
         if self.features is None:
-            self.features = ['coastline']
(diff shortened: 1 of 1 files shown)


ahuang11 (approved):
Looks good to me! The only thing is tiles doesn't seem to work without geographic projections, but I think that's a separate issue.

maximlt:
@ahuang11 Isn't it that you need to provide a dataset that has coordinates already projected to easting/northing to use `tiles`? Not all geo datasets are in lat/lon so I'm not sure the explorer should perform the conversion by default.

ahuang11:
Maybe an extra toggle in the future that allows conversion to easting northing?

maximlt:
Yes I think that'd make sense.

PR #1224, opened 2023-12-21
Setup GoatCounter and remove Google Analytics

maximlt:
Exactly what the title says :)


maximlt:
It works, merging:
image

PR #1235, opened 2023-12-22
Dynamic generate unit test matrix and add cache option

Hoxbro:
See https://github.com/holoviz/holoviews/pull/6043 for more details.

image

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -56,11 +56,7 @@ jobs:
         run: |
           MATRIX=$(jq -nsc '{
               "os": ["ubuntu-latest", "macos-latest", "windows-latest"],
-              "python-version": ["3.8", "3.11"],
-              "include": [
-                  {"os": "ubuntu-latest", "python-version": "3.9"},
-                  {"os": "ubuntu-latest", "python-version": "3.10"}
-              ]
+              "python-version": ["3.8", "3.11"]
           }')
           echo "MATRIX=$MATRIX" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -20,16 +33,60 @@ jobs:
...
+          echo "MATRIX_OPTION=$OPTION" >> $GITHUB_ENV
+      - name: Set test matrix with 'default' option
+        if: env.MATRIX_OPTION == 'default'
+        run: |
+          MATRIX=$(jq -nsc '{
+              "os": ["ubuntu-latest", "macos-latest", "windows-latest"],
+              "python-version": ["3.8", "3.11"],
+              "include": [
+                  {"os": "ubuntu-latest", "python-version": "3.9"},
+                  {"os": "ubuntu-latest", "python-version": "3.10"}
```
Hoxbro, line 62: I have added these two extra tests by default. You can remove it if you don't like it.

maximlt: Just wondering why you've added them?

Hoxbro: I added them to check how different versions of Python packages work. 

An example is that Python 3.8 only works up to Bokeh 3.1 and, therefore, not the latest Panel version. 

The next couple of releases for packages that follow [SPEC 0000](https://scientific-python.org/specs/spec-0000/) could drop Python 3.9 support.

maximlt: Ok I've removed them but it's good to know we can easily extend the matrix this way.

PR #1238, opened 2023-12-24
Show categorical legend with hvplot's `by` keyword and datashader.

JelmerBot:
Currently using hvplot with datashader and the `by` keyword does not show a legend. This PR makes the `by` keyword behave as if only a categorical aggregator was specified when datashader is used, ensuring a legend is created.

A small example where the change is effective:

```python
import numpy as np
import pandas as pd
import hvplot
import hvplot.pandas
import datashader as ds

df = pd.DataFrame(
    np.random.normal(size=(20, 2)),
    columns=['x', 'y']
)
df['cat'] = df.x.apply(lambda x: 'c0' if x < 0 else 'c1')
df['cat'] = df.cat.astype('category')

df.hvplot.points(
    x='x',
    y='y',
    by='cat',
    # aggregator=ds.by('cat'),
    color_key={'c0': '#ff0000', 'c1': '#0000ff'},
    datashade=True,
    dynspread=True,
)
```

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -2335,7 +2335,7 @@ class HoloViewsConverter:
             if 'tools' in opts_ and kind in ["polygons", "paths"] and not vdims:
                 opts_["tools"] = [t for t in opts_["tools"] if t != "hover"]
         if self.geo: params['crs'] = self.crs
-        if self.by:
+        if self.by and not self.datashade:
             obj = Dataset(data, self.by+kdims, vdims).to(element, kdims, vdims, self.by, **params)
             if self.subplots:
                 obj = obj.layout(sort=False)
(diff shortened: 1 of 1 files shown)


maximlt:
Hi @JelmerBot, sorry I took so long to come back to you! And thanks for opening a PR :) I checked the effects of your changes on the code you shared and the results look good at first glance.

Before:
image

With the changes as of https://github.com/holoviz/hvplot/pull/1238/commits/b098ea3e96e5fb75e2d22711c318f427c46ccaea:
image

However, I noticed that setting `subplots=True` no longer works after the changes:
image

I need to spend a bit more time trying to understand the various things at stake here (likely related to more or less recent developments in HoloViews too). Once it's clearer in my mind, I'll offer you to keep working on the PR, for instance it will require some unit tests.

PR #1243, opened 2023-12-30
add downstream polars test

MarcoGorelli:
Hey - just wanted to suggest this as a "sister PR" to https://github.com/pola-rs/polars/pull/13238

Then, we could be confident that the "Polars 🤝 hvPlot" compatibility stays stable?


maximlt:
> Then, we could be confident that the "Polars 🤝 hvPlot" compatibility stays stable?

We'll 100% make sure of that!

PR #1247, opened 2024-01-02
correctly fall back to all numeric polars cols when y is not specified

### LINKED ISSUE #1246: y=None doesn't work for Polars dataframes
```
>>> import hvplot
>>> hvplot.__version__
'0.9.1'
>>> import holoviews
>>> holoviews.__version__
'1.18.1'
>>> import polars
>>> polars.__version__
'0.20.3'
```

If you call, say `df.plot.line(x="foo")`, the docs say that all other numeric cols should be used as y values, but instead we get no plot and the following message: `matplotlib backend could not plot any Elements in the Overlay.`

```python
import polars as pl
pl.DataFrame({"a": [1, 2], "b": [3, 4]}).plot.line(x="a")
```

In the [_get_converter](https://github.com/holoviz/hvplot/blob/734c0cc922152ea6a0f3332bc2bf19f5430d7629/hvplot/plotting/core.py#L1867) method of hvPlotTabularPolars, we see that (if hover_cols is not set to "all") only columns from `x`, `y` and `kwds` will be selected:
```python
                possible_columns = [
                    [v] if isinstance(v, str) else v
                    for v in params.values()
                    if isinstance(v, (str, list))
                ]
                columns = (
                    set(self._data.columns) & set(itertools.chain(*possible_columns))
                ) or {self._data.columns[0]}
                xs = x if is_list_like(x) else (x,)
                ys = y if is_list_like(y) else (y,)
                columns |= {*xs, *ys}
                columns.discard(None)
```

### Proposed solution

When the y value is not specified, select all numeric columns (plus whatever is specified) in _get_converter. We can do this in Polars like this:

```python
import polars as pl
pl.DataFrame({"a": [1, 2], "b": [3, 4], "c": ["foo", "bar"], "d": ["ping", "pong"]}).select([pl.col("d"), pl.col(pl.NUMERIC_DTYPES)])
```
This code will select columns d, a and b.

There is still some filtering based on pandas dtype later in the code but I think this is applied to non-Polars sources too, so we may as well just leave it as it is for now.

- [x] I may be interested in making a pull request to address this

kevinheavey:
Fixes #1246

### DIFF
--- a/hvplot/plotting/core.py
+++ b/hvplot/plotting/core.py
@@ -1868,8 +1868,16 @@ class hvPlotTabularPolars(hvPlotTabular):
         import polars as pl
 
         params = dict(self._metadata, **kwds)
-        x = x or params.pop("x", None)
-        y = y or params.pop("y", None)
+        x = (
+            x
+            or params.pop("x", None)
+            or self._data.select(pl.col(pl.NUMERIC_DTYPES)).columns
+        )
+        y = (
+            y
+            or params.pop("y", None)
+            or self._data.select(pl.col(pl.NUMERIC_DTYPES)).columns
+        )
         kind = kind or params.pop("kind", None)
 
         # Find columns which should be converted for LazyDataFrame and DataFrame
--- a/hvplot/plotting/core.py
+++ b/hvplot/plotting/core.py
@@ -1882,12 +1882,11 @@ class hvPlotTabularPolars(hvPlotTabular):
                     for v in params.values()
                     if isinstance(v, (str, list))
                 ]
-                include_all_numeric_cols = x is None or y is None
 
                 columns = (
                     set(self._data.columns) & set(itertools.chain(*possible_columns))
                 ) or {self._data.columns[0]}
-                if include_all_numeric_cols:
+                if x is None or y is None:
                     columns |= set(self._data.select(pl.col(pl.NUMERIC_DTYPES)).columns)
                 xs = x if is_list_like(x) else (x,)
                 ys = y if is_list_like(y) else (y,)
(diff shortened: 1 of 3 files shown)


hvplot/plotting/core.py
```diff
@@ -1882,9 +1882,12 @@ def _get_converter(self, x=None, y=None, kind=None, **kwds):
                     for v in params.values()
                     if isinstance(v, (str, list))
                 ]
+
                 columns = (
                     set(self._data.columns) & set(itertools.chain(*possible_columns))
                 ) or {self._data.columns[0]}
```
Hoxbro, line 1888:
```suggestion
                ) or {next(c for c in self._data.select(pl.col(pl.NUMERIC_DTYPES)).columns if c != x or c not in x)}
```

I think the logic of selecting the first column is what is wrong. Can you try this?

kevinheavey: @Hoxbro that would still drop numeric cols, no? It'll just get the first col that isn't `x`

Hoxbro: Yeah, you are right. I just took a quick look at your example code in the issue. To get parity with pandas, we should get all the numeric columns like what you are doing in this PR. (This does not give the same result with polars as it does with pandas `DataFrame({"a": [1, 2], "b": [3, 4], "c": [5, 6]})` with my suggested change.)

I think we should only add all numeric columns if `y is None` because if `x is None` we will get a range from 0 to the size of the column.

kevinheavey: Yes you are right, I had misread the docs for the x param. Pushed another commit

Hoxbro:
Failing tests are likely to be because of https://github.com/conda-forge/polars-feedstock/issues/205

Can you make a polars pin in setup.py for now?

Hoxbro:
The failing test is because of the change in this PR.

The test should be fixed, either with:
1) Update the instance to be checked for `NdOverlay`
2) Update the plot call with only a `y` argument

I don't have any preference for the two options.

Hoxbro (approved):

maximlt (approved):

Issue #1150, opened 2023-09-21
CLI wrapped around hvplot.explorer


ahuang11:
From my experience, scientists call [ncview](https://cirrus.ucsd.edu/~pierce/software/ncview/quick_intro.html) or [panoply](https://www.giss.nasa.gov/tools/panoply/) in the terminal to do a quick validation on their datasets. This is useful and convenient because they don't have to:

1. Create a new .py / .ipynb (or juggle with an old one, updating file paths), which could be tedious + repetitive
2. Type all the imports, loads, and visualize

Plus, it often supports most legacy, file formats.

The edge that hvplot has over these tools is probably:

1. a newer looking interface
2. interactivity on the map itself
3. datashader support

I think it's valuable to wrap a CLI around hvplot explorer, but not just a simple argparse one, but one that's super user-friendly, like auto-complete, so that it's able to new users are able to immediately jump in and get starting using it--imagine, if the auto-complete can [auto-complete the desired -x and -y from the file](https://typer.tiangolo.com/tutorial/options-autocompletion/#access-other-cli-parameters-with-the-context).

philippjfr:
@Hoxbro and I were just discussing this and we both felt that there is a huge risk of scope creep here. The explorer is great functionality and fits in well in hvPlot itself because it is simply offering a UI around functionality we already provide. However as we get into the application and the CLI around it we start having to build out a lot of other functionality, including the CLI itself, the data loaders, a nice Panel template and application, and lots more. My feeling is that this should be shipped as a separate package entirely.

maximlt:
I also believe this should be a separate project, and to gain adoption I'd say it'd need to be distributed not only as a standalone application (e.g. .exe on Windows). For sure that'd be an interesting project, i.e. picking one scientific domain and building a tool that solves practitioners' needs in that field. This is very likely out of scope of hvPlot though.

ahuang11:
I still think this is important. Personally, when I used to post-process model output, it was nice to do a quick check to ensure the data looked right. This meant I had to navigate the the data dir, e.g. `cd models/output/run/1`, `pwd`, copy the directory path and file name, and then paste it into a notebook. This was quite tedious and lots of boilerplate code (import, read, plot) I had to type every time I wanted to verify output.

So, I imagine a very thin CLI wrapper around hvplot, the kind defaulting to explorer:

`hvplot test.nc -x lon -y lat -c air --groupby time`

This internally invokes
```python
import xarray as xr
import hvplot.xarray

[...]

PR #1256, opened 2024-01-20
Add Command Line Interface Wrapper

ahuang11:
Addresses https://github.com/holoviz/hvplot/issues/1150

```bash
This command-line tool allows for quick and easy visualization of various
data file formats using HoloViews and hvPlot, including CSV, JSON, HTML, Excel,
Parquet, NetCDF, and HDF5, and more.

The tool will automatically determine the appropriate reader for the file
extension and use it to load the data. Ensure the correct dependencies for
file format support are installed (e.g., pandas for CSV, xarray for NetCDF).

All of the keywords supported by hvPlot can be passed as arguments to the
command-line tool. See the hvPlot customization docs for available kwargs:
https://hvplot.holoviz.org/user_guide/Customization.html

Examples:
    hvplot iris.csv
    hvplot air.nc x=lon y=lat groupby=time,level geo=true
    hvplot detrend.nino34.ascii.txt x=YR y=ANOM -r pandas.read_csv -rk delimiter='\s+'

positional arguments:
  file_path             Path to the data file
  hvplot_kwargs         HoloViews options in 'key=value' format

options:
  -h, --help            show this help message and exit
  --opts [OPTS ...]     HoloViews plot options in 'key=value' format
  -r READER, --reader READER
                        Fully-qualified name of the reader function to use, e.g. pandas.read_csv
  -rk READER_KWARGS, --reader_kwargs READER_KWARGS
                        Reader options in 'key=value' format
  --port PORT           Port to use for displaying the plot
  -o OUTPUT_PATH, --output_path OUTPUT_PATH
                        Path to save the output file
```

Demo:

https://github.com/holoviz/hvplot/assets/15331990/cb4005d0-aa95-4481-bf46-e879e717c75d

### DIFF
--- a/hvplot/cli.py
+++ b/hvplot/cli.py
@@ -102,7 +102,7 @@ def parse_arguments():
     )
     parser.add_argument("file_path", type=str, help="Path to the data file")
     parser.add_argument(
-        "hvplot_kwargs", nargs="*", help="HoloViews options in 'key=value' format"
+        "hvplot_kwargs", nargs="*", help="hvPlot arguments in 'key=value' format"
     )
     parser.add_argument(
         "--opts", nargs="*", help="HoloViews plot options in 'key=value' format"
@@ -111,7 +111,7 @@ def parse_arguments():
         "-r",
         "--reader",
         type=str,
-        help="Fully-qualified name of the reader function to use, e.g. pandas.read_csv",
+        help="Fully-qualified name of the data-reader function to use, e.g. pandas.read_csv",
     )
     parser.add_argument(
         "-rk",
--- a/hvplot/cli.py
+++ b/hvplot/cli.py
@@ -1,0 +1,159 @@
+import argparse
+import os
+import difflib
+
+import hvplot
+import panel as pn
+
+_MODULES_KWARGS = {
+    "xarray": {
+        "extensions": [".nc", ".hdf5", ".h5", ".zarr", ".grib"],
+        "default_reader": "open_dataset",
+        "extras": {
+            ".zarr": {"kwargs": {"engine": "zarr"}},
+            ".grib": {"kwargs": {"engine": "cfgrib"}},
+        },
+    },
+    "pandas": {
+        "extensions": [".csv", ".json", ".html", ".xls", ".xlsx", ".parquet"],
+        "default_reader": "read_csv",
+        "extras": {
+            ".json": {"reader": "read_json"},
+            ".html": {"reader": "read_html"},
+            ".xls": {"reader": "read_excel"},
+            ".xlsx": {"reader": "read_excel"},
+            ".parquet": {"reader": "read_parquet"},
+        },
+    },
+    "geopandas": {
+        "extensions": [".shp", ".geojson", ".gml"],
+        "default_reader": "read_file",
+    },
+}
+_EXTENSION_READERS = {}
+
+
+def import_and_register_module(module_name, extensions, default_reader, extras=None):
+    try:
+        module = __import__(module_name)
+        if module_name == "geopandas":
+            module_name = "pandas"  # Ensure hvplot.pandas is imported for geopandas
+        __import__("hvplot." + module_name)
+
+        for ext in extensions:
+            reader_kwargs = {"reader": default_reader}
+            if extras and ext in extras:
+                reader_kwargs.update(extras[ext])
+
+            if hasattr(module, reader_kwargs["reader"]):
+                _EXTENSION_READERS[ext] = (
+                    getattr(module, reader_kwargs["reader"]),
+                    reader_kwargs.get("kwargs", {}),
+                )
+    except ImportError:
+        pass
+
+
+def setup_extension_readers(extension):
+    for module_name, module_kwargs in _MODULES_KWARGS.items():
+        if extension in module_kwargs["extensions"]:
+            import_and_register_module(module_name, **module_kwargs)
+            break
+
+
+def ensure_variable_exists(data, input_, variables):
+    if input_ is None:
+        return
+
+    if input_ not in variables:
+        fuzzy_matches = difflib.get_close_matches(input_, variables, n=1)
+        if fuzzy_matches:
+            print(f"Using {fuzzy_matches[0]} instead of {input_!r}")
+            input_ = fuzzy_matches[0]
+        else:
+            raise ValueError(
+                f"{input_!r} not in data, available variables are {variables!r}"
+            )
+    return input_
+
+
+def parse_inputs(inputs):
+    kwargs = {}
+    if not inputs:
+        return kwargs
+
+    for input_ in inputs:
+        key, value = input_.split("=")
+        if "," in value:
+            value = [v for v in value.strip("[]").strip("()").split(",") if v]
+        elif value.lower() in ("true", "false"):
+            value = value == "true"
+        kwargs[key] = value
+    return kwargs
+
+
+def parse_arguments():
+    parser = argparse.ArgumentParser(description="hvPlot CLI")
+    parser.add_argument("file_path", type=str, help="Path to the data file")
+    parser.add_argument(
+        "hvplot_kwargs", nargs="*", help="HoloViews options in 'key=value' format"
+    )
+    parser.add_argument(
+        "--opts", nargs="*", help="HoloViews plot options in 'key=value' format"
+    )
+    parser.add_argument(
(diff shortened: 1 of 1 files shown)


ahuang11:
I think to make it easier to change file name and use the same option, I should make path last: e.g.
`hvplot x=lon y=lat groupby=time,level geo=true --path air.nc`
`hvplot x=lon y=lat groupby=time,level geo=true --path airx2.nc`

hvplot/cli.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def setup_extension_readers(extension):
+    for module_name, module_kwargs in _MODULES_KWARGS.items():
+        if extension in module_kwargs["extensions"]:
+            import_and_register_module(module_name, **module_kwargs)
+            break
+
+
+def parse_inputs(inputs):
```
jbednar, line 81: Please add docstrings to this and other functions in this file, stating the intent of the function and the assumptions it makes about its arguments.

hvplot/cli.py
```diff
@@ -0,0 +1,179 @@
...
+    return kwargs
+
+
+def parse_arguments():
+    parser = argparse.ArgumentParser(
+        description=HELP_DESCRIPTION, formatter_class=argparse.RawTextHelpFormatter
+    )
+    parser.add_argument("file_path", type=str, help="Path to the data file")
+    parser.add_argument(
+        "hvplot_kwargs", nargs="*", help="HoloViews options in 'key=value' format"
```
jbednar, line 105:
```suggestion
        "hvplot_kwargs", nargs="*", help="hvPlot arguments in 'key=value' format"
```

(Is that what you meant, to distinguish this from `--opts`?)

hvplot/cli.py
```diff
@@ -0,0 +1,179 @@
...
+    )
+    parser.add_argument(
+        "--opts", nargs="*", help="HoloViews plot options in 'key=value' format"
+    )
+    parser.add_argument(
+        "-r",
+        "--reader",
+        type=str,
+        help="Fully-qualified name of the reader function to use, e.g. pandas.read_csv",
```
jbednar, line 114:
```suggestion
        help="Fully-qualified name of the data-reader function to use, e.g. pandas.read_csv",
```

(maybe clearer than just saying "reader"?)

hvplot/cli.py
```diff
@@ -0,0 +1,179 @@
...
+            hvplot_kwargs[key] = [value]
+
+    # Process opts
+    opts_kwargs = parse_inputs(args.opts)
+    if opts_kwargs and hvplot_kwargs["kind"] == "explorer":
+        hvplot_kwargs["opts"] = opts_kwargs
+
+    plot = data.hvplot(**hvplot_kwargs)
+    if opts_kwargs and hvplot_kwargs["kind"] != "explorer":
+        plot = plot.opts(**opts_kwargs)
```
jbednar, line 170: I can't quite follow the logic over these few lines, but does it achieve making the Explorer just as configurable as a non-explorer plot (with HoloViews options supported, not just hvPlot keywords)? If not, can that be supported (both for Explorer in Jupyter and Explorer via the CLI)?

jbednar (commented):
This looks just like what I had in mind!  Lightweight, clear, and useful!

Is it too much to ask for the Explorer "code" tab to show both Python code and CLI code to generate the selected plot? :-)

> "PR with an MVP of the proposed CLI along with a list of desired but unimplemented features and a list of non-features (things explicitly not considered in scope)."

Can you add such a list to this PR or as a separate issue? I.e. one list of desired future features (with checkboxes indicating they are not implemented), plus another list of things you do not feel belong in an hvPlot CLI (e.g. input filtering and transformation beyond what Explorer does).

Separately from this, I've been thinking that the Explorer should initially usually start with just a hamburger button, without the various selectors visible, and those would show up only when the user clicks on the hamburger. That's independent of this PR, but it would help avoid the user having to make a strict choice between a "plot" and an "explorer"; it would just be a plot that then can be reconfigured if desired. I think that approach works well from the cli, and also in other contexts where the Explorer is used, e.g. Jupyter.

ahuang11:
> Is it too much to ask for the Explorer "code" tab to show both Python code and CLI code to generate the selected plot? :-)

Super easy I think.

> Separately from this, I've been thinking that the Explorer should initially usually start with just a hamburger button, without the various selectors visible, 

I think Panel needs a Sidebar card layout that does this.

hvplot/cli.py
ahuang11: hvplot_explorer is not a native HoloViews object (it's a HvDataFrameExplorer for example) so it doesn't accept an opts method. In the near future, I imagine we can add an opts method to forward the opts rather than these logic.

jbednar: Ok, makes sense!

jbednar:
@maximlt , would your error message have been any different when invoking hvPlot from within Python on the same file?

TomNicholas:
This seems extremely useful, especially if it can be run simply over ssh. The number of co-workers who know about pangeo but still use ncview sometimes is high. One might argue that its an alternative path to lowering the difficulty of using hvplot for simple visualizations.  I have no strong opinions about which repo it lives in.

philippjfr:
Over `ssh` you'd have to tunnel the port I suppose, but that does seem workable.

ahuang11:
Yes tunneling is quite straightforward and common; that's how I used Jupyter notebooks before. https://bluewaters.ncsa.illinois.edu/pythonnotebooks

maximlt:
We haven't managed to reach a consensus on whether this feature belongs to hvPlot or not. @droumis did an amazing job when setting up the HoloViz and per-project governance models, giving us a [decision model](https://github.com/holoviz/holoviz/blob/hvplot-gov/doc/governance/project-docs/GOVERNANCE.md#2-decisions) for this kind of case. Any maintainer can call a vote, let's do that then.

@philippjfr @Hoxbro @ahuang11 please reply with your vote, either `+1` or `-1`. If you want to add more motivation than what you wrote in this PR or https://github.com/holoviz/hvplot/issues/1150, please do so.

-1 for me:

- I have doubt the CLI would in fact replace the more dedicated applications scientists use to explore their files, it's hard to compete with specialized tools.
- My wish for hvPlot, as explained in its new [roadmap](https://hvplot.holoviz.org/roadmap.html), would be to clean up what the existing (docs, developer experience, bug fixes, tests) to make the library more robust and pleasant to use. Adding a CLI goes in the opposite direction in my opinion, requiring us to add another entry point in the doc, write new kinds of tests, manage more dependencies on the CI, take care of more bug reports and feature requests, etc.

To precise my vote, I would like to say that it is a temporary -1, I would be open to revisiting this idea in the future. If we vote *no*, I'd suggest keeping https://github.com/holoviz/hvplot/issues/1150 open to let users interested in this feature chime in/upvote, updating the OP with some more context and the scope of the feature (the issue focused on xarray but the implementation in this PR interfaces with other libraries/file types).

ahuang11:
+1:
- other tools require `sudo apt-get install`, conda, or manual build; hvplot is just a `pip install` away
- additionally, hvPlot is extremely interactive; zoom, hover over values, and also adjusting variable selection
- the CLI is a thin wrapper; not much addition besides a file ingestion engine so not much to additionally maintain; we can do a feature freeze to work out other issues. understandably this is a bit of a scope creep, but I imagine covers 90% of use cases

However, I'm okay if this lives elsewhere, and actually starting to lean towards Lumen using `lumen explore test.nc` which will format a template Lumen spec with the necessary inputs. This can help bring attention to Lumen too and expose Lumen to the gridded dataset world too. Else, I'd still like to release this somewhere like under a new holoviz-incubator or even my personal account so others can take advantage of it.

Hoxbro:
-1

philippjfr:
On balance I will abstain from the vote and record my vote as -0. I am not the day to day maintainer for hvPlot so the maintenance concerns are not front of mind for me, so I primarily want to respect @maximlt's concerns as lead maintainer. I do see the utility for the CLI particularly for the climate community working with gridded data but would like to work towards supporting that at the Lumen level.

maximlt:
Oops, I forgot to come back to this! With 2 `-1`, 1 `+1` and 1 `-0`, we're not going to move forward with this feature. I'm going to close this PR. However, as I already wrote, I suggest keeping the issue open and updating its initial message to better reflect the scope of the intended CLI, and encourage people to upvote the issue/chime in.

PR #1257, opened 2024-01-22
Compatibility with Pandas 2.2

maximlt:
- Vendoring some of the `make<>` private testing utilities that were removed in Pandas 2.2
- Update a test to remove a FutureWarning

### DIFF
--- a/hvplot/tests/testinteractive.py
+++ b/hvplot/tests/testinteractive.py
@@ -13,6 +13,7 @@ import xarray as xr
 from holoviews.util.transform import dim
 from hvplot import bind
 from hvplot.interactive import Interactive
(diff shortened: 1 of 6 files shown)


hvplot/tests/testinteractive.py
```diff
@@ -194,7 +195,7 @@ def sel_col(sel):
 
 
 def test_interactive_nested_widgets():
-    df = pd._testing.makeDataFrame()
+    df = makeDataFrame()
```
philippjfr, line 198: Maybe let's start using the fixture more.

maximlt: The `df` fixture in this module that retunes `makeMixedDataFrame` is used a lot. Planning to refactor the tests in a future PR.

Issue #1244, opened 2024-01-02
Document that hvPlot works with Polars


MarcSkovMadsen:
I think it would be valuable for our users to know it works with Polars as well as other backends. 

For example 

- [x] In any architecture drawing

![hvPlot Architecture](https://camo.githubusercontent.com/592b9657aa69d6f6a0f1f1ee81fdea1766929d429a83d45da5d7d809705443de/68747470733a2f2f6876706c6f742e686f6c6f76697a2e6f72672f6173736574732f6469616772616d2e737667)

- [x] In the README section [hvPlot works with the tools you know and love](https://github.com/holoviz/hvplot?tab=readme-ov-file#hvplot-works-with-the-tools-you-know-and-love)
- [x] At the [top of the home page](https://hvplot.holoviz.org/index.html#hvplot)
- [x] At the [backend tabs section](https://hvplot.holoviz.org/index.html#hvplot)

PR #1262, opened 2024-01-28
Doc: better expose polars

maximlt:
Addresses https://github.com/holoviz/hvplot/issues/1244

The main diagram has been updated to include Polars and Ibis. The logos of other libraries has been updated too.


MarcSkovMadsen:
Does Ibis work in practice. I tried it But found it unfinished and not working a year ago. Not knowing how much had to be solved I started some PRs to try to fix. But I ended up giving up.

maximlt:
There have been a few minor fixes but I have no doubt there are many others. This is exactly why I put *Bug fixes* in the roadmap after docs and developer experience:

>  hvPlot provides already many (many!) features to its users. However, it also has many bugs that prevent its adoption.
    - hvPlot’s development should favor bug fixes vs. new features or feature enhancements.
    - hvPlot’s test suite should be re-factored and improved to catch bugs before users have to report them and to prevent potential regressions.
    - Bugs reported by hvPlot users regularly come from one of its dependencies, typically HoloViews. It should be made easy for hvPlot users to find these bugs on GitHub. These bugs should be fixed too.

PR #1269, opened 2024-02-01
Clarify CRS and output projection

ahuang11:
A user brought this to my attention and mentioned that it was hard to figure out crs and projection.

Turns out the section `Declaring an output projection` was hidden under Spatialpandas
image

This fixes it by reducing the heading of Spatialpandas from `##` to `###`, moving "Declaring an output projection" under "Declaring an output projection".

Lastly, it clarifies the role of crs / projection and adds `projection` to the missing geographic options


maximlt (approved):
LGTM!

PR #1275, opened 2024-02-05
Pin intake to version 2.0.0

Hoxbro:
I could see downstream tests failing because of the intake 2.0 release. https://github.com/holoviz/hvplot/actions/runs/7781746476/job/21216765359


maximlt (approved):
Thanks!

PR #1278, opened 2024-02-12
Pin holoviews for test runs

Hoxbro:
Can see we are pulling in 1.18 alpha versions instead of 1.17.X, causing us to not see regression, spotted by Maxime.


Hoxbro:
I was confused about the cause. I will close the issue.

Issue #1251, opened 2024-01-05
`robust` or `clim_percentile` option to get the 2nd / 98th percentile 


ahuang11:
```python
from pystac_client import Client
from odc.stac import load
import numpy as np

client = Client.open("https://earth-search.aws.element84.com/v1")
collection = "sentinel-2-l2a"
tas_bbox = [146.5, -43.6, 146.7, -43.4]
search = client.search(collections=[collection], bbox=tas_bbox, datetime="2023-12")

ds = load(search.items(), bbox=tas_bbox, groupby="solar_day", chunks={})
da = ds[["red", "green", "blue"]].isel(time=2).load().to_array()
da.plot.imshow(robust=True)
```

<img width="717" alt="image" src="https://github.com/holoviz/hvplot/assets/15331990/cde2fe18-c239-4b66-a52b-7dd4d6f2cc59">

Without robust:
<img width="705" alt="image" src="https://github.com/holoviz/hvplot/assets/15331990/b74e4644-c0db-4542-9279-df6061c9b851">

Which is what hvplot yields.
<img width="924" alt="image" src="https://github.com/holoviz/hvplot/assets/15331990/c453b781-a433-4252-8b32-f1d29ddff8cd">

When I try to manually calculate it and pass a custom `clim`, there's no effect
https://github.com/pydata/xarray/blob/492aa076bd1812af71d68dc221312fb9a199d2d3/xarray/plot/utils.py#L215-L225

maximlt:
HoloViews has the `clim_percentile` plot option but that only applies to plot types that inherit from `ColorbarPlot`, while `RGBPlot` inherits from `LegendPlot`. @ahuang11 I'd say we should first add this plot option to HoloViews? To eventually expose it in hvPlot, where I'd say `robust` would be a good choice given that `robust=True` would be already familiar to Xarray users.

ahuang11:
I was looking at https://github.com/holoviz/holoviews/blob/6aa6c9ddb7c9308531753efae41cbd0873c975a3/holoviews/plotting/bokeh/raster.py#L134 and I'm not sure there's a way to add clim_percentile to RGBPlot without updating Bokeh too.

I tried doing:
```python     
        if self.clim_percentile:
            low, high = 0, 1
            style["color_mapper"] = LinearColorMapper(low=low, high=high)
```

```python
AttributeError [Call holoviews.ipython.show_traceback() for details]
[...]

PR #1291, opened 2024-03-04
Add robust keyword

ahuang11:
Closes https://github.com/holoviz/hvplot/issues/1251

```python
from pystac_client import Client
from odc.stac import load
import hvplot.xarray

client = Client.open("https://earth-search.aws.element84.com/v1")
collection = "sentinel-2-l2a"
tas_bbox = [146.5, -43.6, 146.7, -43.4]
search = client.search(collections=[collection], bbox=tas_bbox, datetime="2023-12")

ds = load(search.items(), bbox=tas_bbox, groupby="solar_day", chunks={})
da = ds[["red", "green", "blue"]].isel(time=2).load().to_array()

da.hvplot.rgb("x", "y", bands="variable", robust=True)
```

image


hvplot/converter.py
```diff
@@ -819,6 +826,18 @@ def _process_data(self, kind, data, x, y, by, groupby, row, col,
...
+                # https://github.com/pydata/xarray/blob/main/xarray/plot/utils.py#L729
+                vmax = np.nanpercentile(data[z], 100 - 2)
+                vmin = np.nanpercentile(data[z], 2)
+                # Scale interval [vmin .. vmax] to [0 .. 1], with darray as 64-bit float
+                # to avoid precision loss, integer over/underflow, etc with extreme inputs.
+                # After scaling, cast to 8-bit integer.
+                data[z] = (
+                    ((data[z].astype("f8") - vmin) / (vmax - vmin)) * 255
+                ).astype(int)
+                print(data[z])
```
ahuang11, line 840:
```suggestion
```

ahuang11:
Any thoughts about this? I'd like to use it for a tutorial.

maximlt:
Oh sorry, I didn't realize this was ready. I'll review it today.

maximlt:
> I'd like to use it for a tutorial.

Do you need a release? If so, when?

ahuang11:
Not in a rush! Maybe next week?

hvplot/converter.py
```diff
@@ -381,7 +387,7 @@ def __init__(
         self, data, x, y, kind=None, by=None, use_index=True,
         group_label=None, value_label='value', backlog=1000,
         persist=False, use_dask=False, crs=None, fields={},
-        groupby=None, dynamic=True, grid=None, legend=None, rot=None,
+        groupby=None, dynamic=True, grid=None, legend=None, robust=None, rot=None,
```
maximlt, line 390: Perhaps let's stuff it after `clim`? Trying to have some sort of semantic ordering in that mess :D

hvplot/converter.py
```diff
@@ -819,6 +826,17 @@ def _process_data(self, kind, data, x, y, by, groupby, row, col,
...
+                # adapted from xarray
+                # https://github.com/pydata/xarray/blob/main/xarray/plot/utils.py#L729
+                vmax = np.nanpercentile(data[z], 100 - 2)
+                vmin = np.nanpercentile(data[z], 2)
+                # Scale interval [vmin .. vmax] to [0 .. 1], with darray as 64-bit float
+                # to avoid precision loss, integer over/underflow, etc with extreme inputs.
+                # After scaling, cast to 8-bit integer.
+                data[z] = (
+                    ((data[z].astype("f8") - vmin) / (vmax - vmin)) * 255
+                ).astype(int)
```
maximlt, line 839: I was a bit surprised to see this warning running this example with `robust=True`:

> WARNING:param.RGBPlot00947: Clipping input data to the valid range for RGB data ([0..1] for floats or [0..255] for integers).

It seems to me the different with xarray's implementation is that they clip the data to [0..1]

https://github.com/pydata/xarray/blob/6af547cdd9beac3b18420ccb204f801603e11519/xarray/plot/utils.py#L759

Why do we deviate from xarray's implementation? There may be a good reason I'm missing! If there is, perhaps the docstring shouldn't say the data is clipped?

hvplot/converter.py
```diff
@@ -819,6 +826,17 @@ def _process_data(self, kind, data, x, y, by, groupby, row, col,
             data, x, y, by_new, groupby_new = process_xarray(
                 data, x, y, by, groupby, use_dask, persist, gridded,
                 label, value_label, other_dims, kind=kind)
+            if robust:
```
maximlt, line 829: Shouldn't this block only run for RGB elements?

hvplot/converter.py
```diff
@@ -819,6 +826,17 @@ def _process_data(self, kind, data, x, y, by, groupby, row, col,
...
+                # adapted from xarray
+                # https://github.com/pydata/xarray/blob/main/xarray/plot/utils.py#L729
```
maximlt, line 831: Nit but I prefer GH permalinks.

```suggestion
                # https://github.com/pydata/xarray/blob/6af547cdd9beac3b18420ccb204f801603e11519/xarray/plot/utils.py#L729
```

hvplot/converter.py
ahuang11: The clipping is happening in HoloViews if that's what you're asking.

maximlt: But this leads to HoloViews displaying this warning. Do you expect that?

> WARNING:param.RGBPlot00947: Clipping input data to the valid range for RGB data ([0..1] for floats or [0..255] for integers).

hvplot/converter.py
maximlt: In hvPlot if I now set `robust=True` to display an Image element, I think that's going to combine the rescaling you're doing in this block, together with `clim_percentile=True`. Is that expected?

ahuang11: Ah, you're right about the scaling part.

ahuang11:
Follows xarray now.

image

image

maximlt (approved):
LGTM!

Issue #1293, opened 2024-03-07
Allow to pass options to tiles


maximlt:
It'd be useful to be able to pass down options to style the tiles set up with `tiles=...`, with e.g. `tile_opts={'alpha': 0.5}`.

PR #1299, opened 2024-03-26
Add `tiles_opts` parameter to customize the tiles

maximlt:
Addresses https://github.com/holoviz/hvplot/issues/1293

An alternative implementation might have been overloading `tiles` to accept a dictionary like `{"tiles": "ESRI", "alpha": 0.5}`. Interested in what other maintainers think about the two different approaches.

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -247,6 +247,9 @@ class HoloViewsConverter:
         Whether to overlay the plot on a tile source. Tiles sources
         can be selected by name or a tiles object or class can be passed,
         the default is 'Wikipedia'.
+    tiles_opts (default=None): dict
+        Options to customize the tiles layer created when `tiles` is set
+        (e.g. `dict(alpha=0.5)`.
     """
 
     _gridded_types = [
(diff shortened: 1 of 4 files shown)


ahuang11:
tiles_opts follows convention of what other keywords do:
`legend_opts` `colorbar_opts`

maximlt:
Yes although these two keywords aren't hvPlot parameters but HoloViews plot options. The `features` parameter of hvPlot can be overloaded: 

    features (default=None): dict or list
        A list of features or a dictionary of features and the scale
        at which to render it. Available features include 'borders',
        'coastline', 'lakes', 'land', 'ocean', 'rivers' and 'states'.
        Available scales include '10m'/'50m'/'110m'.

It's because of this parameter I thought that maybe `tiles` could also be overloaded, instead of adding `tiles_opts`, as hvPlot suffers more from new parameters added to its flat signature than HoloViews. Perhaps with something like:

- `tiles={"alpha":0.5}` (== `tiles=True, tiles_opts={"alpha": 0.5}`)
- `tiles={"tiles": "ESRI", "alpha": 0.5}` (== `tiles="ESRI", tiles_opts={"alpha": 0.5}`)

I think in the end I have a slight preference for `tiles_opts`. Opinions @philippjfr @Hoxbro ?

ahuang11:
`tiles={"tiles": "ESRI", "alpha": 0.5}` is funky; if dict, I think
`tiles={"name": "ESRI", "alpha": 0.5}` would be more suitable, but yes I lean towards `tiles_opts` or `tiles_kwargs`

Hoxbro:
I like `tiles_opts` too.

philippjfr:
+1 on tiles_opts.

hvplot/converter.py
```diff
@@ -416,6 +419,7 @@ def __init__(
         self.coastline = coastline
         self.features = features
         self.tiles = tiles
+        self.tiles_opts = {} if tiles_opts is None else tiles_opts
```
ahuang11, line 422: What about:
```suggestion
        self.tiles_opts = tiles_opts or {}
```

PR #1302, opened 2024-04-08
Add user guide for working with large time-series datasets

droumis:
supersedes https://github.com/holoviz/hvplot/pull/1205

This adds a notebook that explains the different ways of working with large time-series datasets with holoviz

### DIFF
--- a/doc/user_guide/Large_Timeseries.ipynb
+++ b/doc/user_guide/Large_Timeseries.ipynb
@@ -5,6 +5,8 @@
    "id": "artificial-english",
    "metadata": {},
    "source": [
+    "# Large Timeseries Data\n",
+    "\n",
     "Effectively representing temporal dynamics in large datasets requires selecting appropriate visualization techniques that ensure responsiveness while providing both a macroscopic view of overall trends and a microscopic view of fine details. This guide will explore various methods, such as **WebGL Rendering**, **LTTB Downsampling**, **Datashader Rasterizing**, and **Minimap Contextualizing**, each suited for different aspects of large timeseries data visualization. We predominantly demonstrate the use of hvPlot syntax, leveraging HoloViews for more complex requirements. Although hvPlot supports multiple backends, including Matplotlib and Plotly, our focus will be on Bokeh due to its advanced capabilities in handling large timeseries data.\n",
     "\n",
     "\n",
@@ -45,11 +47,15 @@
     "\n",
     "## WebGL Rendering\n",
     "\n",
+    "[WebGL](https://developer.mozilla.org/en-US/docs/Web/API/WebGL_API) is a JavaScript API that allows rendering content in the browser using hardware acceleration from a Graphics Processing Unit (GPU). WebGL is standardized and available in all modern browsers.\n",
+    "\n",
     "### Canvas Rendering - Prior Default\n",
+    "\n",
     "Rendering Bokeh plots in hvPlot or HoloViews has evolved significantly. Prior to 2023, Bokeh's custom HTML **Canvas** rendering was the default. This approach works well for datasets up to a few tens of thousands of points but struggles above 100K points, particularly in terms of zooming and panning speed. These days, if you want to utilize Bokeh's Canvas rendering, use `import holoviews as hv; hv.renderer(\"bokeh\").webgl = False` prior to creating your hvPlot or HoloViews object.\n",
     "\n",
     "### WebGL Rendering - Current Default\n",
-    "Around mid-2023, the adoption of improved **WebGL** as the default for hvPlot and HoloViews allowed for smoother interactions with larger datasets by utilizing GPU-acceleration. It's important to note that WebGL performance can vary based on your machine's specifications. For example, some Mac models may not exhibit a marked improvement in WebGL performance over Canvas due to GPU hardware configuration."
+    "\n",
+    "Around mid-2023, the adoption of improved **WebGL** as the default for hvPlot and HoloViews allowed for smoother interactions with larger datasets by utilizing GPU-acceleration. It's important to note that WebGL performance can vary based on your machine's specifications. For example, some Apple Mac models may not exhibit a marked improvement in WebGL performance over Canvas due to GPU hardware configuration."
    ]
   },
   {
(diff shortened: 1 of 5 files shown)


droumis:
Not sure why this page isn't showing up in the toc on the dev site

image

doc/user_guide/Large_Timeseries.ipynb
```diff
@@ -166,22 +173,6 @@
     "This makes LTTB an ideal default method for exploring timeseries datasets, particularly when the dataset size is unknown or too large for standard WebGL rendering."
    ]
   },
-  {
```
maximlt: I'm not really for documenting unreleased features. But something we could do would be to document this as if HoloViews 1.19.0 was already released. In that case, we should also update the code so it's already able to accept and pass the new values. What do you think?

doc/user_guide/Large_Timeseries.ipynb
```diff
@@ -242,7 +233,7 @@
    "outputs": [],
    "source": [
     "df0.hvplot(x=\"time\", y=\"value\", rasterize=True, cnorm='eq_hist', padding=(0, 0.1),\n",
-    "           min_height=300, autorange='y', title=\"Datashade\", colorbar=False, line_width=2)"
+    "           min_height=300, autorange='y', title=\"Rasterize\", colorbar=False, line_width=2)"
```
maximlt, line 236: Changed the title since `rasterize=True` is set.

doc/user_guide/Large_Timeseries.ipynb
```diff
@@ -298,6 +289,7 @@
    "outputs": [],
    "source": [
     "waves = pd.read_parquet(\"https://datasets.holoviz.org/waveform/v1/waveforms.parq\")\n",
+    "print(len(waves))\n",
```
maximlt, line 292: Just to see how many points the dataset has!

maximlt (commented):
I've pushed a few changes, please let me know if you disagree with some of them, they can always be reverted.

My main comment, which applies to pretty much every user guide that attempts to demonstrate Datashader, is that the file size ends up being quite large, close to 30MB in this case. If you don't have a super fast connection, you can see the pages being loaded slowly (this one takes so long for me https://holoviews.org/user_guide/Large_Data.html!), and there's also some time spent rendering the page and its plots that have many data points. At the same, most of these plots would deserve to be inspected with a live Python kernel to see the full benefit of the applied approached (LTTB, datashader).
Instead of displaying the real Bokeh plots, wouldn't it be better if we displayed pretty images or GIFs? (I don't want to block this PR if we think that's what we should do, that could be done in a 2nd iteration).

maximlt:
> Not sure why this page isn't showing up in the toc on the dev site

It's now showing up.

droumis:
> My main comment, which applies to pretty much every user guide that attempts to demonstrate Datashader, is that the file size ends up being quite large, close to 30MB in this case. If you don't have a super fast connection, you can see the page being loaded slowly (this one takes so long for me https://holoviews.org/user_guide/Large_Data.html!), and there's also some time spent rendering the page and its plots that have many data points. At the same time, most of these plots would deserve to be inspected with a live Python kernel to see the full benefit of the applied approach (LTTB, datashader).
Instead of displaying the real Bokeh plots, wouldn't it be better if we displayed pretty images or GIFs? (I don't want to block this PR if we think that's what we should do, that could be done in a 2nd iteration).

We decided in a meeting to not block this PR.. it would be nice in the future to have some mechanism to optionally facilitate images in docs rather than big plots

doc/user_guide/Large_Timeseries.ipynb
droumis: sounds good, but just curious if there was a reason why you also didn't change the subsequent datashader categories example to `rasterize`? Anyways, I changed it to rasterize, let me know if there's a good reason to use datashade instead.

doc/user_guide/Large_Timeseries.ipynb
droumis: Ok, I included a note that these options are available starting in HoloViews 1.19.0. @Hoxbro please be sure to include the tsdownsample [PR](https://github.com/holoviz/holoviews/pull/6059) in the next minor release. I don't think anything further needs to be updated in hvPlot code.


```markdown
### Enhanced Downsampling Options

Starting in HoloViews version 1.19.0, integration with the [tsdownsample](https://github.com/predict-idlab/tsdownsample) library introduces enhanced downsampling functionality with the following methods, which will be accepted as inputs to `downsample` in hvPlot:

- **lttb**: Implements the Largest Triangle Three Buckets ([LTTB](https://github.com/predict-idlab/tsdownsample?tab=readme-ov-file#:~:text=performs%20the-,Largest%20Triangle%20Three%20Buckets,-algorithm)) algorithm, optimizing the selection of points to retain the visual shape of the data.
- **minmax**: For each segment of the data, this method retains the minimum and maximum values, ensuring that peaks and troughs are preserved.
- **minmax-lttb**: A hybrid approach that combines the minmax strategy with LTTB.
- **m4**: A [multi-step process](https://www.vldb.org/pvldb/vol7/p797-jugel.pdf) that leverages the min, max, first, and last values for each time segment.
```

droumis:
ready for merge?

doc/user_guide/Large_Timeseries.ipynb
maximlt: Either I missed it or I wasn't sure what the intention was, anyway, it's now fixed :)

doc/user_guide/Large_Timeseries.ipynb
maximlt: >  I don't think anything further needs to be updated in hvPlot code.

Not quite right I think, `downsample` can just be a boolean to use `downsample1d`, I don't think there is a way to customize how it's called from hvPlot at the moment (or if there is, it's not obvious and needs to be documented). I'll open an issue, but since multiple algorithms are documented here, it should be implemented before the release.

maximlt:
> We decided in a meeting to not block this PR.. it would be nice in the future to have some mechanism to optionally facilitate images in docs rather than big plots

Yep, this sort of this is falling in our ever-growing bucket of nice-to-have-but-will-likely-never-happen! By the way, the [Large Data guide](https://holoviews.org/user_guide/Large_Data.html) in HoloViews does have a couple of GIFs, so adding GIFs to show what Datashader is capable of isn't unprecedented. I still think that would be the best thing to do to both make the page faster to load and better demonstrate hvPlot with Datashader, but as I don't really want to do it now, I'll just put that in the bucket too!

> ready for merge?

Yes!

Issue #636, opened 2021-07-09
Add bgcolor argument


jbednar:
As far as I can tell, the Pandas .plot() API does not have a way to [change the background color](https://stackoverflow.com/questions/23123272/change-pandas-plot-background-color) without dropping down to matplotlib. hvPlot appears to inherit this the same limitation. We can of course do `.opts(bgcolor='black')`, but I've found that in many examples that's the only use of .opts I have to resort to, and without it I wouldn't need to explain .opts.  Can we have a `bgcolor=` argument to cover this case?

maximlt:
Implemented in https://github.com/holoviz/hvplot/pull/1308

PR #1308, opened 2024-04-10
Allow to set the background color with `bgcolor`

maximlt:
Implements https://github.com/holoviz/hvplot/issues/636

@droumis FYI I have noticed that the converter `__init__` signature does not include this kind of plot options passed down to HoloViews. Perhaps they should be added to the signature to later be included in some API reference docs.

### DIFF
--- a/doc/reference/tabular/area.ipynb
+++ b/doc/reference/tabular/area.ipynb
@@ -47,8 +47,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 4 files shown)


droumis:
> @droumis FYI I have noticed that the converter __init__ signature does not include this kind of plot options passed down to HoloViews. Perhaps they should be added to the signature to later be included in some API reference docs.

Yep, I think it would help most for discoverability via auto-complete

Issue #1105, opened 2023-07-14
Bouncing / stretching when coastline=True and rasterize=True in VSCode


ahuang11:
```
import numpy as np
import pandas as pd
import hvplot.pandas

df = pd.DataFrame(np.random.multivariate_normal((0,0), [[0.1, 0.1], [0.1, 1.0]], (500000,)))
df.hvplot.points('0', '1', rasterize=True, coastline=True, global_extent=True)
```

https://github.com/holoviz/hvplot/assets/15331990/d4004c88-f90c-42fa-a57d-34b32ae46a4a

ahuang11:
Maybe more apparent with tiles:

https://github.com/holoviz/hvplot/assets/15331990/3feecfae-693c-4133-b868-0873206b79d4

ahuang11:
Duplicate of https://github.com/holoviz/geoviews/issues/673

ahuang11:
`pip install "jupyter-bokeh>=4.0.4"` should fix this.

PR #1311, opened 2024-04-12
Remove auto setting data aspect for geo

ahuang11:
This setting actually caused issues

Before:

https://github.com/holoviz/hvplot/assets/15331990/d4004c88-f90c-42fa-a57d-34b32ae46a4a

After:

https://github.com/holoviz/hvplot/assets/15331990/0ba3b2e0-647c-4591-874d-ae54cd166b73

Closes https://github.com/holoviz/hvplot/issues/1105

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -568,9 +568,6 @@ class HoloViewsConverter:
         if flip_yaxis:
             plot_opts['invert_yaxis'] = True
 
-        if self.geo and not plot_opts.get('data_aspect'):
-            plot_opts['data_aspect'] = 1
-
         ignore_opts = ['responsive', 'aspect', 'data_aspect', 'frame_height', 'frame_width']
         if not any(plot_opts.get(opt) for opt in ignore_opts):
             plot_opts['width'] = plot_opts.get('width', 700)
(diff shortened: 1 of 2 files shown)


maximlt (commented):
Isn't this change working around a bug in HoloViews? It's a pretty bad bug, but it seems like the change is removing a requested feature (https://github.com/holoviz/hvplot/issues/192). We can decide the feature is not worth it, but it needs to be explicit and we need to assess the full extent of its effects.

ahuang11:
I suppose I could dig deeper and see if the underlying bug can be fixed; if not I'd still say this PR is worth merging because it's practically unusable right now.

ahuang11:
I suppose it's not a geo/hvplot specific issue, rather, a HoloViews issue.

Going to close this and if I figure it out, open a PR in holoviews.

```
import holoviews as hv
from holoviews.operation.datashader import rasterize
hv.extension("bokeh")

rasterize(hv.Curve([0, 1, 2])).opts(data_aspect=1)
```

ahuang11:
Resolving here instead: https://github.com/holoviz/holoviews/pull/6189

PR #1314, opened 2024-04-15
Allow algorithm as str input to `downsample` operation

### LINKED ISSUE #1312: Expose more downsample algorithms and options
At the moment, I think, `downsample=True` is the only way to toggle the `downsample1d` and there's no way to customize how it's called, i.e. there's no way to set `algorithm` or some of its other options. Multiple options to implement that:

- Overload `downsample` to accept a string (`minmax`) or a dict (`{'algorithm': 'minmax', **other_downsample1d_opts}`)
- Just allow extra kwargs to be passed to hvPlot and catch them internally, like `max_px` and `threshold` for `dynspread`.
- Probably many other ways

Small preference for the first option.

cc @droumis

droumis:
fixes #1312

This pull request updates the handling of the `downsample` parameter in hvPlot. Previously, the parameter only accepted a boolean value. With these changes, users can now specify the downsampling algorithm directly, provided they have 'tsdownsample' library installed.

```python
"""
downsample (default=False):
        Controls the application of downsampling to the plotted data,
        which is particularly useful for large timeseries datasets to
        reduce the amount of data sent to browser and improve
        visualization performance. Requires HoloViews >= 1.16. Additional
        dependencies: Installing the `tsdownsample` library is required
        for using any downsampling methods other than the default 'lttb'.
        Acceptable values:
        - False: No downsampling is applied.
        - True: Applies downsampling using HoloViews' default algorithm
            (LTTB - Largest Triangle Three Buckets).
        - 'lttb': Explicitly applies the Largest Triangle Three Buckets
          algorithm.
        - 'minmax': Applies the MinMax algorithm, selecting the minimum
          and maximum values in each bin. Requires `tsdownsample`.
        - 'm4': Applies the M4 algorithm, selecting the minimum, maximum,
          first, and last values in each bin. Requires `tsdownsample`.
        - 'minmax-lttb': Combines MinMax and LTTB algorithms for
          downsampling, first applying MinMax to reduce to a preliminary
          set of points, then LTTB for further reduction. Requires
          `tsdownsample`.
        Other string values corresponding to supported algorithms in
        HoloViews may also be used.
"""
```

- [x] add tests
- [x] add docs (already in large timeseries nb)

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -209,19 +209,23 @@ class HoloViewsConverter:
         Controls the application of downsampling to the plotted data,
         which is particularly useful for large timeseries datasets to
         reduce the amount of data sent to browser and improve
-        visualization performance. Requires HoloViews >= 1.16. Acceptable
-        values: - False: No downsampling is applied. - True: Applies
-        downsampling using HoloViews' default algorithm
+        visualization performance. Requires HoloViews >= 1.16. Additional
+        dependencies: Installing the `tsdownsampler` library is required
+        for using any downsampling methods other than the default 'lttb'.
+        Acceptable values:
+        - False: No downsampling is applied.
+        - True: Applies downsampling using HoloViews' default algorithm
             (LTTB - Largest Triangle Three Buckets).
         - 'lttb': Explicitly applies the Largest Triangle Three Buckets
           algorithm.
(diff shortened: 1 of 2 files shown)


maximlt (approved):
LGTM! The day we need to allow configuring more downsample1d we can still overload `downsample` to accept a dict, for now this looks way more than enough. Thanks!

PR #1320, opened 2024-04-18
Format and lint the code base

maximlt:
Flake8 was almost entirely disabled. This PR introduces `ruff` as both the linter and formatter. It is set up with pre-commit. I haven't extended the list of linting rules, but planning to do that in the future (after the next release). I have opted for formatting with single quotes as that was the most used kind of quote in the code base, I'm also fine with double quotes if someone prefers that.


philippjfr:
Looks great, thank you!

PR #1323, opened 2024-04-19
CI: fix the test matrix option on releases

maximlt:
The full tests are not running when making a release. Not sure this is the correct fix, I hope pushing a tag on the main branch won't trigger two events but just the tag one. Let's see.


maximlt:
Damned it 
image

Issue #501, opened 2020-08-14
Memory leak


jmontoyam:
Hello,

first of all, thank you very much for this amazing project! ;).  I think I have detected a possible memory leak. In my use case, I use a function that generates a big xarray DataArray, and another function that receives as input such big xarray DataArray an generates an image from it using hvplot.image(rasterize=True). I use jupyterlab to interact with such functions and visualize the generated image. I have noticed that every time I re-execute the code cell containing both functions, the memory usage keeps growing and growing. Please see the memory usage showed in the attached images (top right corner):

First execution of the code cell:
![execution1](https://user-images.githubusercontent.com/3098532/90255460-9d836d80-de44-11ea-99ec-d5e84c60d4bc.png)

Second execution of the code cell:
![execution2](https://user-images.githubusercontent.com/3098532/90255626-d28fc000-de44-11ea-9d3c-590aef73f411.png)

Third execution of the code cell:
![execution3](https://user-images.githubusercontent.com/3098532/90255664-e20f0900-de44-11ea-9cb7-22d06b3921c1.png)

I tried the suggestion given by @philippjfr in holoviz/holoviews#1821 (%reset out) but it did not work.

Expected behavior: I expected the memory usage to keep constant between re-execution of the same code cell  (the memory occupied but the data generated in the previous execution is supposed to be garbage collected), am I right?, or am I missing something?.

Software info:

hvplot: 0.6.0
holoviews: 1.13.3
datashader: 0.11.0
bokeh: 2.1.1

jupyter core     : 4.6.3
jupyter-notebook : 6.0.3
qtconsole        : 4.7.5
ipython          : 7.16.1
ipykernel        : 5.3.0
jupyter client   : 6.1.3
jupyter lab      : 2.1.5
nbconvert        : 5.6.1
ipywidgets       : 7.5.1
nbformat         : 5.0.7
traitlets        : 4.3.3

JupyterLab v2.1.5
Known labextensions:
        @bokeh/jupyter_bokeh v2.0.2  enabled  OK
        @jupyter-widgets/jupyterlab-manager v2.0.0  enabled  OK
        @pyviz/jupyterlab_pyviz v1.0.4  enabled  OK

Google Chrome
Version 84.0.4147.125 (Official Build) (64-bit)

OS
Ubuntu 18.04.4 LTS

Thank you very much for all your help! ;)

jmakov:
[...]

PR #1350, opened 2024-06-04
Optimize handling of wide datasets and Pandas indexes 

maximlt:
By avoiding calling `redim` on each element of an `NdOverlay` which meant that every element had its own copy of the data

---

Edit: This PR bundles a few optimizations:

- https://github.com/holoviz/hvplot/pull/1350/commits/1e61065d041db889dbfed2634e5918e07295dab2: as described above, `.redim` is no longer called the elements of an `NdOverlay` so they can share the same data. This is demonstrated with the example below, the `data` wrapped by each element of the `NdOverlay` is now the same object.

```python
import hvplot.pandas  # noqa
from bokeh.sampledata.degrees import data as deg

plot = deg.hvplot.line(x='Year', y=['Art and Performance', 'Business', 'Biology', 'Education', 'Computer Science'])
plot
```
image

```python
plot['Business'].data is plot['Biology'].data  # True now :)
```

- https://github.com/holoviz/hvplot/pull/1350/commits/326ee64e06ebc721eed20dc92696611137cf25c8: HoloViews added in https://github.com/holoviz/holoviews/pull/6061 support for retaining `Pandas` indexes, which means that hvPlot no longer needs to call `reset_index` in a few cases, avoiding creating internal copies of the data. This change was made in https://github.com/holoviz/hvplot/pull/1350/commits/326ee64e06ebc721eed20dc92696611137cf25c8, by skipping some `reset_index` calls when the type of the dataset is a Pandas DataFrame (so far the Pandas Interface in HoloViews is the only one to have this support, so not Dask, not Geopandas, etc. which still require the `reset_index` calls). The preceding commits add various tests to make sure all the `reset_index` calls are at least called once in the unit tests. This commit https://github.com/holoviz/hvplot/pull/1350/commits/adbddf7f0c6890988da0fa390756c89a928e0d14 updates the instantiation of a `Dataset` object with an explicit list of kdims instead letting HoloViews infer them, as HoloViews when given a Pandas DataFrame doesn't include its indexes in the list of automatically inferred kdims.

- https://github.com/holoviz/hvplot/pull/1350/commits/abb485a0b253f3c8428f96af9af39b7d819db2ae: Only redim and/or relabel a HoloViews element when needed, as suggested in https://github.com/holoviz/hvplot/pull/1350#discussion_r1644152459 since apparently these operations return a clone even when not given any argument.

I hope these optimizations will help/fix this issue https://github.com/holoviz/hvplot/issues/501.


droumis:
issue noted in meeting: "When printing the ndoverlay, the vdim is not a value label.. it's the label of the last element in the ndoverlay"

philippjfr:
With https://github.com/holoviz/holoviews/pull/6262 this should now behave well.

maximlt:
> When printing the ndoverlay, the vdim is not a value label.. it's the label of the last element in the ndoverlay

I confirm this is fixed, i.e. printing `out` from the code below:

```python
import numpy as np
import pandas as pd
import holoviews as hv
hv.extension('bokeh')

df = pd.DataFrame(np.random.random((5, 4)), columns=list('ABCD'))

value_label = 'value'
group_label = 'Variable'

charts = []
for column in df.columns:
    kdims = ['index']
    vdims = hv.Dimension(column, label=value_label)
    chart = hv.Curve(df, kdims, vdims)
    charts.append((column, chart))

out = hv.NdOverlay(charts, group_label, sort=False)
out
```

now displays:

```
:NdOverlay   [Variable]
   :Curve   [index]   (value)
```

hvplot/converter.py
```diff
@@ -2123,8 +2123,9 @@ def chart(self, element, x, y, data=None):
 
         charts = []
         for c in y:
-            kdims, vdims = self._get_dimensions([x], [c])
-            chart = element(data, kdims, vdims).redim(**{c: self.value_label})
+            ydim = hv.Dimension(c, label=self.value_label)
+            kdims, vdims = self._get_dimensions([x], [ydim])
+            chart = element(data, kdims, vdims)
             charts.append((c, chart.relabel(**self._relabel).redim(**self._redim)))
```
philippjfr, line 2129: Maybe let's also skip `relabel` and `redim` if `self._relabel` and `self._redim` are empty respectively.

droumis:
ready to merge?

maximlt:
OP updated with a description of the changes made https://github.com/holoviz/hvplot/pull/1350#issue-2333794995.

philippjfr (approved):

philippjfr:
Great work, all looked good to me and thanks for the new tests! Let's test this as much as possible, would love to see a new alpha release for that purpose.

PR #1352, opened 2024-06-12
do not mutate data

ahuang11:
I noticed that it previously mutated the data (e.g. added a `label` column).

```python
import hvplot.pandas
import pandas as pd

df = pd.DataFrame(
    {
        "City": ["Buenos Aires", "Brasilia", "Santiago", "Bogota", "Caracas"],
        "Country": ["Argentina", "Brazil", "Chile", "Colombia", "Venezuela"],
        "Latitude": [-34.58, -15.78, -33.45, 4.60, 10.48],
        "Longitude": [-58.66, -47.91, -70.66, -74.08, -66.86],
    }
)
df.hvplot.points("Longitude", "Latitude", by="City") * df.hvplot.labels(
    "Longitude", "Latitude", text="City ({Latitude:.1f}N, {Longitude:.1f}E)", by="City"
)
df
```

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -2566,6 +2566,7 @@ class HoloViewsConverter:
         self._error_if_unavailable('labels')
         self.use_index = False
         data, x, y = self._process_chart_args(data, x, y, single_y=True)
+        data = data.copy()
 
         text = self.kwds.get('text')
         if not text:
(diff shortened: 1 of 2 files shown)


hvplot/converter.py
```diff
@@ -2566,6 +2566,7 @@ def labels(self, x=None, y=None, data=None):
         self._error_if_unavailable('labels')
         self.use_index = False
         data, x, y = self._process_chart_args(data, x, y, single_y=True)
+        data = data.copy()
```
maximlt, line 2569: Is there a HoloViews-y way to avoid making a copy? That would be the first call to `copy()` in the converter.

If not, I'd suggest making the copy only in the `elif text not in data.columns` block.

ahuang11: Not entirely sure; I just put it in the block.

philippjfr (approved):
I think this is the correct thing to do.

maximlt:
Unrelated test failure.

Issue #1147, opened 2023-09-20
`paths` is not documented in hvplot reference gallery


ahuang11:
I don't see it here: https://hvplot.holoviz.org/reference/index.html#xarray
```
import pandas as pd

df = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})

df.hvplot.paths("a", "b")
```

PR #1354, opened 2024-06-13
Add paths reference notebook

ahuang11:
Closes https://github.com/holoviz/hvplot/issues/1147

image

### DIFF
--- a/doc/reference/pandas/paths.ipynb
+++ b/doc/reference/pandas/paths.ipynb
@@ -1,0 +1,80 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Paths"
+   ]
+  },
+  {
+   "cell_type": "code",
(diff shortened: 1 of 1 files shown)


maximlt (commented):
At the moment the geo-specific plots are under the `Geopandas` section. It's certainly not optimal since Geopandas is not a strict requirement to create these plots. Yet, I'm not sure it's a good idea to divert from the current approach by adding the `Paths` guide under the `Tabular` section?

image

ahuang11:
The other two use a geopandas dataframe and directly call hvplot(). I also didn't have a Geopandas Paths dataframe readily available.
image

maximlt (approved):

maximlt:
@ahuang11 I had forgotten this step but when we add a new gallery page we need to upload the thumbnail to S3 (here assets.holoviz.org/hvplot/thumbnails/reference/tabular/paths.png). I've just done that.

Issue #1232, opened 2023-12-21
No output in jupyter.


LinuxIsCool:
#### ALL software version info
bokeh==3.3.2
holoviews==1.18.1
hvplot==0.9.0
jupyterlab==4.0.9
jupyterlab-widgets==3.0.9
jupyterlab_pygments==0.3.0
jupyterlab_server==2.25.2
panel==1.3.6

Python 3.10.12

Linux Kernel: 6.6.6-76060606-generic

#### Description of expected behavior and the observed behavior

Expecting hvplot to work in jupyter...

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import pandas as pd
import hvplot.pandas
import numpy as np

df = pd.DataFrame({'x':range(100), 'y': np.linspace(0,100,100)})

df.hvplot.line(x='x',y='y')

```

#### Stack traceback and/or browser JavaScript console output
In the web console in jupyter:
```
TypeError: Bokeh is undefined
Panel: ERROR: Unable to run Panel code because Bokeh or Panel library is missing
```

#### Screenshots or screencasts of the bug in action
No output: 

![image](https://github.com/holoviz/hvplot/assets/31582215/e81b4208-18ee-4f05-96c7-6741117c9468)

However, if I refresh the web page on jupyter then I now see output: 

![image](https://github.com/holoviz/hvplot/assets/31582215/49da5c2e-6c13-4991-96cc-cbc495e2528e)

LinuxIsCool:
I downgraded from holoviews 1.18.1 to 1.18.0 and bokeh from bokeh from 3.3.2 to 3.3.1 and things are currently working.

ea42gh:
I have the same problem, but downgrading did not resolve the issue for me.

Refreshing the page appears to work occasionally.
When it does (I do not know what triggers it?!)
plotting hv elements in panels works for a while,
but eventually disappear...

Hoxbro:
It is likely some old js cache that messes things up. 

Try the following steps.
1) Clear the notebook and save.
2) Close all the notebooks.
3) Hard refresh the browser.

maximlt:
Hi @ea42gh could you report whether following the steps suggest by @Hoxbro fixed your issue? I agree it's likely an issue with some cache. If you're using JupyterLab, you may also want to make sure you don't have multiple notebooks opened at the same time that run with various versions of Bokeh.

LinuxIsCool:
@Hoxbro What do you mean by Clear the notebook and save?

[...]

PR #1359, opened 2024-06-20
Ensure hvplot.<ext> hooks are run on every import

philippjfr:
The `hvplot.<ext>` import mechanism is a convenient way to allow users to have to avoid running the holoviews/panel extensions. However since imports are cached only the first import actually embeds the extension JS code, meaning that if you re-run the cell(s) containing `import hvplot.pandas` (or some other integration) then the JS will no longer be available and on subsequent reloads/re-runs of the notebook plots may not appear. This probably accounts for most issues encountered and described in https://github.com/holoviz/hvplot/issues/1232.

Here we add an IPython hook which simply deletes the modules before every cell execution. This is a big hammer but the best @hoxbro and I could find and at least it is restricted to IPython environments.


hoxbro:
We need to update the logic to only register hvplot extensions once, have updated it for some but we should update it for all. 

Screenshot 2024-06-20 14 44 48

hvplot/__init__.py
```diff
@@ -116,6 +118,25 @@
             # the package.
             __version__ = '0.0.0+unknown'
 
+_extensions = {
```
hoxbro, line 121: We should update the rest of these.

maximlt (commented):
This is indeed quite a hammer but it's fixing a pretty bad issue (although it's hard to tell if it's the source of the majority of the reported problems). Not sure whether we could abuse Python's import system to achieve the same behavior, that would anyway be a big hammer too, and the one suggested at least scopes it to IPython usage.

I'd appreciate some comments in the code that explain why we have to resort to running code on each cell execution and removing modules from `sys.modules` 🙃 

Isn't this a good example of why relying on imports with side-effects is a bad idea in Python? Should we consider migrating to a more explicit API, even if it'd mean having to type a bit more?

While this code change improves the situation, I still think there's a major UX issue. We're injecting crucial JS/CSS in the notebook, that can be easily removed by users by mistake, without letting them know that this is wrong. This is a general issue through HoloViz packages that needs `.extension()` to be called (at least HoloViews and GeoViews display some logos). I would be in favor for showing a very explicit message. This might be a good topic for the next HoloViz meeting.

maximlt (commented):
I think that as it stands this PR leads to `import hvplot.<ext>` fully loading the extension (JS/CSS) on every call. I don't think it's too bad, just noting this behavior change.

image

maximlt:
We need to look more into how this is going to affect Polars' users.

maximlt:
@hoxbro can you please review?

hvplot/__init__.py
```diff
@@ -116,6 +118,28 @@
...
+# The hvplot.<ext> import mechanism is a convenient way to allow users to have
+# to avoid running the holoviews/panel extensions. However since imports are
+# cached only the first import actually embeds the extension JS code, meaning
+# that if you re-run the cell(s) containing import hvplot.pandas (or some other
+# integration) then the JS will no longer be available and on subsequent
+# reloads/re-runs of the notebook plots may not appear.
+# Here we add an IPython hook which simply deletes the modules before every
+# cell execution. This is a big hammer but at least it is restricted to
+# IPython environments.
+_extensions = set()
```
hoxbro, line 130: Maybe another name for this variable? It can cause confusion between `pn.extension` / `hv.extension`. 

Some sugestions could be `_data_extensions`, `_module_extensions`, or `_data_backends`.

hvplot/__init__.py
```diff
@@ -213,8 +237,7 @@ def help(kind=None, docstring=True, generic=True, style=True):
 
 
 def post_patch(extension='bokeh', logo=False):
-    if extension and not getattr(_hv.extension, '_loaded', False):
-        hvplot_extension(extension, logo=logo)
+    hvplot_extension(extension, logo=logo)
```
hoxbro, line 240: We could have an argument that restores the old behavior, but I don't have a strong opinion.

hoxbro (approved):
A few comments, you can ignore them if you don't agree :+1:

hoxbro (approved):

michaelaye:
oh my word, this explains so much, thanks for hunting it down...

PR #1362, opened 2024-06-29
Replace geopandas.datasets by geodatasets in the examples

maximlt:
Since `geopandas.datasets` has been removed in GeoPandas 1.0.


maximlt:
And no longer upgrade pip/setuptools, see https://github.com/actions/setup-python/issues/791.

PR #1367, opened 2024-07-09
Fix geopandas type check

ahuang11:
If geometry is None, it crashes:
```python
import hvplot.pandas
import censusdis.data as ced
from censusdis.datasets import ACS5
from censusdis.counties.washington import KING
from censusdis.states import WA

GROUP = "B25008"

gdf_king = ced.download(
    ACS5,
    2020,
    # Instead of listing the variables, we can ask for a whole group.
    group=GROUP,
    state=WA,
    county=KING,
    tract="*",
    with_geometry=True,
    # An optional flag that produces better maps in coastal areas.
    # More on this in the next lesson.
    remove_water=True,
)
gdf_king
```

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -1012,6 +1012,8 @@ class HoloViewsConverter:
             if kind is None:
                 if datatype == 'geopandas':
                     geom_types = {gt[5:] if gt and 'Multi' in gt else gt for gt in data.geom_type}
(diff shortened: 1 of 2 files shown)


maximlt (approved):
In itself the change looks good to me. I'd be curious to know how geopandas' `.plot` behaves in the same situation.

PR #1368, opened 2024-07-11
Support cticks

ahuang11:
I tried very long time to get the matplotlib test to work, but couldn't :/

If I set hv.extension("matplotlib"), this works; but without that extension, it doesn't detect cbar weirdly.

```python
import pandas as pd
import holoviews as hv
import hvplot.xarray
import numpy as np
import xarray as xr

backend = "matplotlib"
da2 = xr.DataArray(
    data=np.arange(27).reshape((3, 3, 3)),
    coords={"y": [0, 1, 2], "x": [0, 1, 2]},
    dims=["y", "x", "other"],
    name="test2",
)

plot = da2.isel(other=2).hvplot(cticks=[5, 10])
handles = hv.renderer(backend).get_plot(plot).handles
if backend == 'matplotlib':
    print(handles)
    assert handles["cbar"].get_ticks().tolist() == [5, 10]
```


hvplot/converter.py
```diff
@@ -404,7 +404,7 @@ class HoloViewsConverter:
         'colormap',
         'fontsize',
         'c',
-        'cmap',
+        'cticks' 'cmap',
```
ahuang11, line 407:
```suggestion
        'cticks',
        'cmap',
```

hvplot/converter.py
```diff
@@ -552,6 +553,7 @@ def __init__(
         invert=False,
         stacked=False,
         colorbar=None,
+        cticks=None,
```
maximlt, line 556: I think `cticks` should be added to the docstring and relevant page in the docs. Adding an example *somewhere* would also be nice.

hvplot/converter.py
```diff
@@ -209,8 +209,8 @@ class HoloViewsConverter:
         Axis labels for the x-axis, y-axis, and colorbar
     xlim/ylim (default=None): tuple or list
         Plot limits of the x- and y-axis
-    xticks/yticks (default=None): int or list
-        Ticks along x- and y-axis specified as an integer, list of
+    xticks/yticks/cticks (default=None): int or list
```
maximlt, line 212: Minor update, reflecting the changes made to the Customization guide to the converter docstring.

maximlt (approved):

Issue #1375, opened 2024-07-19
Automatically or manually toggle to automatically use lon_lat_to_easting_northing


ahuang11:
Sometimes, I don't want to use geoviews because it's extra dependency and import (and sometimes [it breaks](https://github.com/holoviz/hvplot/issues/1374)):

```python
import hvplot.xarray
import xarray as xr
import holoviews as hv
from holoviews.util.transform import lon_lat_to_easting_northing

ds = xr.tutorial.load_dataset("air_temperature").isel(time=slice(0, 100))
ds["easting"], ds["northing"] = lon_lat_to_easting_northing(ds["lon"], ds["lat"])
ds = ds.swap_dims({"lon": "easting", "lat": "northing"})
hv.element.tiles.CartoDark() * ds.hvplot("easting", "northing")
```
<img width="1081" alt="image" src="https://github.com/user-attachments/assets/18b53a35-6a59-47f5-9ef0-74cc7872d3d4">

I'm wondering whether there should be a kwarg to transform so the above can be simplified down to

```python
import hvplot.xarray
import xarray as xr

ds = xr.tutorial.load_dataset("air_temperature").isel(time=slice(0, 100))
ds.hvplot("lon", "lat", tiles=True, to_easting_northing=True)  # or some other kwarg
```

Or, if it detects the x is between -180, 360 and y is between -90, 90 without any crs/projection kwargs, automatically convert to easting/northing?

```python
import hvplot.xarray
import xarray as xr

ds = xr.tutorial.load_dataset("air_temperature").isel(time=slice(0, 100))
# automatic detection
ds.hvplot("lon", "lat", tiles=True)
```

jbednar:
I'd be ok with assuming lat,lon inputs with `geo=True` and projecting them using our own internal functions from HoloViews without having to bring in geoviews/cartopy/proj for that one common case. I would not be ok with checking the ranges and automatically converting; at best that could be a warning (maybe people really _do_ want to look at some tiny area of the earth, or vice versa!).

ahuang11:
Okay I think a warning is probably good, because without it, the data is centered around 0 and that can be confusing.

<img width="1025" alt="image" src="https://github.com/user-attachments/assets/36428535-5731-424b-88ea-d2ad7eb1b320">

PR #1377, opened 2024-07-20
Automatically convert lonlat to xy when tiles=True

ahuang11:
Closes https://github.com/holoviz/hvplot/issues/1375

Will add tests next week if this logic is ok.

Basically if user sets `geo=True` without crs/projection, it does a `pseudo-geo` (i.e. perform projection and toggle off)

image

### DIFF
--- a/doc/user_guide/Geographic_Data.ipynb
+++ b/doc/user_guide/Geographic_Data.ipynb
@@ -60,7 +60,7 @@
    "source": [
     "## Tiled web map\n",
     "\n",
-    "A [tiled web map](https://en.wikipedia.org/wiki/Tiled_web_map), or tile map, is an interactive map displayed on a web browser that is divided into small, pre-rendered image tiles, allowing for efficient loading and seamless navigation across various zoom levels and geographic areas. hvPlot allows to add a tile map as a basemap to a plot with the `tiles` parameter. Importantly, `tiles` is a parameter that can be used **without installing GeoViews if the data is already projected in Web Mercator**.\n",
+    "A [tiled web map](https://en.wikipedia.org/wiki/Tiled_web_map), or tile map, is an interactive map displayed on a web browser that is divided into small, pre-rendered image tiles, allowing for efficient loading and seamless navigation across various zoom levels and geographic areas. hvPlot allows to add a tile map as a basemap to a plot with the `tiles` parameter. Importantly, `tiles` is a parameter that can be used **without installing GeoViews**.\n",
     "\n",
     "We'll display this dataframe of all US airports (including military bases overseas), the points are expressed in latitude/longitude coordinates:"
    ]
@@ -78,7 +78,43 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We'll first start by displaying the airports **without GeoViews**. To do so, we must convert the coordinates to Web Mercator, which can be easily achieved using the `lon_lat_to_easting_northing` function provided by HoloViews, that doesn't require installing any of the usual geo dependencies like `pyproj` or `cartopy`."
+    "We'll first start by displaying the airports **without GeoViews** with tiles by setting `tiles=True`. \n",
+    "\n",
+    "Under the hood, hvPlot projects lat/lon to easting/northing (EPSG:4326 to EPSG:3857) coordinates without additional package dependencies if it detects that the values falls within expected lat/lon ranges.\n",
+    "\n",
+    "Note, **this feature is only available after `hvplot>0.10.0`**; older versions, `hvplot<=0.10.0`, require manual projection (see below)."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "airports.hvplot.points('Longitude', 'Latitude', tiles=True, color='red', alpha=0.2)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "If you do not want it to be auto-projected under the hood for any reason, set `projection=False`, **but note it will not be properly placed on the tiles** without manual intervention. In doing so, the data will be plotted around the [null island location](https://en.wikipedia.org/wiki/Null_Island), roughly 600km off the coast of West Africa. `padding` was added to demonstrate this."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "airports.hvplot.points('Longitude', 'Latitude', tiles=True, projection=False, color='red', alpha=0.2, padding=10000)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "To manually project use the `lon_lat_to_easting_northing` function provided by HoloViews, that doesn't require installing any of the usual geo dependencies like `pyproj` or `cartopy`."
    ]
   },
   {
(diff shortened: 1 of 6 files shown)


ahuang11:
> you assigned me as a reviewer but this is still marked as Draft

> Will add tests next week if this logic is ok.

Wanted to first see if you're okay with it because of

```python
            min_x = np.min(data[x])
            max_x = np.max(data[x])
            min_y = np.min(data[y])
            max_y = np.max(data[y])
```

> seems to differ from what was discussed in https://github.com/holoviz/hvplot/issues/1375 (no warning, automatic conversion).

I believe it matches what Jim proposed:

> I'd be ok with assuming lat,lon inputs with geo=True and projecting them using our own internal functions from HoloViews without having to bring in geoviews/cartopy/proj for that one common case.

maximlt (commented):
Overall I'm very supportive of this idea! Happy we're working on making it easier to display lat/lon data. It would have made updating some of the HoloViz examples even easier, for instance https://examples.holoviz.org/gallery/gull_tracking/gull_tracking.html.

(And that's for a future iteration but I'm sure we can also streamline plotting/converting GeoDataFrame objects)

---

> I believe it matches what Jim proposed

Yep true sorry, I got confused by his second sentence, since in the PR the ranges are checked, except that it's done only when `geo=True` so that seems to fit with what he described. Pinging him to get his feedback anyway since he had opinions @jbednar.

> I'd be ok with assuming lat,lon inputs with geo=True and projecting them using our own internal functions from HoloViews without having to bring in geoviews/cartopy/proj for that one common case. I would not be ok with checking the ranges and automatically converting; at best that could be a warning (maybe people really do want to look at some tiny area of the earth, or vice versa!).

---

> Wanted to first see if you're okay with it because of

I'm ok with how the ranges are computed. However, I'm not sure this should be computed there, right at the beginning of the converters' `__init__` method?

- Isn't there code that automatically infer `x` and `y`?
- There's `self.source_data = data` further down the processing pipeline that is a handle on the original data. It is later used to create an `hv.Dataset` to update the element's `_dataset` attribute with `obj._dataset = dataset`. Unfortunately, I know that removing this line doesn't break any unit test. I vaguely remember @philippjfr mentioned it had something to do with streams. So I'd recommend checking with him where the conversion should take place, and if this going to affect that part of the code.

---

> Basically if user sets `geo=True` without crs/projection

Cannot we enable the automatic conversion when users set `tiles=True`, without requiring `geo=True`?

- In my mental model, `geo=True` means that GeoViews will be involved, so this change as is would divert from that.
- Users who set `tiles=True` want a geo-plot by definition. We let them not set `geo=True`, and don't set it internally, as a way to avoid bringing in GeoViews. So we've already set a precedent there, and I think we should pursue in the same direction with checking the ranges and converting when just `tiles=True` is passed (so that's against what @jbednar said). For the small proportion of people that will want to look at that tiny area of the earth, they could disable it with `projection=False`.

ahuang11:
Haha I love that there's both a `self.data_source` and `self.source_data`.

```python
        # Validate DataSource
        self.data_source = data

        self.is_series = is_series(data)
        if self.is_series:
            data = data.to_frame()
        if is_intake(data):
            data = process_intake(data, use_dask or persist)
        # Pandas interface in HoloViews doesn't accept non-string columns.
        # The converter stores a reference to the source data to
        # update the `_dataset` property (of the hv object its __call__ method
        # returns) with a hv Dataset created from the source data, which
        # is done for optimizating some operations in HoloViews.
        data = _convert_col_names_to_str(data)

        self.source_data = data
```

> Users who set tiles=True want a geo-plot by definition. We let them not set geo=True, and don't set it internally, as a way to avoid bringing in GeoViews. So we've already set a precedent there, and I think we should pursue in the same direction with checking the ranges and converting when just tiles=True is passed (so that's against what @jbednar said). For the small proportion of people that will want to look at that tiny area of the earth, they could disable it with projection=False.

I support this idea; it's a better default than having the user to figure out to use tiles, additionally need `geo=True`.

ahuang11:
Wondering if I should rename x,y.
image

Separately, I think hover needs fixing, but I think it's better for it to be implemented in HoloViews; adapted from GeoViews https://github.com/holoviz/geoviews/blob/8a14f9c71298c86e1b364c18ac952bf310e002ac/geoviews/plotting/bokeh/__init__.py#L162

```python

            from bokeh.models import HoverTool
            from bokeh.models import CustomJSHover
            
            hover = HoverTool()
            xdim = x if x in data else self.x
            ydim = y if y in data else self.y
            formatters, tooltips = dict(hover.formatters), []
            xhover = CustomJSHover(code=_hover_code % 0)
            yhover = CustomJSHover(code=_hover_code % 1)
            for name, formatter in hover.tooltips:
                customjs = None
                if formatter in (f'@{{{xdim}}}', '$x'):
                    dim = xdim
                    formatter = '$x'
                    customjs = xhover
                elif formatter in (f'@{{{ydim}}}', '$y'):
                    dim = ydim
                    formatter = '$y'
                    customjs = yhover
                if customjs:
                    key = formatter if formatter in ('$x', '$y') else dim
                    formatters[key] = customjs
                    formatter += '{custom}'
                tooltips.append((name, formatter))
            hover.tooltips = tooltips
            hover.formatters = formatters
            self._plot_opts['tools'] = [hover]
```

hvplot/converter.py
```diff
@@ -2637,6 +2625,33 @@ def _process_gridded_args(self, data, x, y, z):
             not_found = [dim for dim in dimensions if dim not in self.variables]
             _, data = process_derived_datetime_pandas(data, not_found, self.indexes)
 
+        print(self.tiles, self.output_projection)
+        if self.tiles and self.output_projection is not False:
```
maximlt, line 2629: Should this only be done for gridded datasets? I would hope this would also be done for tabular datasets, like in https://examples.holoviz.org/gallery/gull_tracking/gull_tracking.html

ahuang11: Ah you're right. Was only thinking of gridded to maps only.

doc/user_guide/Geographic_Data.ipynb
```diff
@@ -415,7 +417,7 @@
     "- `global_extent` (default=False): Whether to expand the plot extent to span the whole globe\n",
     "- `project` (default=False): Whether to project the data before plotting (adds initial overhead but avoids projecting data when plot is dynamically updated)\n",
     "- `projection` (default=None): Coordinate reference system of the plot (output projection) specified as a string or integer EPSG code, a CRS or Proj pyproj object, a Cartopy CRS object or class name, a WKT string, or a proj.4 string. Defaults to PlateCarree.\n",
-    "- `tiles` (default=False): Whether to overlay the plot on a tile source. Accept the following values:\n",
+    "- `tiles` (default=False): Whether to overlay the plot on a tile source. If coordinate values fall within lat/lon bounds, auto-projects to EPSG:3857 unless `projection=False`. Accepts the following values:\n",
```
ahuang11, line 420: Thoughts on removing this cell and only link directly to Customization geographic section to reduce redundancy?

ahuang11:
Maybe just use 

https://github.com/holoviz/geoviews/blob/main/geoviews/plotting/bokeh/plot.py#L81-L84

jbednar:
I'm happy with the proposal to project only if `tiles` is True, not `geo`; this projection is explicitly about getting things in shape for Bokeh using Web Mercator, and not just generally when working with geo data (where the data can be left unprojected).

doc/user_guide/Geographic_Data.ipynb
maximlt: I prefer to keep them here for now, I think it's nice to have all the options there within the appropriate context.

hvplot/converter.py
```diff
@@ -2092,6 +2096,40 @@ def _process_chart_args(self, data, x, y, single_y=False, categories=None):
...
+        Tiles without requiring geoviews/cartopy.
+        """
+        if self.geo or not self.tiles or self.output_projection is False:
+            return data, x, y
+        elif not is_geodataframe(data) and (x is None or y is None):
+            return data, x, y
+
+        if is_geodataframe(data):
+            if data.crs is not None:
+                data = data.to_crs(epsg=3857)
```
maximlt, line 2113: Nice!

maximlt (approved):

PR #1383, opened 2024-07-29
Conditionally skip testing the Gridded data user guide

maximlt:
Because of https://github.com/pydata/xarray/pull/9182 which has been merged but not yet released.


maximlt:
Xarray was released with the fix (https://github.com/pydata/xarray/releases/tag/v2024.07.0) but merging this anyway to avoid hitting this issue in the future.

PR #1385, opened 2024-07-30
Add hover_tooltips, hover_formatters, hover_mode

ahuang11:
Adds opts from https://github.com/holoviz/holoviews/pull/6180/

Don't think a test is required since it's just adding plot_opts?


droumis (approved):
LGTM, but as a monitor rather than a maintainer for hvPlot, I'm not sure if my approval is sufficient to merge, so you might want to wait for Maxime to review or ask another maintainer.

hvplot/converter.py
```diff
@@ -148,6 +148,12 @@ class HoloViewsConverter:
     hover_cols (default=[]): list or str
         Additional columns to add to the hover tool or 'all' which will
         includes all columns (including indexes if use_index is True).
+    hover_tooltips list[str] or list[tuple]:
+        A list of dimensions to be displayed in the hover tooltip.
+    hover_formatters:
+        A dict of formatting options for the hover tooltip.
+    hover_mode (default='mouse'):
+        The hover mode determines how the hover tool is activated.
```
hoxbro, line 156: Maybe have a link to Bokeh and/or HoloViews documentation. 

For `hover_mode` maybe add all the options that are available.

hoxbro (approved):

maximlt:
I've got a few comments.

Doesn't `hover_mode` duplicate what is already possible with setting `hover` to e.g. `'vline'`? This was added in https://github.com/holoviz/hvplot/pull/493 but never documented I believe.

---

> Don't think a test is required since it's just adding plot_opts?

Adding tests is pretty much always worth in my opinion, in particular when they're easy to write.

---

I would suggest alphabetically ordering the new `hover_*` parameters in the signature, docstring, and doc.

---

We need to find a way to better document these sorts of options. As it stands, I don't think the docs provide sufficient information for users to leverage these new parameters.

ahuang11:
Filed as an issue for follow up
https://github.com/holoviz/hvplot/issues/1394

Issue #1100, opened 2023-07-06
shared_axes=False not working for hvplot.hist()


JanHomann:
## Versions
bokeh         3.1.1
holoviews     1.16.2
hvplot        0.8.4

## Issue
The plotting parameter `shared_axes=False` in the function `.hvplot.hist()`, when using `subplots=True`, does not lead to a rescaling of the x-axis for the individual subplots to accommodate the data. Instead the plots keeps having the same x-scale. Other plot types seem to work, so the issue seems specific to `.hist()`.

## Code

```
import hvplot.pandas
from hvplot.sample_data import us_crime

us_crime.hvplot.hist(subplots=True, shared_axes=False,
                     width=300, height=200)

```

## Screenshot of output of code demostrating the problematic behavior.
Each plot is scaled equally, and most end up having only one bin with data.

![171f6cc1840f74d65f96411045e88c96c9f8a6a9](https://github.com/holoviz/hvplot/assets/35338267/81b5df72-2efd-4e94-ab0d-c22457200395)

heikone:
Can confirm this behavior.
Just started to use hvplot and exactly this was one of my first plots... 🫣
Would appreciate a solution or a workaround.

Greetings,
Heiko

xuJ14:
Any updates?

ahuang11:
Thanks for bumping; I think I have a fix in https://github.com/holoviz/hvplot/pull/1391

xuJ14:
Thank you!

PR #1391, opened 2024-08-07
Do not set bin_range if shared_axes=False

ahuang11:
Closes https://github.com/holoviz/hvplot/issues/1100
image

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -2380,9 +2380,10 @@ class HoloViewsConverter:
                 .opts(compat_opts, backend=self._backend_compat)
             )
 
-        ranges = []
-        for col in y:
-            if 'bin_range' not in self.kwds:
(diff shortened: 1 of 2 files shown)


hvplot/converter.py
```diff
@@ -2380,18 +2380,19 @@ def hist(self, x=None, y=None, data=None):
                 .opts(compat_opts, backend=self._backend_compat)
             )
 
-        ranges = []
-        for col in y:
-            if 'bin_range' not in self.kwds:
+        if 'bin_range' not in self.kwds and 'axiswise' not in self._norm_opts:
```
philippjfr, line 2383:
```suggestion
        if 'bin_range' not in self.kwds and not self._norm_opts.get('axiswise', False):
```

philippjfr (approved):

PR #1396, opened 2024-08-21
Added daskgeopandas check to `is_geodataframe` function

Azaya89:
- [x] Modified the `is_geodataframe` function to support `DaskGeoDataFrame`

- [x] Test added.

### DIFF
--- a/hvplot/tests/testutil.py
+++ b/hvplot/tests/testutil.py
@@ -6,6 +6,8 @@ import numpy as np
 import pandas as pd
 import panel as pn
 import pytest
+import spatialpandas
+import dask
 
 from unittest import TestCase, SkipTest
 
@@ -16,6 +18,7 @@ from hvplot.util import (
     process_xarray,
     _convert_col_names_to_str,
     instantiate_crs_str,
+    is_geodataframe,
 )
 
 
@@ -184,7 +187,7 @@ class TestProcessXarray(TestCase):
 
 class TestDynamicArgs(TestCase):
     def test_dynamic_and_static(self):
-        from ..util import process_dynamic_args
+        from hvplot.util import process_dynamic_args
 
         x = 'sepal_width'
         y = pn.widgets.Select(
@@ -198,7 +201,7 @@ class TestDynamicArgs(TestCase):
         assert arg_deps == []
 
     def test_dynamic_kwds(self):
-        from ..util import process_dynamic_args
+        from hvplot.util import process_dynamic_args
 
         x = 'sepal_length'
         y = 'sepal_width'
@@ -211,7 +214,7 @@ class TestDynamicArgs(TestCase):
         assert arg_deps == []
 
     def test_fn_kwds(self):
-        from ..util import process_dynamic_args
+        from hvplot.util import process_dynamic_args
 
         x = 'sepal_length'
         y = 'sepal_width'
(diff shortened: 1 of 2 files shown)


hvplot/tests/testutil.py
```diff
@@ -7,6 +7,13 @@
 import panel as pn
 import pytest
 
+try:
+    import spatialpandas
+    import dask
+except ImportError:
+    spatialpandas = None
+    dask = None
```
ahuang11, line 15: I think you might want to add spatialpandas/dask to the tests requirements.
https://github.com/holoviz/hvplot/blob/main/pyproject.toml#L73-L80

ahuang11:
Looks good to me! Just a minor addition.

hvplot/tests/testutil.py
Azaya89: Done

hvplot/tests/testutil.py
```diff
@@ -361,3 +369,22 @@ def test_instantiate_crs_str_kwargs():
...
     assert isinstance(crs.globe, ccrs.Globe)
     assert crs.globe.datum == 'WGS84'
+
+
+@pytest.mark.skipif(
+    spatialpandas is None or dask is None, reason='spatialpandas or dask is not available'
+)
+def test_is_geodataframe_spatialpandas_dask():
+    import dask.dataframe as dd
+    import spatialpandas as spd
```
ahuang11, line 379: One other thought are these can be moved up, and the pytest.mark.skipif can check if dd or spd is None.
```suggestion
```

hvplot/util.py
```diff
@@ -473,8 +473,9 @@ def process_intake(data, use_dask):
 def is_geodataframe(data):
     if 'spatialpandas' in sys.modules:
         import spatialpandas as spd
+        from spatialpandas.dask import DaskGeoDataFrame
```
ahuang11, line 476: Does spatialpandas always include dask as dep? If not, this might crash.

hvplot/tests/testutil.py
Azaya89: OK. Done

hvplot/util.py
Azaya89: > Maybe you can just use `spd.dask.DaskGeoDataFrame` instead of another import for consistency, but no strong opinion.

Yeah, I think that is better.

hvplot/tests/testutil.py
```diff
@@ -6,6 +6,15 @@
...
 import pytest
+import dask.dataframe as dd
+import spatialpandas as spd
+
+try:
+    import spatialpandas
+    import dask
+except ImportError:
+    spatialpandas = None
+    dask = None
```
ahuang11, line 17: Sorry for the late re-review. I think these should be consolidated into a single try/except, or else it'll crash on import if dask/spatialpandas is unavailable.
```suggestion
import dask.dataframe as dd
import spatialpandas as spd

try:
    import spatialpandas
    import dask
except ImportError:
    spatialpandas = None
    dask = None
```

Azaya89: Done!

ahuang11 (approved):

ahuang11:
Thank you looks good now!

Issue #1406, opened 2024-09-04
hvplot Explorer is giving nuisance Line plot when using over 10,000 points


hagaishalevaei:
#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc)
Windows 11, 
Python=3.11.9,
 Jupyterlab  4.2.5, 
hvplot 0.10.0 
pandas 2.2.2 

#### Description of expected behavior and the observed behavior
Using Python hvplot Explorer is giving nuisance Line plot when using over 10,000 points. Is that a bug? can I configure the threshold value?

#### Complete, minimal, self-contained example code that reproduces the issue

```
import numpy as np
import pandas as pd
import hvplot.pandas

N=1001
Range = 10
DF = []
for n in range(Range):
    x = np.linspace(0.0, 6.4, num=N)
    y = np.sin(x) + n/10
    df = pd.DataFrame({'x': x, 'y': y})
    df['#'] = n 
    DF.append(df)

DF = pd.concat(DF)
print(f'param number= {N*Range}')
DF.hvplot.explorer(x='x', y='y', by = ['#'], kind='line')
```

```
```

#### Stack traceback and/or browser JavaScript console output
https://stackoverflow.com/questions/78950183/python-hvplot-explorer-limit-of-10-000-point

#### Screenshots or screencasts of the bug in action
![image](https://github.com/user-attachments/assets/4c2cae79-ef01-4b95-b27b-995067dfbd3c)

![image](https://github.com/user-attachments/assets/e52166c1-2727-4fbf-9acd-b2799e63f50c)

- [ ] I may be interested in making a pull request to address this

hagaishalevaei:
Digging in the hvplot code I found in \site-packages\hvplot\ui.py MAX_ROWS = 10000. Changing to MAX_ROWS = 100001 solved the issue.

ahuang11:
To clarify, nuisance in this sense means that it's not showing over 10k lines?

I suppose to improve this, we can add a editable slider widget to max out number of lines.

hagaishalevaei:
@ahuang11 - it is not 10k lines. The lines starting to act strange when the total amount of points is over 10,000. for example if you got 10 lines of 999 points there will be no problem, but if each line will be 1001 points you will get this strange thing.   


Good
```
N=999
Range = 10
DF = []
for n in range(Range):
    x = np.linspace(0.0, 6.4, num=N)
    y = np.sin(x) + n/10
[...]

PR #1408, opened 2024-09-09
Warn on imposing max rows and fix for curve

ahuang11:
Closes https://github.com/holoviz/hvplot/issues/1406

Adds a warning and checks if kind == curve to prevent random sampling; only sample head.

### DIFF
--- a/hvplot/tests/testui.py
+++ b/hvplot/tests/testui.py
@@ -11,7 +11,7 @@ import xarray as xr
 import pytest
 
 from bokeh.sampledata import penguins
-from hvplot.ui import hvDataFrameExplorer, hvGridExplorer
+from hvplot.ui import hvDataFrameExplorer, hvGridExplorer, MAX_ROWS
 
 df = penguins.data
 ds_air_temperature = xr.tutorial.open_dataset('air_temperature')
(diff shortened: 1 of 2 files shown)


hvplot/ui.py
```diff
@@ -688,10 +688,22 @@ def _plot(self):
         if len(df) > MAX_ROWS and not (
             self.kind in KINDS['stats'] or kwargs.get('rasterize') or kwargs.get('datashade')
         ):
-            df = df.sample(n=MAX_ROWS)
+            if self.kind == 'line':
+                param.main.param.warning(
```
maximlt, line 692: How about displaying these warnings in the alert box in addition/in place of as a programmatic warning?

maximlt (commented):
- I feel like `MAX_ROWS` could be increased, 10000 is kind of low specially considering that now HoloViews uses the webGL backend by default for Bokeh. How about 100000?
- Just for discussion for now, do you think we should expose this kind of setting in the explorer? Maybe in an advanced tab?

hvplot/ui.py
hagaishalevaei: I suggest to do df = df.sample(n=MAX_ROWS).sort_index()`. Otherwise the line plot will not work as it should.

ahuang11: I disagree with sort_index; the x may not always be index.

hagaishalevaei: good point. is there an option to add sort_index as an option? 
something like `df.hvplot.explorer(x='x', y='y', kind='line', sort_index=True)` 
where the defualt is `sort_index=False` ?

ahuang11: I think it probably makes more sense to do `sort_values("x")` when sampling for line.

hvplot/tests/testui.py
```diff
@@ -395,3 +395,22 @@ def test_explorer_geo_revise_kind(kind_tuple):
...
+    N = 10001
+    x = np.linspace(0.0, 6.4, num=N)
+    y = np.sin(x) + 10
+    df = pd.DataFrame({'x': x, 'y': y})
+    ui = hvplot.explorer(df, x='x', y='y', by=['#'], kind='line')
+    assert ui._data.equals(df.head(MAX_ROWS))
+
+
+def test_max_rows_sample():
+    N = 10001
```
ahuang11, line 410:
```suggestion
    N = 100001
    x = np.linspace(0.0, 6.4, num=N)
    y = np.sin(x) + 10
    df = pd.DataFrame({'x': x, 'y': y})
    ui = hvplot.explorer(df, x='x', y='y', by=['#'], kind='line')
    assert ui._data.equals(df.head(MAX_ROWS))


def test_max_rows_sample():
    N = 100001
```

maximlt (approved):

PR #1430, opened 2024-10-02
Support datashade points hover

### LINKED ISSUE #1428: Add support for hover when datashade points
Through ship traffics, I realized I can add hover tooltips to datashaded/rasterized plots; either through `hd.inspect_points` or a lookup_hook.
https://examples.holoviz.org/gallery/ship_traffic/ship_traffic.html#selecting-specific-datapoints

Can we have a reasonable default for hvPlot when `hover=True`? Or even HoloViews?

```python
import pandas as pd
import hvplot.pandas
import panel as pn
import datashader as ds
import holoviews as hv

pop = pd.DataFrame(
    {
        "lon": [-86.75, -86.5, -86.25, -86.0],
        "lat": [33.75, 34.0, 34.25, 34.5],
        "population": [100, 200, 300, 400],
        "R": ["A", "B", "A", "B"],
    }
)

# lon_lat_to_easting_northing
pop["x"], pop["y"] = hv.util.transform.lon_lat_to_easting_northing(
    pop["lon"], pop["lat"]
)
tiles = hv.element.tiles.EsriStreet()

alpha_slider = pn.widgets.FloatSlider(name="Population", start=0, end=1, step=0.1)

highlighter = hd.inspect_points.instance(streams=[hv.streams.Tap])

points = pop.hvplot.points(
    "x",
    "y",
    datashade=True,
    aggregator=ds.count_cat("R"),
    dynspread=True,
    height=800,
    width=1000,
)
highlight = highlighter(points).opts(
    color="white", tools=["hover"], marker="square", size=10, fill_alpha=0
)
points * highlight
```
image

Currently, if you just do hover=True alongside datashade, it gives a useless RGB value
```python
points = pop.hvplot.points(
    "x",
    "y",
    datashade=True,
    aggregator=ds.count_cat("R"),
    dynspread=True,
    height=800,
    width=1000,
    hover=True
)
points
```
image

ahuang11:
Adds support for hovering over datashaded points

Closes https://github.com/holoviz/hvplot/issues/1428

```python
import pandas as pd
import hvplot.pandas
import datashader as ds

df = pd.DataFrame(
    {
        "lon": [-86.75, -86.75, -86.25, -86.25],
        "lat": [33.75, 34.0, 34.49, 34.5],
        "population": [100, 200, 300, 400],
        "c": ["A", "B", "A", "B"],
    }
)

p = df.hvplot.points(
    "lon",
    "lat",
    hover_cols="all",
    datashade=True,
    dynspread=True,
    aggregator=ds.count_cat("c"),
    hover_tooltips=["lat", "lon", ("Population", "@population"), ("Alpha", "@A"), ("Beta", "@B")],  # not necessary
    padding=0.2
)
p
```

https://github.com/user-attachments/assets/a8a71946-5745-42c6-ba14-e40ca51b2865

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -45,7 +45,7 @@ from holoviews.element import (
 from holoviews.plotting.bokeh import OverlayPlot, colormap_generator
 from holoviews.plotting.util import process_cmap
 from holoviews.operation import histogram, apply_when
-from holoviews.streams import Buffer, Pipe
+from holoviews.streams import Buffer, Pipe, PointerXY
 from holoviews.util.transform import dim, lon_lat_to_easting_northing
 from pandas import DatetimeIndex, MultiIndex
 
@@ -842,13 +842,22 @@ class HoloViewsConverter:
         if kind == 'errorbars':
             hover = False
         elif hover is None:
-            hover = not self.datashade
+            hover = True
+
         if hover and not any(
             t for t in tools if isinstance(t, HoverTool) or t in ['hover', 'vline', 'hline']
         ):
             if hover in {'vline', 'hline'}:
                 plot_opts['hover_mode'] = hover
-            tools.append('hover')
+                self.hover_mode = hover
+            else:
+                self.hover_mode = 'mouse'
+            if not self.datashade:
+                tools.append('hover')
+
+        self.hover = bool(hover)
+        self.hover_tooltips = hover_tooltips
+        self.hover_formatters = hover_formatters
         if 'hover' in tools:
             if hover_tooltips:
                 plot_opts['hover_tooltips'] = hover_tooltips
@@ -1760,7 +1769,7 @@ class HoloViewsConverter:
             return layers
 
         import_datashader()
-        from holoviews.operation.datashader import datashade, rasterize, dynspread
+        from holoviews.operation.datashader import datashade, rasterize, dynspread, inspect_points
 
         categorical, agg = self._process_categorical_datashader()
         if agg:
@@ -1819,10 +1828,72 @@ class HoloViewsConverter:
                 threshold=self.kwds.get('threshold', 0.5),
             )
 
+        # a workaround to show hover info for datashaded points
+        if self.hover and self.datashade and self.kind == 'points':
+            inspector = inspect_points.instance(
+                streams=[PointerXY], transform=self._datashade_hover_transform
+            )
+            processed *= inspector(processed).opts(
+                size=10,
+                alpha=0,
+                tools=['hover'],
+                hover_mode=self.hover_mode,
+                hover_tooltips=self.hover_tooltips,
+                hover_formatters=self.hover_formatters,
+            )
+
         opts = filter_opts(eltype, dict(self._plot_opts, **style), backend='bokeh')
         layers = self._apply_layers(processed).opts(eltype, **opts, backend='bokeh')
         layers = _transfer_opts_cur_backend(layers)
         return layers
+
+    def _datashade_hover_transform(self, df):
+        if not len(df):
+            return df
+
+        # show at least the x and y columns
+        cols = self.hover_cols.copy()
+        if self.x not in cols:
+            cols.append(self.x)
+        if self.y not in cols:
+            cols.append(self.y)
+
+        # handle aggregator, e.g. ds.sum('column') or ds.count_cat('column')
+        agg_col = None
+        agg_series_map = {}
+        if self.aggregator and not isinstance(self.aggregator, str) and self.aggregator.column:
+            agg_col = self.aggregator.column
+            agg_op = type(self.aggregator).__name__
+            if hasattr(df, agg_op):  # df.sum/df.count
+                agg_value = df.agg({agg_col: agg_op})
+            elif agg_op == 'count_cat':
+                agg_value = df[agg_col].value_counts()
+
+            if agg_col in cols:
+                cols.remove(agg_col)
+
+        # take the mean of numeric columns
+        num_series = df[cols].select_dtypes(include=['number']).mean()
+        if len(num_series):
+            agg_series_map['number_cols'] = num_series
+
+        # take the first value of object columns
+        obj_series = df[cols].select_dtypes(exclude=['number']).iloc[0]
+        if len(obj_series):
+            agg_series_map['object_cols'] = obj_series
+
+        # to preserve order of other columns, add this last
+        if agg_col:
+            agg_series_map[agg_col] = agg_value
+
+        # concat all series into a single dataframe which has one row
+        df_hover = pd.concat(agg_series_map.values()).to_frame().transpose()
+
+        # remove index if it wasn't in the original dataset
+        if 'index' not in self.data.columns:
+            df_hover = df_hover.drop(columns=['index'], errors='ignore')
+
+        return df_hover
 
     def _resample_obj(self, operation, obj, opts):
         def exceeds_resample_when(plot):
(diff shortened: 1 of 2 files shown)


ahuang11:
Tap instead of PointerXY for big datasets

https://github.com/user-attachments/assets/64ff6c5d-d390-4de9-8f30-ca54d650e892

hvplot/converter.py
```diff
@@ -1819,11 +1828,94 @@ def method_wrapper(ds, x, y):
...
+        # a workaround to show hover info for datashaded points
+        if self.hover and self.datashade and self.kind == 'points':
+            if self.hover_mode != 'mouse':
+                param.main.param.warning(
+                    f'Got unsupported hover_mode={self.hover_mode!r} for '
+                    f"datashaded points; reverting to 'mouse'."
+                )
+
+            stream = PointerXY
+            if len(self.data) > 10000:
```
hoxbro, line 1840: I am not a fan of these magic numbers, with no way to change it.

hvplot/converter.py
```diff
@@ -1819,11 +1828,94 @@ def method_wrapper(ds, x, y):
...
+
+            if agg_col in cols:
+                cols.remove(agg_col)
+        else:
+            key = 'Count'
+            for i in range(1, 10):
+                if key in df.columns:
+                    key = f'Count_{i}'
+                else:
+                    break
```
hoxbro, line 1894: This looks weird to me

hvplot/converter.py
```diff
@@ -1819,11 +1828,94 @@ def method_wrapper(ds, x, y):
...
+            agg_value = pd.Series([len(df)], index=[key])
+
+        # take the mean of numeric columns
```
hoxbro, line 1897: Is the rest of the method generated with AI? If so, please clearly state so.  

Otherwise, please explain _why_ you take the mean. In general, comments should not explain _what_ is done but _why_ it is done.

hvplot/converter.py
```diff
@@ -842,13 +842,22 @@ def __init__(
         if kind == 'errorbars':
             hover = False
         elif hover is None:
-            hover = not self.datashade
+            hover = True
```
hoxbro, line 845:
```suggestion
            hover = (self.datashade and self.kind == 'points') or not self.datashade
```
I'm not sure if hover should always be True.

hoxbro (commented):
Some comments:

1) Does this work with other data backends like `dask.dataframe`? 
2) This is a (small) breaking change as the output of the `DynamicMap` generated by hvplot (with `datashade=True, hover=True`) is no longer an `RGB` but an `Overlay`. This could lead to problems in the future.
3) We could add `inspect` as an option to `hover` to enable this feature for now. Then, we can always make it the default in the next minor release.

hvplot/converter.py
ahuang11: No it's not AI. I decided to add comments since the code was getting chunky, and easier to navigate.

> Comments should not explain what is done but why it is done.

Sometimes it's easier to read English than code and guidelines don't necessarily have to be followed all the time

hvplot/converter.py
```diff
@@ -1819,11 +1828,94 @@ def method_wrapper(ds, x, y):
...
+        # take the first value of object columns
+        obj_series = df[cols].select_dtypes(exclude=['number']).iloc[0]
+        if len(obj_series):
+            agg_series_map['object_cols'] = obj_series
+
+        # to preserve order of other columns, add this last
+        agg_series_map[agg_col] = agg_value
+
+        # concat all series into a single dataframe which has one row
+        df_hover = pd.concat(agg_series_map.values()).to_frame().transpose()
```
ahuang11, line 1911: This especially, without the comment, would be really confusing of what all the methods are doing.

hvplot/converter.py
ahuang11: What's weird about it? It's deduplicating columns; do you have a suggestion for an alternative?

hvplot/converter.py
hoxbro: Agree. But why is this needed? Why is a `pd.Series` not good enough?  We could add an extra line explaining this:
```suggestion
        # This is needed as the function used in `inspect_points.transform` must return a DataFrame.
        df_hover = pd.concat(agg_series_map.values()).to_frame().transpose()
```

At some point, a person could look at this line and remove `.to_frame().transpose()`, and if no tests fails people could think it was not a needed conversion.

hvplot/converter.py
hoxbro: Sorry about that. It just had the structure of an AI output. 

> ... guidelines don't necessarily have to be followed all the time.

I agree, but the use of `mean` need to be explained, either in the code or in the original comment. 

This can return an output that does not exist in the original DataFrame when more points are in the input DataFrame. For example, the two close points in your example will return 350 for the population when they are close to each other. Is this okay? Maybe, but you need to state that you have made this decision.

hvplot/converter.py
hoxbro: Why 10? 

I can't wrap my head around what the code does.

philippjfr:
>This is a (small) breaking change as the output of the DynamicMap generated by hvplot (with datashade=True, hover=True) is no longer an RGB but an Overlay. This could lead to problems in the future.

Yes, I haven't reviewed in detail but if this is the approach taken we should wait for https://github.com/orgs/holoviz/projects/9 to be addressed properly rather than "hacking" it with a fake layer at the hvplot layer.

hvplot/converter.py
jbednar: Just about any scalar value other than 0 or 1 should be documented very clearly, usually by making a Parameter or at least attribute declaration with a meaningful name and associated docstring.

hvplot/converter.py
jbednar: @ahuang11 , @hoxbro thinks you are so intelligent that it must be artificial! :-)

hvplot/tests/testoperations.py
```diff
@@ -324,6 +319,23 @@ def test_downsample_resample_when(self, kind, eltype):
         assert isinstance(element, eltype)
         assert len(element) == 0
 
+    @parameterized.expand([(None,), (True,), ('vline',), ('hline',)])
```
maximlt, line 322: Many more tests will be needed to cover all the branches of the code introduced in the converter, these first two are a good start but they're very basic and just check whether the code doesn't error and the HoloViews type is the expected one.

ahuang11:
> this is the approach taken we should wait for https://github.com/orgs/holoviz/projects/9 to be addressed properly rather than "hacking" it with a fake layer at the hvplot layer.

To clarify, which tasks in https://github.com/orgs/holoviz/projects/9 should be completed first? Should we ditch this PR?

maximlt:
Marked this PR as Draft per the comments before. @hoxbro let us know if you can share the progress of https://github.com/orgs/holoviz/projects/9 .

maximlt:
> bumping this, now that Datashader inspections for HoloViews is merged

It has been merged in HoloViews but has not yet made it into a final release. As far as I can see it hasn't been documented in HoloViews yet (I wouldn't release hvPlot with an undocumented HoloViews feature).

I am unsure if we should pursue this PR as the implementation seems to differ from how datashader inspections ended up being implemented in HoloViews (see https://github.com/holoviz/holoviews/pull/6422).

jbednar:
Whether via this PR or another, I strongly support surfacing the improved inspection support from HoloViews so that it is conveniently expose in hvPlot.

maximlt:
Yes that's the plan, once it's documented and released in Holoviews, and then implemented and documented in hvPlot.

Issue #1431, opened 2024-10-03
Tiles with Dask DataFrame no longer works in 0.11.0


hoxbro:
#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc should be added within the dropdown below.)

<details>
    <summary>Software Version Info</summary>

```plaintext
hvplot 0.11.0
```
</details>

#### Description of expected behavior and the observed behavior
Tiles with Dask.DataFrame no longer works in 0.11.0. This is an regression introduced in https://github.com/holoviz/hvplot/pull/1377

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import dask.dataframe as dd
import hvplot.dask  # noqa
import pandas as pd

data = {"x": [-9700947, -9584977], "y": [3890186, 3824923]}
df = dd.from_pandas(pd.DataFrame(data), npartitions=2)
df.hvplot.points("x", "y", tiles=True)

```

#### Stack traceback and/or browser JavaScript console output

``` python-traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[1], line 7
      5 data = {"x": [-9700947, -9584977], "y": [3890186, 3824923]}
      6 df = dd.from_pandas(pd.DataFrame(data), npartitions=2)
----> 7 df.hvplot.points("x", "y", tiles=True)

File [~/projects/holoviz/repos/hvplot/hvplot/plotting/core.py:1672](http://localhost:8888/lab/workspaces/auto-W/tree/repos/hvplot/hvplot/plotting/core.py#line=1671), in hvPlotTabular.points(self, x, y, **kwds)
   1614 def points(self, x=None, y=None, **kwds):
   1615     """
   1616     A `points` plot visualizes positions in a 2D space. This is useful for example for
   1617     geographic plots.
   (...)
   1670     - HoloViews: https://holoviews.org/reference/elements/bokeh/Points.html
   1671     """
-> 1672     return self(x, y, kind='points', **kwds)

File ~/projects/holoviz/repos/hvplot/hvplot/plotting/core.py:95, in hvPlotBase.__call__(self, x, y, kind, **kwds)
     92         plot = self._get_converter(x, y, kind, **kwds)(kind, x, y)
     93         return pn.panel(plot, **panel_dict)
---> 95 return self._get_converter(x, y, kind, **kwds)(kind, x, y)

[...]

PR #1432, opened 2024-10-04
Fix dask tiles regression

ahuang11:
Closes https://github.com/holoviz/hvplot/issues/1431
```python
import dask.dataframe as dd
import hvplot.dask  # noqa
import pandas as pd

data = {"x": [-9700947, -9584977], "y": [3890186, 3824923]}
df = dd.from_pandas(pd.DataFrame(data), npartitions=2)
df.hvplot.points("x", "y", tiles=True)
```

image

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -2173,16 +2173,13 @@ class HoloViewsConverter:
             if data.crs is not None:
                 data = data.to_crs(epsg=3857)
             return data, x, y
-        else:
+        elif not (is_dask(data) or is_ibis(data)):
+            # no dask/ibis as to prevent eager evaluation
+            # https://github.com/holoviz/hvplot/pull/1432/files#r1789862990
             min_x = np.min(data[x])
             max_x = np.max(data[x])
             min_y = np.min(data[y])
             max_y = np.max(data[y])
-
-            if hasattr(min_x, 'compute'):
-                min_x, max_x = min_x.compute(), max_x.compute()
-            if hasattr(min_y, 'compute'):
-                min_y, max_y = min_y.compute(), max_y.compute()
 
             x_within_bounds = -180 <= min_x <= 360 and -180 <= max_x <= 360
             y_within_bounds = -90 <= min_y <= 90 and -90 <= max_y <= 90
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -2175,8 +2175,7 @@ class HoloViewsConverter:
                 data = data.to_crs(epsg=3857)
             return data, x, y
         elif not is_lazy_data(data):
-            # no dask/ibis as to prevent eager evaluation
-            # https://github.com/holoviz/hvplot/pull/1432/files#r1789862990
+            # To prevent eager evaluation: https://github.com/holoviz/hvplot/pull/1432
             min_x = np.min(data[x])
             max_x = np.max(data[x])
             min_y = np.min(data[y])
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -60,6 +60,7 @@ from .util import (
     is_cudf,
     is_streamz,
     is_ibis,
+    is_lazy_data,
     is_xarray,
     is_xarray_dataarray,
     process_crs,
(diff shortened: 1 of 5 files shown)


hoxbro (changes requested):
Failing test

hvplot/converter.py
```diff
@@ -2178,6 +2178,12 @@ def _process_tiles_without_geo(self, data, x, y):
             max_x = np.max(data[x])
             min_y = np.min(data[y])
             max_y = np.max(data[y])
+
+            if hasattr(min_x, 'compute'):
+                min_x, max_x = min_x.compute(), max_x.compute()
+            if hasattr(min_y, 'compute'):
+                min_y, max_y = min_y.compute(), max_y.compute()
```
philippjfr, line 2185: This is not a good idea for a number of reasons:

1. Adding additional compute calls can potentially be VERY expensive. Recall when you implemented `symmetric=True` and I suggested that we should not do automatic computations like this because they can potentially run extra computations on terrabytes of data.
2. Even if we decided this was a good idea, you are running four completely separate computes here!!! Dask has ways to fuse multiple calls like this into a single pass over the data.

maximlt: @philippjfr what do you suggest we do instead?

1. Force the user to set the projection in this case? (Ibis data too I guess)
2. Would it be sufficient to check that just a few values (or even 1) are within the bounds?

philippjfr: 1. Yes, I think we need to document that certain automatic behaviors are not enabled by default on lazy datasets.
2. Maybe but not super convinced, you'd have to write logic to fetch only one chunk I guess.

hvplot/converter.py
```diff
@@ -2173,11 +2173,14 @@ def _process_tiles_without_geo(self, data, x, y):
             if data.crs is not None:
                 data = data.to_crs(epsg=3857)
             return data, x, y
-        else:
+        elif not (is_dask(data) or is_ibis(data)):
```
hoxbro, line 2176:
```suggestion
        elif not is_lazy_data(data):
```

So we can easy expand it in the future

hvplot/converter.py
```diff
@@ -2173,11 +2174,14 @@ def _process_tiles_without_geo(self, data, x, y):
             if data.crs is not None:
                 data = data.to_crs(epsg=3857)
             return data, x, y
-        else:
+        elif not is_lazy_data(data):
+            # no dask/ibis as to prevent eager evaluation
+            # https://github.com/holoviz/hvplot/pull/1432/files#r1789862990
```
hoxbro, line 2179:
```suggestion
            # To prevent eager evaluation: https://github.com/holoviz/hvplot/pull/1432
```

hvplot/util.py
```diff
@@ -446,6 +446,16 @@ def is_xarray(data):
     return isinstance(data, (DataArray, Dataset))
 
 
+def is_lazy_data(data):
```
hoxbro, line 449:
```suggestion
def is_lazy_data(data):
    """ Check if data is lazy

    This checks if the datatype is Dask, Ibis, or Polars' LazyFrame.
    It is useful to avoid eager evaluation of the data.
    """
```

hvplot/util.py
```diff
@@ -446,6 +446,16 @@ def is_xarray(data):
...
+    if is_dask(data) or is_ibis(data):
+        return True
+    elif is_polars(data):
+        import polars as pl
+
+        return isinstance(data, pl.LazyFrame)
```
maximlt, line 455: Doesn't `is_polars` take care of this already?

maximlt (commented):
What do we recommend to users that have a dask dataset with lat/lon coordinates?

hvplot/util.py
philippjfr: No, because there's eager and lazy polars DataFrames.

philippjfr:
>What do we recommend to users that have a dask dataset with lat/lon coordinates?

Good question, you should still be able to make it project by explicitly requesting it.

ahuang11:
> What do we recommend to users that have a dask dataset with lat/lon coordinates?

Same behavior from <0.11 releases; manually setting `crs` and `projection`.

maximlt:
> > What do we recommend to users that have a dask dataset with lat/lon coordinates?
> 
> Same behavior from <0.11 releases; manually setting `crs` and `projection`; not sure if this needs to be explicitly called out though since I do mention "automatically unless..."
> 
> > If coordinate values fall within lat/lon bounds, auto-projects to EPSG:3857 unless `projection=False` or if the data is lazily loaded (dask / ibis).

But that still requires GeoViews? If so, maybe there's a way to special case `crs="4326"` (or some of the other allowed values) so it doesn't use GeoViews. But I'd say that's for another PR, this one was about fixing the regression and it achieved that, thanks.

maximlt (approved):

PR #1446, opened 2024-10-21
Fix using HoloViews private API to get datashader's version in scatter_matrix

maximlt:


hvplot/plotting/scatter_matrix.py
```diff
@@ -113,10 +113,11 @@ def scatter_matrix(
         raise ValueError('dynspread or spread need rasterize or datashade to be set to True.')
 
     if rasterize:
+        import datashader as ds
         import holoviews.operation.datashader as hd
 
         if dynspread or spread:
-            if hd.ds_version < Version('0.12.0'):
+            if Version(ds.__version__) < Version('0.12.0'):
```
hoxbro, line 120: FYI: Datashader 0.12.0 was the first version released on Python 3.9.

PR #1451, opened 2024-10-29
Fix logic for no geoviews explorer

ahuang11:
Even though `geo=False`, explorer would raise `raise ImportError('GeoViews must be installed to enable the geographic options.')`

```python
import pandas as pd
import hvplot.pandas
df = pd.DataFrame({'a': [1,2,3], 'b': [4,5,6]})

df.hvplot.explorer(geo=False)
```

Not sure how to add a test for this because the dev dependencies installs geoviews.

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -76,7 +76,7 @@ from .util import (
     process_derived_datetime_pandas,
     _convert_col_names_to_str,
     import_datashader,
-    geoviews_is_available,
+    import_geoviews,
 )
 from .utilities import hvplot_extension
 
@@ -649,7 +649,7 @@ class HoloViewsConverter:
         self.geo = any([geo, crs, global_extent, projection, project, coastline, features])
         # Try importing geoviews if geo-features requested
         if self.geo or self.datatype == 'geopandas':
-            geoviews_is_available(raise_error=True)
+            import_geoviews()
 
         self.crs = self._process_crs(data, crs) if self.geo else None
         self.output_projection = self.crs
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -76,6 +76,7 @@ from .util import (
     process_derived_datetime_pandas,
     _convert_col_names_to_str,
     import_datashader,
+    geoviews_is_available,
 )
 from .utilities import hvplot_extension
 
@@ -646,6 +647,10 @@ class HoloViewsConverter:
 
         self.dynamic = dynamic
         self.geo = any([geo, crs, global_extent, projection, project, coastline, features])
+        # Try importing geoviews if geo-features requested
+        if self.geo or self.datatype == 'geopandas':
+            geoviews_is_available(raise_error=True)
+
         self.crs = self._process_crs(data, crs) if self.geo else None
         self.output_projection = self.crs
         self.project = project
@@ -655,17 +660,6 @@ class HoloViewsConverter:
         self.tiles_opts = tiles_opts or {}
         self.sort_date = sort_date
 
-        # Import geoviews if geo-features requested
-        if self.geo or self.datatype == 'geopandas':
-            try:
-                import geoviews  # noqa
-            except ImportError:
-                raise ImportError(
-                    'In order to use geo-related features '
-                    'the geoviews library must be available. '
-                    'It can be installed with:\n  conda '
-                    'install geoviews'
-                )
         if self.geo:
             if self.kind not in self._geo_types:
                 param.main.param.warning(
--- a/hvplot/util.py
+++ b/hvplot/util.py
@@ -4,6 +4,7 @@ Provides utilities to convert data and projections
 
 import sys
 import types
+import typing
 
 from collections.abc import Hashable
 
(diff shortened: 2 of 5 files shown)


ahuang11:
ooh that's fancy and good to know! Turns out there actually was a marker for `geo` tests so I didn't need it.

hoxbro:
Are you sure?

maximlt:
> ooh that's fancy and good to know! Turns

It's also a little crazy right 🙃 

A suggestion could be to refactor `gv_available` into a utility function that is mocked for this test to return `False`.

ahuang11:
Hm I guess there is a marker, but the tests still run everything
https://github.com/holoviz/hvplot/blob/main/.github/workflows/test.yaml#L171
https://github.com/holoviz/hvplot/blob/main/hvplot/tests/testui.py#L394

I guess I'll go with Maxime's suggestion

hvplot/converter.py
```diff
@@ -654,17 +659,6 @@ def __init__(
...
-        # Import geoviews if geo-features requested
-        if self.geo or self.datatype == 'geopandas':
-            try:
-                import geoviews  # noqa
-            except ImportError:
-                raise ImportError(
-                    'In order to use geo-related features '
-                    'the geoviews library must be available. '
-                    'It can be installed with:\n  conda '
-                    'install geoviews'
```
ahuang11, line 666: Refactored it into geoviews is available, and moved it up because `self.crs = ...` was crashing without geoviews.

hvplot/util.py
```diff
@@ -750,3 +750,21 @@ def relabel_redim(hv_obj, relabel_kwargs, redim_kwargs):
...
+
+        gv_available = True
+    except ImportError:
+        gv_available = False
+
+    if not raise_error:
+        return gv_available
+
+    if not gv_available and raise_error:
+        raise ImportError('GeoViews must be installed to enable the geographic options.')
```
hoxbro, line 770:
```suggestion
    if not gv_available and raise_error:
        raise ImportError('GeoViews must be installed to enable the geographic options.')
    return gv_available
```

You could also use `importlib.find_spec`

hvplot/util.py
```diff
@@ -750,3 +751,14 @@ def relabel_redim(hv_obj, relabel_kwargs, redim_kwargs):
...
+
+
+def geoviews_is_available(raise_error: bool = False):
+    """
+    Check if GeoViews is available and raise an ImportError if not.
+    """
+    gv_available = find_spec('geoviews') is not None
+
+    if not gv_available and raise_error:
+        raise ImportError('GeoViews must be installed to enable the geographic options.')
```
maximlt, line 763: I got confused a minute here, does `find_spec('geoviews')` actually import GeoViews?

hoxbro: No, but it checks if it is installed. 

image

maximlt: Yes ok that's what I thought. I was confused as then an `ImportError` is raised. Since importing GeoViews comes with side effects (registering lots of stuff), I'd like to make sure we're not removing an important step there. @hoxbro did you suggest `find_spec` for performance reasons?

hoxbro: Just as an alternative and also easier to patch in the test suite.

I think it is okay to use `find_spec` when we check if it is available. We should import geoviews in the code, where there previously was a `try/except`, to get those juicy side effects.

hoxbro: I also think the error message is wrong here. As it no longer only affects the UI code.

hvplot/util.py
```diff
@@ -730,7 +731,7 @@ def import_datashader():
     return datashader
 
 
-def import_geoviews() -> types.ModuleType | None:
+def import_geoviews() -> typing.Union[types.ModuleType, None]:
```
ahuang11, line 734: An alternative is `from __future__ import annotations`

maximlt: The issue was, I think, with `|` not being supported on Python 3.9. I don't think that would be solved by `from __future__ import annotations`?

```
    def import_geoviews() -> types.ModuleType | None:
E   TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

maximlt:
@ahuang11 what do you think about the refactor done in https://github.com/holoviz/hvplot/pull/1451/commits/31df383665bcbd5d14c1e49bc5a030f9e62b9357? I'm fine with reverting if you prefer the former way. I've preferred creating a function that does the actual import since the side effects of importing geoviews as required I think.

hvplot/util.py
ahuang11: I think it does backport; I remember doing it before if my memory is correct :P
https://www.reddit.com/r/learnpython/comments/14jsyj0/from_future_import_annotation_behavior/

ahuang11:
I'm fine with the refactor.

hvplot/util.py
ahuang11: > The annotations feature doesn't do anything other than implicitly convert all annotations into strings, meaning there's nothing for the interpreter to evaluate, so even invalid operator use would stop raising errors. The binary OR operator is no exception.

Oh maybe it's just stringifying it?

maximlt: Removed the type hints, let's add them in another PR with mypy testing.

maximlt (approved):

PR #1452, opened 2024-10-31
ci: Fix failing CI

hoxbro:
See failing test here: https://github.com/holoviz/hvplot/actions/runs/11600853910/job/32302285188

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -150,7 +150,7 @@ jobs:
       matrix: ${{ fromJson(needs.setup.outputs.matrix) }}
     defaults:
       run:
-        shell: bash -l {0}
+        shell: bash -e {0}
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -150,7 +150,7 @@ jobs:
       matrix: ${{ fromJson(needs.setup.outputs.matrix) }}
     defaults:
       run:
-        shell: bash -l {0}
+        shell: bash -e {0}
```
hoxbro, line 153: There is no need to load the settings file (`-l`), like `.bashrc` which could pollute the PATH. 

Though it is important to exit if one command fails (`-e`).

.github/workflows/test.yaml
```diff
@@ -159,9 +159,11 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: install with geo
-        run: pip install -v --prefer-binary -e '.[tests, examples-tests, geo, hvdev, hvdev-geo, dev-extras]'
-      - name: pip list
-        run: pip list
+        run: python -m pip install -v --prefer-binary -e '.[tests, examples-tests, geo, hvdev, hvdev-geo, dev-extras]'
```
hoxbro, line 162: Good practice to call `python -m pip`, see https://pip.pypa.io/en/stable/cli/pip_install/.

.github/workflows/test.yaml
```diff
@@ -159,9 +159,11 @@ jobs:
...
+      - name: python version and pip list
+        run: |
+          python --version --version
```
hoxbro, line 165: Just nice to know :shrug:

maximlt (approved):

PR #1466, opened 2024-12-16
compat: Add upper pin for intake-xarray

hoxbro:
intake-xarray 2.0 has been released, see https://pypi.org/project/intake-xarray/#history

Failing the current examples on conda. 

image


maximlt (approved):
(annoying timing)
Thanks!

PR #1470, opened 2025-01-20
ci: migrate to pixi

maximlt:
This PR migrates the developer workflow to Pixi. This is pretty much aligned with what has already been done for most HoloViz packages, with some differences:

- Adopted recommending using the `default` environment with pretty much all that is required to develop hvPlot (tests, jupyterlab, etc.), excluding building the docs (could be included). This approach was recently adopted by HoloViews. EDIT: With a bit more streamlined experience provided by `pixi run install-dev` (create env, editable install, set up pre-commit, download test datasets).
- It didn't move yet to `hatchling` to build hvPlot as I don't want to mix things too much in this PR that is already big enough.
- It doesn't include the "Waiting room" in the GH build workflow, my experience releasing Param was that I always forget clicking on the button to publish and I didn't see much benefit using it.
- It keeps the extras in `pyproject.toml` and some jobs running `pip`, I'd like to keep the option to develop hvPlot with pip (even if undocumented), I'm concerned Pixi makes the life of re-distributors too difficult (making it less likely for hvPlot to be re-distributed).

---

- [x] Add PR description
- [x] Update pixi.toml to use new approach adopted by HoloViews (1 working dev env)
- [x] Ensure the tests / docs / build workflow work as expected
- [x] Update the developer guide


maximlt:
Merging while it works, can be refined later.

.github/workflows/nightly_lock.yaml
```diff
@@ -0,0 +1,25 @@
+name: nightly_lock
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 0 * * *"
+
+env:
+  PACKAGE: "holoviews"
```
hoxbro, line 8: hvplot

PR #1477, opened 2025-01-24
Removed 'Home' from website index and replace hide-cell with remove-cell

Azaya89:
- [x] Removed `Home` from website index toctree. This reduces redundancy in the website seeing as clicking on the hvPlot image already redirects to the main homepage.
- [x] Changes all instance of `hide-cell` in the notebooks to `remove-cell`

### DIFF
--- a/doc/index.md
+++ b/doc/index.md
@@ -420,7 +420,6 @@ align: center
 :hidden:
 :maxdepth: 2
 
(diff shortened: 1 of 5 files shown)


maximlt (approved):
Thanks! (Best not to mix too much separate changes in the same PR)

PR #1482, opened 2025-01-30
Pandas API update

Azaya89:
- [x] update notebook to match current pandas API
- [x] remove test cell from output in `user_guide/Viewing`


maximlt:
If it doesn't lead to a notebook that is too heavy, we could leverage Panel to display the outputs generated for the same calls (possibly with some slight differences) with hvplot and pandas plotting backends in tabs.

```python
import matplotlib.pyplot as plt
import numpy as np
import pandas as pd
from bokeh.sampledata.penguins import data as df
import hvplot.pandas
import panel as pn

pn.extension()

def compare(pd_hvplot, pd_plot):
    # Pandas .plot
    fig = pd_plot.figure
    pn_mpl = pn.pane.Matplotlib(fig, dpi=80)
    plt.close(fig)

    # Pandas .hvplot
    pn_hv = pn.pane.HoloViews(pd_hvplot)
    return pn.Tabs(
        ('.hvplot()', pn_hv),
        ('pandas .plot()', pn_mpl),
    )

np.random.seed(123456)
ts = pd.Series(np.random.randn(1000), index=pd.date_range("1/1/2000", periods=1000))
ts = ts.cumsum()

compare(
    ts.plot(backend='hvplot'),
    ts.plot(),
)
```

image
image

Azaya89:
Thanks. I will test this and see how it goes.

maximlt:
> Thanks. I will test this and see how it goes.

Cool, though let's wait until this is the appropriate time to work on this.

maximlt:
I've pushed some significant changes in https://github.com/holoviz/hvplot/pull/1482/commits/c9355b5ea26ef81ce5063bd14ba5c42674b4185e. I went through the Pandas_API notebook, adding for each plot the original Pandas version. This allowed me to find many differences between hvPlot and API, which are highlighted in note admonitions in the notebook. At the same time, I added a new API compatibility section to the reference, with a page for Pandas that includes an API compatibility table (which plotting methods are available in Pandas and/or hvPlot) and the main notable differences I thought were worth being mentioned specifically.

maximlt (approved):

Issue #1483, opened 2025-01-31
Using the hvplot backend for a hist() plot shows an AttributeError


Azaya89:
#### ALL software version info

<details>
    <summary>Software Version Info</summary>

```plaintext
HoloViews = 1.20.1a0
pandas = 2.2.3
numpy = 2.0.2
hvPlot = 0.11.3.dev6+g3c69ddb.d20250131
bokeh = 3.6.2
```
</details>

#### Description of expected behavior and the observed behavior

**Expected Behavior**
When I set `pd.options.plotting.backend = 'hvplot'` or pass `backend='hvplot'` to the `.hist()` method, I expect the histogram to render correctly using hvPlot (just as it does when using the `'holoviews'` backend).

**Observed Behavior**
Instead, calling `ts.hist(backend='hvplot')` results in a Traceback, indicating that hvPlot does not have a `hist_series` attribute.

#### Complete, minimal, self-contained example code that reproduces the issue

**Some data**
```python
import numpy as np
import pandas as pd

np.random.seed(123456)

ts = pd.Series(np.random.randn(1000), index=pd.date_range('1/1/2000', periods=1000))
```

**HoloViews backend**
```python
ts.hist(backend='holoviews')
```

**Output**

![Image](https://github.com/user-attachments/assets/283075a6-ebe0-43bf-86e4-cb35c6f17f1e)

**hvPlot backend**
```python
ts.hist(backend='hvplot')
```
*Traceback error*

#### Stack traceback and/or browser JavaScript console output
```
AttributeError                            Traceback (most recent call last)
Cell In[3], line 1
----> 1 ts.hist(backend='hvplot')

File ~/hvplot/.pixi/envs/default/lib/python3.12/site-packages/pandas/plotting/_core.py#line=128, in hist_series(self, by, ax, grid, xlabelsize, xrot, ylabelsize, yrot, figsize, bins, backend, legend, **kwargs)
     62 """
     63 Draw histogram of the input series using matplotlib.
     64 
   (...)
    126     >>> hist = ser.groupby(level=0).hist()
    127 """
    128 plot_backend = _get_plot_backend(backend)
--> 129 return plot_backend.hist_series(
    130     self,
    131     by=by,
    132     ax=ax,
    133     grid=grid,
    134     xlabelsize=xlabelsize,
    135     xrot=xrot,
    136     ylabelsize=ylabelsize,
    137     yrot=yrot,
    138     figsize=figsize,
    139     bins=bins,
[...]

PR #1491, opened 2025-02-05
Fix calling special Pandas plotting methods with the hvplot backend

maximlt:
Fixes https://github.com/holoviz/hvplot/issues/1483

When a plotting backend is configured on pandas via e.g. `pd.options.plotting.backend =  '<backend>'`, Pandas has two ways to use it:

1. it will try to see if it is registered as an entry point and if so will load the registered module (`holoviews = "hvplot:plotting"` registers `holoviews` as an entry point pointing to the `hvplot.plotting` module).
2. if 1/ doesn't succeed, it will load the module as is

In both cases, the only check Pandas performs is to verify that the backend has a `plot` attribute.

https://github.com/holoviz/hvplot/pull/347 exposed the `plot` function from `hvplot.plotting` in the top-level module, allowing users to register the backend with `pd.options.plotting.backend = 'hvplot'` (leveraging 2/). However, that is not enough to give users access to the special plot methods Pandas expects like `boxplot` or `hist`.

This PR fixes that by adding another `hvplot` entry point.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ dependencies = [
 
 [project.entry-points."pandas_plotting_backends"]
 holoviews = "hvplot:plotting"
+hvplot = "hvplot:plotting"
 
 [project.urls]
 Homepage = "https://hvplot.holoviz.org"
--- a/hvplot/tests/testplotting.py
+++ b/hvplot/tests/testplotting.py
@@ -17,7 +17,6 @@ frame_specials = [
     ('hist', hv.Histogram),
 ]
 series_specials = [('hist', hv.Histogram)]
-
 no_args_mapping = [
     (kind, el) for kind, el in HoloViewsConverter._kind_mapping.items() if kind in no_args
 ]
@@ -26,8 +25,12 @@ x_y_mapping = [(kind, el) for kind, el in HoloViewsConverter._kind_mapping.items
 
 @pytest.fixture(params=['holoviews', 'hvplot'])
 def plotting_backend(request):
+    initial_backend = pd.options.plotting.backend
     pd.options.plotting.backend = request.param
-    return request.param
+    try:
+        yield request.param
+    finally:
+        pd.options.plotting.backend = initial_backend
 
 
 @pytest.mark.parametrize('kind,el', no_args_mapping)
--- a/hvplot/tests/testplotting.py
+++ b/hvplot/tests/testplotting.py
@@ -13,10 +13,15 @@ from hvplot.tests.util import makeDataFrame
 no_args = ['line', 'area', 'hist', 'box', 'kde', 'density', 'bar', 'barh']
 x_y = ['scatter', 'hexbin']
 frame_specials = [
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -45,6 +45,7 @@ dependencies = [
 
 [project.entry-points."pandas_plotting_backends"]
 holoviews = "hvplot:plotting"
+hvplot = "hvplot:plotting"
```
maximlt, line 48: This is the fix.

Azaya89:
There are multiple ways a `hist` plot can be plotted and some of them still don't work while the ones that work give different outputs. I checked out this branch and tried them:

**Some data**
```python
import numpy as np
import pandas as pd

np.random.seed(123456)

ts = pd.Series(np.random.randn(1000), index=pd.date_range('1/1/2000', periods=1000))
```

1. **Using `hvplot.pandas` works**
```python
import hvplot.pandas

ts.hvplot.hist()
```
Default number of bins = 20
image


2. **Setting the plotting backend**
2a. 
```python

ts.hist()
```
```
AttributeError                            Traceback (most recent call last)
Cell In[3], line 3
      1 pd.options.plotting.backend = 'hvplot'
----> 3 ts.hist()

File ~/hvplot/.pixi/envs/default/lib/python3.12/site-packages/pandas/plotting/_core.py#line=128, in hist_series(self, by, ax, grid, xlabelsize, xrot, ylabelsize, yrot, figsize, bins, backend, legend, **kwargs)
     62 """
     63 Draw histogram of the input series using matplotlib.
     64 
   (...)
    126     >>> hist = ser.groupby(level=0).hist()
    127 """
    128 plot_backend = _get_plot_backend(backend)
--> 129 return plot_backend.hist_series(
    130     self,
    131     by=by,
    132     ax=ax,
    133     grid=grid,
    134     xlabelsize=xlabelsize,
    135     xrot=xrot,
    136     ylabelsize=ylabelsize,
    137     yrot=yrot,
    138     figsize=figsize,
    139     bins=bins,
    140     legend=legend,
    141     **kwargs,
    142 )

AttributeError: module 'hvplot' has no attribute 'hist_series'
```

2b. **Prepending with `plot` works**
```python

ts.plot.hist()
```
Default number of bins = 10
image

maximlt:
Did you re-install `hvplot` in your environment? I think it needs to be done for the new entry-point to be registered.

Azaya89:
> Did you re-install `hvplot` in your environment? I think it needs to be done for the new entry-point to be registered.

Thank you. I didn't know that.

Azaya89 (approved):

maximlt:
> > Did you re-install `hvplot` in your environment? I think it needs to be done for the new entry-point to be registered.
> 
> Thank you. I didn't know that.

Relieved to see it works as it's what is tested with https://github.com/holoviz/hvplot/pull/1491/files#diff-8d066bf9f51083145a38e3d3346ad7f86ac33f994b9008217f2cc92a91aa0b05R72-R77

PR #1493, opened 2025-02-06
Set OHLC x-axis tooltip to datetime format

### LINKED ISSUE #1487: OHLC hover text date format bug
<details>
    <summary>Software Version Info</summary>

```plaintext
python 3.12.8
bokeh 3.6.2
holoviews 1.20.0
```
</details>

The tooltip/hover text in ohlc plots does not recognise datetime formats, shows a long integer instead. See screenshot below of hvplot reference gallery example.

https://hvplot.holoviz.org/reference/tabular/ohlc.html

#### Screenshots or screencasts of the bug in action

Image

Azaya89:
- [x] fixes #1487 
- [x] Added tests

```python
import pandas as pd
import hvplot.pandas  # noqa
from bokeh.sampledata import stocks

df = pd.DataFrame(stocks.AAPL)
df['date'] = pd.to_datetime(df.date)

df.iloc[-50:].hvplot.ohlc(grid=True)
```

image

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -2632,7 +2632,9 @@ class HoloViewsConverter:
         if 'hover' in tools:
             x_data = data[x] if x in data.columns else data.index
             if pd.api.types.is_datetime64_any_dtype(x_data):
-                x_tooltip = f'@{x}{{%F}}'
+                # %F %T: strftime code for %Y-%m-%d %H:%M:%S.
+                # See https://man7.org/linux/man-pages/man3/strftime.3.html
+                x_tooltip = f'@{x}{{%F %T}}'
                 formatter = {f'@{x}': 'datetime'}
             else:
                 x_tooltip = f'@{x}'
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -2630,10 +2630,17 @@ class HoloViewsConverter:
         seg_cur_opts, seg_compat_opts = self._get_compat_opts('Segments')
         tools = seg_cur_opts.pop('tools', [])
         if 'hover' in tools:
+            x_data = data[x] if x in data.columns else data.index
+            if np.issubdtype(x_data.dtype, np.datetime64):
+                x_tooltip = f'@{x}{{%F}}'
+                formatter = {f'@{x}': 'datetime'}
+            else:
+                x_tooltip = f'@{x}'
+                formatter = {}
             tools[tools.index('hover')] = HoverTool(
-                formatters={f'@{x}': 'datetime'},
+                formatters=formatter,
                 tooltips=[
-                    (x, f'@{x}{{%F}}'),
+                    (x, x_tooltip),
                     ('Open', f'@{o}'),
                     ('High', f'@{h}'),
                     ('Low', f'@{l}'),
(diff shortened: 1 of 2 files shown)


hvplot/converter.py
```diff
@@ -2631,14 +2631,15 @@ def ohlc(self, x=None, y=None, data=None):
         tools = seg_cur_opts.pop('tools', [])
         if 'hover' in tools:
             tools[tools.index('hover')] = HoverTool(
+                formatters={f'@{x}': 'datetime'},
```
hoxbro, line 2634: You still need support the case where x is not datetime.

Azaya89: Done

Azaya89:
Ah, @%#$!

hvplot/tests/plotting/testohlc.py
```diff
@@ -32,3 +32,39 @@ def test_ohlc_hover_cols_all():
...
+def test_ohlc_date_tooltip_format():
+    plot = df.hvplot.ohlc(y=ohlc_cols)
+    segments = plot.Segments.I
+    hover_tool = segments.opts.get('plot').kwargs['tools'][0]
+    tooltips = hover_tool.tooltips
+    x_label, x_tooltip = tooltips[0]
+    assert '{%F}' in x_tooltip
+    formatter_key = '@' + x_label
+    formatter = hover_tool.formatters
+    assert formatter.get(formatter_key) == 'datetime'
```
maximlt, line 46:
```suggestion
    assert formatter[formatter_key] == 'datetime'
```

hvplot/tests/plotting/testohlc.py
```diff
@@ -32,3 +32,39 @@ def test_ohlc_hover_cols_all():
     tooltips = segments.opts.get('plot').kwargs['tools'][0].tooltips
     assert len(tooltips) == len(df.columns) + 1
     assert tooltips[-1] == ('Volume', '@Volume')
+
+
+def test_ohlc_date_tooltip_format():
```
maximlt, line 37: Another test could be added with a DataFrame whose date column isn't the index.

hvplot/converter.py
```diff
@@ -2630,15 +2630,23 @@ def ohlc(self, x=None, y=None, data=None):
         seg_cur_opts, seg_compat_opts = self._get_compat_opts('Segments')
         tools = seg_cur_opts.pop('tools', [])
         if 'hover' in tools:
+            x_data = data[x] if x in data.columns else data.index
+            if pd.api.types.is_datetime64_any_dtype(x_data):
+                x_tooltip = f'@{x}{{%F}}'
```
maximlt, line 2635: Two comments on `%F`:

- I can never remember what these sort of format codes mean :) I looked up Bokeh's documentation and think I found the place where it's documented as part of the [`DateFormatter`](https://docs.bokeh.org/en/latest/docs/reference/models/widgets/tables.html#bokeh.models.DateFormatter.format). `%F` is a *strftime* format code *Equivalent to %Y-%m-%d (the ISO 8601 date format)*. Could you add a little comment like `%F: strftime code for %Y-%m-%d`?
- Are OHLC plots limited to daily time series? Asked ChatGPT that says it's the main use case but hourly also happens ("mainly day traders and algorithmic traders focusing on short-term price movements"). What do you think about using instead `%F %T`, equivalent to `%Y-%m-%d %H:%M:%S`?

Azaya89: > What do you think about using instead %F %T, equivalent to %Y-%m-%d %H:%M:%S?

That makes sense. I'll add it.

Azaya89: Done.

image

maximlt (approved):

Issue #1442, opened 2024-10-15
Allow setting xaxis/yaxis as string


ahuang11:
Like xaxis="top" "bottom" etc
https://holoviews.org/user_guide/Customizing_Plots.html#axis-positions

maximlt:
Would it also be OK for you to use an enum like `hvplot.AXIS_LOC.Top` (naming can be improved)?

ahuang11:
Yeah I'm fine with that, as long it supports both (i.e. "top" and hvplot.AXIS_LOC.Top)

maximlt:
@ahuang11 I think this was just a bug, MR to fix it https://github.com/holoviz/hvplot/pull/1497.

As for the enums, I'm not sure why I suggested it exactly now, I was probably using a library that offered options this way and thought it was a great user experience. I agree that should be implemented in HoloViews first.

PR #1497, opened 2025-02-13
Fix setting xaxis/yaxis

maximlt:
Fixes https://github.com/holoviz/hvplot/issues/1442

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -793,12 +793,7 @@ class HoloViewsConverter:
         for axis_name in ['xaxis', 'yaxis']:
             if axis_name in kwds:
                 axis = kwds.pop(axis_name)
-                if not axis:
-                    plot_opts[axis_name] = None
-                elif not axis:
-                    plot_opts[axis_name] = axis
-                elif axis_name in plot_opts:
-                    plot_opts.pop(axis_name, None)
+                plot_opts[axis_name] = axis or None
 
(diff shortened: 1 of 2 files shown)


ahuang11 (approved):

ahuang11:
Not sure if the docs are updated yet, but there's also `bare`


xaxis/yaxis: str or None
    Whether to show the x/y-axis and whether to place it at the
    'top'/'bottom' and 'left'/'right' respectively.

maximlt:
Yes, you're right, thanks for spotting that! Let's fix it in another PR. The docstring is not missing more options:

```
    xaxis = param.Selector(default='bottom',
                                 objects=['top', 'bottom', 'bare', 'top-bare',
                                          'bottom-bare', None, True, False], doc="""
        Whether and where to display the xaxis.
        The "bare" options allow suppressing all axis labels, including ticks and xlabel.
        Valid options are 'top', 'bottom', 'bare', 'top-bare' and 'bottom-bare'.""")

    yaxis = param.Selector(default='left',
                                      objects=['left', 'right', 'bare', 'left-bare',
                                               'right-bare', None, True, False], doc="""
        Whether and where to display the yaxis.
        The "bare" options allow suppressing all axis labels, including ticks and ylabel.
        Valid options are 'left', 'right', 'bare', 'left-bare' and 'right-bare'.""")
```

PR #1514, opened 2025-02-24
Re-categorize HoloViewsConverter docstring and add missing keywords in the correct special lists

Azaya89:
fixes #1513 
fixes #1506

This PR:
- [x] re-organises the `HoloViewsConverter` class docstrings into more meaningful categories as follows:

1. **Data Options**: Options for organizing, grouping, and transforming the dataset before visualization, including labels, sorting, and indexing.
2. **Geographic Options**: Configuration options for geographic plots, including map projections, tile overlays, and geographic features like coastlines and borders.
3. **Size and Layout options**: Customization options for plot sizes and overall layout, including responsive mode.
4. **Axis Options**: Customization options for axis appearance and behavior, including labels, limits, tick formatting, and axis scaling.
5. **Grid and legend options**: Options for setting the grid or legend of plots as well as colorbar options.
6. **Interactivity options**: Bokeh specific options for adding hover tools as well as other interactive tools like tap tool and box select tool.
7. **Style Options**: Visual styling options to adjust colors, fonts, plot dimensions, and other aesthetic elements of the plot.
8. **Resampling Options**: Performance related options for handling large datasets, including downsampling, rasterization, and dynamic plot updates.
9. **Streaming Options**: Configurations for handling live data streams, including buffering, frame management, and support for large datasets.

- [x] Updates the the special lists to include keywords in the docstring that are not in the special lists.

PR #1516, opened 2025-02-28
Update customization notebook to match latest docstring

Azaya89:
This PR updates the [customization notebook](https://hvplot.holoviz.org/user_guide/Customization.html) in the user guide section to match the current state of the docstring following #1514 

This is a temporary work-around until the complete re-organization of the user guide section.


maximlt (commented):
Why didn't you include all the new groups you created?

Azaya89:
> Why didn't you include all the new groups you created?

The original notebook only had three of the categories exposed(Generic, Datashading, and Geo), so I just replaced the two that are different with the latest ones.

There is already some instructions in the beginning of the notebook on how to view the full API and I imagine that adding all of them will only make the notebook much longer and 'uglier'.

maximlt:
The problem is that by doing so you have removed many documented options. `width` for example is no longer documented  while it's a pretty important keyword. I don't think I can merge a PR that removes documentation 🙃

Azaya89:
OK. Fair point...

maximlt (approved):

PR #1517, opened 2025-02-28
Document the `stacked` docstring directly in the relevant methods

### LINKED ISSUE #1515: Document the `stacked` keyword directly in the relevant docstrings.
"In fact, if `stacked` only applies to 3 types of plot, maybe it's best to document it in their docstring directly instead of globally?"

_Originally posted by @maximlt in https://github.com/holoviz/hvplot/pull/1514#discussion_r1969384149_

Azaya89:
fixes #1515

### DIFF
--- a/hvplot/converter.py
+++ b/hvplot/converter.py
@@ -164,8 +164,6 @@ class HoloViewsConverter:
         Column name to use for splitting the plot into separate subplots by columns.
     sort_date (default=True): bool
         Whether to sort the x-axis by date before plotting
-    stacked (default=False): boolean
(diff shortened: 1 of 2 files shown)


hvplot/converter.py
```diff
@@ -452,7 +448,6 @@ class HoloViewsConverter:
         'row',
         'col',
         'sort_date',
-        'stacked',
```
maximlt, line 455: Should it be added to `_kind_options`?

Azaya89: Oh yeah, makes sense.

PR #1520, opened 2025-03-04
Refactor and improve patching of the plotting methods docstring and signature

maximlt:
- Patching only done on import when in IPython or when the env var `HVPLOT_PATCH_PLOT_DOCSTRING_SIGNATURE` is non falsy
- Move the utility functions from the top-level `__init__.py` to the `util.py` module
- Ensure the signature and docstring contain the right options, moving the backend styling options last

### DIFF
--- a/hvplot/__init__.py
+++ b/hvplot/__init__.py
@@ -58,24 +58,22 @@ To ask the community go to https://discourse.holoviz.org/.
 To report issues go to https://github.com/holoviz/holoviews.
 """
 
-import inspect
 import os
 import sys
-import textwrap
 
 import panel as _pn
 import holoviews as _hv
 
-from holoviews import Store, render  # noqa
+from holoviews import render  # noqa
 
 
-from .converter import HoloViewsConverter
+from .converter import HoloViewsConverter  # noqa
 from .interactive import Interactive
 from .ui import explorer  # noqa
-from .util import _in_ipython
-from .utilities import hvplot_extension, output, save, show  # noqa
+from .util import _PatchHvplotDocstrings, _in_ipython
+from .utilities import help, hvplot_extension, output, save, show  # noqa
 from .plotting import (
-    hvPlot,
+    hvPlot,  # noqa
     hvPlotTabular,  # noqa
     andrews_curves,  # noqa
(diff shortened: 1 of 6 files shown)


hvplot/__init__.py
```diff
@@ -58,23 +58,22 @@
...
+from holoviews import render  # noqa
 
 
-from .converter import HoloViewsConverter
+from .converter import HoloViewsConverter  # noqa
 from .interactive import Interactive
 from .ui import explorer  # noqa
-from .utilities import hvplot_extension, output, save, show  # noqa
+from .util import _PatchHvplotDocstrings, _in_ipython
+from .utilities import help, hvplot_extension, output, save, show  # noqa
```
hoxbro, line 74: It's a little weird that there are two files with util/utilities, but this was not added to this PR.

hvplot/util.py
```diff
@@ -741,6 +753,10 @@ def import_geoviews():
     return geoviews
 
 
+def _is_installed(package: str) -> bool:
```
hoxbro, line 756: Maybe cache this result?

hoxbro (approved):
I'm assuming most of the changes are "just" moved code, and no significant logic has been changed in them.

hvplot/__init__.py
maximlt: Yes it's weird :/

maximlt: That's the sort of thing we should clean up for hvPlot 1.0.

hvplot/util.py
maximlt: Yes thanks good catch,  I didn't know `find_spec` was looking at modules. I removed `_is_installed` entirely as I think it was a bit misleading and isn't used much anyway.

maximlt: > Maybe cache this result?

I thought about it before, and then thought people do special things in their environment while working directly from a notebook (install packages, autoreloading modules, etc.). The function that patches plotting signatures and docstrings doesn't only run on import but also when switching HoloViews backend. So I'd rather not cache for now, and keep that for future improvements to improve the import speed of hvPlot, which I'm sure we'll look at at some point.

PR #1523, opened 2025-03-05
Complete the signature of plot methods with all their kind options

### LINKED ISSUE #1479: Complete the signature of plot methods with all their kind options
The converter class has a mapping from plot method name to options `_kind_options`.

```python
class HoloViewsConverter:
    ...
    # Options specific to a particular plot type
    _kind_options = {
        ...
        'hexbin': ['C', 'reduce_function', 'gridsize', 'logz', 'min_count'],
        ...
    }
```

These options do not all appear in the signature of the plot method exposed to users, for instance, `min_count` for `.hexbin()`.

```python
class hvPlotTabular(hvPlotBase):
    ...
    def hexbin(self, x=None, y=None, C=None, colorbar=True, **kwds):
        ...
```

I'm not sure why some options aren't exposed in the signature, if there's one, I haven't found it yet.

Otherwise, I'd suggest adding:

- [ ] making sure the options *already defined in the signature* are correctly documented in the docstring
- [ ] add all the options not defined in the signature but in `_kind_options` to the signature.

In order to complete the second item, the default value of these options will have to be found and exposed. I assume they map to some HoloViews Parameter.

Azaya89:
fixes #1479 

this PR:
- [ ] makes sure the options already defined in the plot signature are correctly documented in the docstring
- [ ] adds all the options not defined in the plot signature but in [`_kind_options`](https://github.com/holoviz/hvplot/blob/main/hvplot/converter.py#L596-L620) to the signature.


Azaya89:
The column keyword in the [dataset method](https://github.com/holoviz/hvplot/blob/main/hvplot/plotting/core.py#L1576) was not documented yet 'cos I can't figure out what it does:

```python
import hvplot.pandas
import pandas as pd

data = pd.DataFrame({"x": ['a', 'b', 'c'], "y": [1, 2, 3]})

ds1 = data.hvplot.dataset()
ds2 = data.hvplot.dataset(columns='x')

# Check the underlying DataFrame columns
print("Columns in ds1:", ds1.data.columns)
print("Columns in ds2:", ds2.data.columns)
print("kdims in ds1:", ds1.kdims)
print("kdims in ds2:", ds2.kdims)
print("vdims in ds1:", ds1.vdims)
print("vdims in ds2:", ds2.vdims)
```

### output
```
Columns in ds1: Index(['x', 'y'], dtype='object')
Columns in ds2: Index(['x', 'y'], dtype='object')
kdims in ds1: [Dimension('x'), Dimension('y')]
kdims in ds2: [Dimension('x'), Dimension('y')]
vdims in ds1: []
vdims in ds2: []
```

@maximlt

maximlt:
Something else to note is that these methods (e.g. `hvPlotTabular.bar`) are only called in these cases:

- `import hvplot.pandas; df.hvplot.bar(...)`
- `from hvplot import hvPlot(Tabular); hvPlot(Tabular)(df).bar(...)`

They're not called in these other cases:

- `from hvplot import hvPlot(Tabular); hvPlot(Tabular)(kind='bar', ...)`
- `import hvplot.pandas; df.hvplot(kind='bar', ...)`
- `pd.options.ploting.backend = 'hvplot'; df.plot.bar(...)`
- `pd.options.ploting.backend = 'hvplot'; df.plot(kind='bar', ...)`

By adding new keywords to their signature with their defaults, we enforce the defaults when these methods are called but not in all the other cases, where the defaults are either set in the converter class (e.g. `levels = self.kwds.get('levels', 5)`) or inherited from whatever default HoloViews has set. So I think that separately, we need to find a way to make sure the defaults are always the same, however a plot is built. Maybe by expanding `HoloViewsConverter._kind_options` from a dict mapping a kind to a list to a dict mapping a kind to a dict whose values are the defaults:

```
    _kind_options = {
        'area': {'y2': <default>, 'stacked': <default>},
        ...
}
```

If we do that, we should also add a test that ensures the defaults defined in the signature and the converter are the same, so they don't drift over time.

Azaya89:
> Yep I'm also unsure. Do you know if this line is ever executed part of the test suite?
> 
> ```
>             return redim_(Dataset(data, self.kwds.get('columns')), **self._redim)
> ```

No it's not
Screenshot 2025-03-07 at 9 56 16 AM

hoxbro:
Try to replace it with `raise ValueError` and run  `pixi run test-example`. Examples don't use test coverage.

maximlt:
@Azaya89 I think you have made more changes locally; can you push them please?

hvplot/plotting/core.py
```diff
@@ -799,7 +799,7 @@ def ohlc(self, x=None, y=None, **kwds):
         """
         return self(kind='ohlc', x=x, y=y, **kwds)
 
-    def heatmap(self, x=None, y=None, C=None, colorbar=True, **kwds):
+    def heatmap(self, x=None, y=None, C=None, colorbar=True, logz=False, **kwds):
```
maximlt, line 802: Is there a reason why `reduce_function` hasn't been added to the signature?

hvplot/plotting/core.py
```diff
@@ -822,6 +822,8 @@ def heatmap(self, x=None, y=None, C=None, colorbar=True, **kwds):
             Whether to apply log scaling to the z-axis. Default is False.
         reduce_function : function, optional
             Function to compute statistics for heatmap, for example `np.mean`.
+            If omitted, no aggregation is applied and duplicate values are dropped.
+            Note: Explicitly setting this parameter to `None` is not allowed and will raise an error.
```
maximlt, line 826: Hmm this sounds like a bug :) Or put differently, why should hvPlot raise an error if this option is set to `None`? But I think I understand now why you didn't add it to the signature. Is it because it would raise this error?

Azaya89: > Hmm this sounds like a bug :) Or put differently, why should hvPlot raise an error if this option is set to `None`? But I think I understand now why you didn't add it to the signature. Is it because it would raise this error?

Yes, although I thought it was by design and not a bug.

maximlt: I think it's more about how hvPlot is implemented rather than an API design choice in this case. Options that don't have an explicit default (e.g. set in the signature, obtained with `self.kwds.get(<option>, DEFAULT)`) aren't necessarily treated consistently.

```python
        if 'reduce_function' in self.kwds:
            hmap = hmap.aggregate(function=self.kwds['reduce_function'])
```

`reduce_function` appears to be enabled when set to whatever value. It is interesting to get an error when passing down `function=None` as the signature of `Dataset.aggregate` seems to allow that (might be a bug in HoloViews).

```python
    def aggregate(self, dimensions=None, function=None, spreadfn=None, **kwargs):
```

Ignoring this error, setting `reduce_function=None` in hvPlot should likely be a no-op.

```python
        if self.kwds.get('reduce_function', None) is not None:
            hmap = hmap.aggregate(function=self.kwds['reduce_function'])
```

Let's not do that in this PR, we can separately work in improving how hvPlot deals with all these options and their default values.

maximlt (approved):

PR #1526, opened 2025-03-14
Automatically document the generic plot options

maximlt:
This PR adds a new `Reference` top-level section (folder is called `ref` as `reference` is already taken by the Reference Gallery) that for now includes this:

- Plot Options
  - Data options

The Plot Options page includes a table per options group (e.g. `Data Options`, `Resampling Options`). These tables are automatically built from the class docstring of the `HoloViewsConverter` object. This is done via a custom Sphinx extension that leverages `numpydoc` to parse the docstring. To get the table to format the content nicely, I had to touch the docstrings in a few places. Ultimately, we should enable some sort of docstring validation to make sure it is appropriately formatted.

The Data Options page is mostly a placeholder at the moment (for future PRs). While the Plot Options page only contains tables, this page should contain the table but also subsections that give more information about the options and most importantly include plots demonstrating their effects. This is why this page is a Notebook.

When this PR is merged, we should quickly be able to remove the `Customization` user guide, redirecting to the `Plot Options` page.

EDIT:
image

### DIFF
--- a/doc/_ext/plotting_options_table.py
+++ b/doc/_ext/plotting_options_table.py
@@ -123,5 +123,4 @@ def setup(app):
 if __name__ == '__main__':
     group = ' '.join(sys.argv[1:])
     opts = get_group_parameters(group)
-    breakpoint()
     print(opts)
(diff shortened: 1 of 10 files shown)


maximlt:
The dev site was built without any problem. I'm going to go ahead and merge it as I have other pieces of work that depend on it.

PR #1533, opened 2025-03-21
docs: Add versioning to documentation

hoxbro:
https://hvplot-test.holoviz.org

TODO:

- [x] correct s3 region?
- [x] redirect to /en/docs/latest/ for versioning (and will also redirect old pages)
- [x] https://hvplot-test.holoviz.org/en/docs/latest/ should not download a file
- [x] choose version switcher url
- [x] Merge Github action and update link https://github.com/holoviz/hvplot/pull/1533#discussion_r2035682618

PR #1540, opened 2025-04-09
Add version as input for docs versioning

hoxbro:
This is to test the Github interface with this flag for https://github.com/holoviz/hvplot/pull/1533. It can always be reverted.


hoxbro:
It was not needed. The problem was indention, I will revert this...

PR #1543, opened 2025-04-10
add hvsampledata to examples

hoxbro:
Currently, pip tests are failing, which has caused some problems with a failing downstream test for a datashader release.

I have changed the pip test to only run on full, as downstream testing should not have this problem.


pyproject.toml
```diff
@@ -137,6 +137,7 @@ examples = [
     "xyzservices >=2022.9.0",
     "geodatasets >=2023.12.0",
     "bokeh_sampledata; python_version >= '3.10'",
+    "hvsampledata",
```
maximlt, line 140:
```suggestion
    "hvsampledata >=0.1.1",
```

PR #1556, opened 2025-04-30
Build the docs in parallel in the pip environment

maximlt:


maximlt:
Docs build succeeded with conda and pip.

Issue #1215, opened 2023-12-13
xformatter does not accept string


MarcSkovMadsen:
I've been working on a chat interface for hvPlot. Its already available at Panel-Chat-Examples.

According to https://hvplot.holoviz.org/user_guide/Customization.html the `xformatter` argument should support a `str` argument. But when I try the plot does not show and I see errors raised in the browser console.

```python
import pandas as pd
import hvplot.pandas
import bokeh.sampledata

bokeh.sampledata.download()
from bokeh.sampledata.stocks import MSFT

data = pd.DataFrame(MSFT)[60:120]
data["date"] = pd.to_datetime(data["date"])
data = data.sort_values(by="date", ascending=False).head(100)
plot = data.hvplot(
    x='date',
    y='close',
    kind='line',
    xformatter='%d-%Y',
    xlabel='Date',
    ylabel='Closing Price',
    title='Closing Price Over Time',
    responsive=True,
)
import panel as pn
pn.panel(plot).servable()
```

![image](https://github.com/holoviz/hvplot/assets/42288570/5223984b-f376-4f14-affc-e2ecab6363cc)

If I don't use the `xformatter`, then the plot shows.

```bash
bokeh==3.3.1 holoviews==1.18.1 hvplot==0.9.0 panel==1.3.4
```

Hoxbro:
This is because the `xformatter` string is not a printf formatted string but a datetime one, this is also what is mentioned in the docstring. 

I can be recreated in HoloViews with `hv.Curve(data, "date", "close").opts(xformatter="%d-%Y")` . The [`wrap_formatter`](https://github.com/holoviz/holoviews/blob/e23c3e97bef29d77e9bb6751caf72a1928e60d5f/holoviews/plotting/bokeh/util.py#L1123)  seems to be the function to change if you want to add more logic around datetime formatted.

maximlt:
[...]

PR #1587, opened 2025-05-19
Document Interactivity options

### LINKED ISSUE #1586: Document the interactivity options

Azaya89:
closes #1586 

This PR adds the Interactivity Options to the reference docs.


doc/ref/plotting_options/interactivity.ipynb
```diff
@@ -0,0 +1,286 @@
...
+    "Hover on the points in each of the examples to see the displayed hover information."
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3e2e2707-5c15-4ae5-bc4d-dc3f14bee454",
+   "metadata": {},
+   "source": [
+    "(options-hover_formatters)=\n",
+    "## `hover_formatters`\n",
```
Azaya89, line 138: The example in this section is not working as expected. In fact, I'm not entirely sure of the accuracy of what is written in this `hover_formatters` section.

maximlt: Can you please open an issue if there's a bug?

doc/ref/plotting_options/interactivity.ipynb
```diff
@@ -0,0 +1,297 @@
...
+    "plot1 + plot2"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "0daabb1e-0ea6-4d81-8f4c-1213d271ee84",
+   "metadata": {},
+   "source": [
+    "::: {seealso}\n",
+    "See [`bokeh formatters`](https://docs.bokeh.org/en/latest/docs/reference/models/formatters.html) for more information about the formatter types.\n",
```
maximlt, line 179: Nice to reference to Bokeh's API. If it exists, it'd be nice to link first to a more user-friendly page.

doc/ref/plotting_options/interactivity.ipynb
```diff
@@ -0,0 +1,297 @@
...
+   "id": "3e2e2707-5c15-4ae5-bc4d-dc3f14bee454",
+   "metadata": {},
+   "source": [
+    "(options-hover_formatters)=\n",
+    "## `hover_formatters`\n",
+    "\n",
+    "A dictionary that controls how values appear in hover tooltips. Keys are column names or dimensions. Values must be one of:\n",
+    "- 'numeral' – for abbreviated numbers (e.g. 1k, 3.2M)\n",
+    "- 'datetime' – for date/time values\n",
+    "- 'printf' – for precision control like '%.2f'\n",
```
maximlt, line 134: Is there a way to tell a bit more about what `printf` is? I remember I digged a bit into that but that had to do with axis formatters so I'm not entirely sure it's the same thing. https://github.com/holoviz/hvplot/issues/1215#issuecomment-1854025766

doc/ref/plotting_options/interactivity.ipynb
```diff
@@ -0,0 +1,297 @@
...
+    "- CustomJSHover – for custom JavaScript formatting logic\n"
```
maximlt, line 135: Unfortunately, I'm going to ask you if you've tried that 🙃 If so, maybe add an example?

doc/ref/plotting_options/interactivity.ipynb
```diff
@@ -0,0 +1,297 @@
...
+   "source": [
+    "import hvplot.pandas  # noqa\n",
+    "import hvsampledata\n",
+    "\n",
+    "df = hvsampledata.penguins(\"pandas\")\n",
+    "\n",
+    "df.hvplot.scatter(\n",
+    "    x='bill_length_mm',\n",
+    "    y='bill_depth_mm',\n",
+    "    hover_cols=['species', 'body_mass_g'],\n",
```
maximlt, line 278: Does this example need to have the `hover_xxx` arguments? I found it a little confusing.

doc/ref/plotting_options/interactivity.ipynb
```diff
@@ -0,0 +1,297 @@
...
+   "cell_type": "markdown",
+   "id": "3e2e2707-5c15-4ae5-bc4d-dc3f14bee454",
+   "metadata": {},
+   "source": [
+    "(options-hover_formatters)=\n",
+    "## `hover_formatters`\n",
+    "\n",
+    "A dictionary that controls how values appear in hover tooltips. Keys are column names or dimensions. Values must be one of:\n",
+    "- 'numeral' – for abbreviated numbers (e.g. 1k, 3.2M)\n",
+    "- 'datetime' – for date/time values\n",
```
maximlt, line 133: A datetime example would be useful, I think users frequently struggle with that (pretty sure there are issues and discourse posts about that).

maximlt (changes requested):
Nice! I left a few comments. Additionally:

- Can you add at the beginning of the page an indication that all these options apply to the Bokeh backend only? (well some may apply to Plotly but we've agreed we ignore it for now)
- Can you add more links to HoloViews and Bokeh user guides?

doc/ref/plotting_options/interactivity.ipynb
Azaya89: OK. I'll change the example to `stocks`.

doc/ref/plotting_options/interactivity.ipynb
Azaya89: It was a bug 😄 

I've opened an issue [here](https://github.com/holoviz/holoviews/issues/6589)

doc/ref/plotting_options/interactivity.ipynb
Azaya89: The bokeh reference guide has a good explanation I think https://docs.bokeh.org/en/latest/docs/reference/models/formatters.html#bokeh.models.PrintfTickFormatter

maximlt: It is a bit confusing this is a tick formatter (`PrintfTickFormatter`). Ah but I see in https://docs.bokeh.org/en/latest/docs/user_guide/interaction/tools.html#formatting-tooltip-fields that they also reference that page, so I guess they re-use this formatter for the tooltip.

maximlt:
@Azaya89 I have noticed that hvPlot accepts a few widget options that I think were never properly documented: 'widgets', 'widget_location', 'widget_layout', 'widget_type'. There's a catch though, when one of these options is set, the object returned by `.hvplot()` is no longer a HoloViews object but a Panel object, which confuses a lot of users. I would love to be able to find a way to fix that in hvPlot (see https://github.com/holoviz/holoviews/issues/6590). Until then, my feeling is that we should at least document `widget_location` as I think it's the most useful one (it's set in other parts of the docs). What do you think?

doc/ref/plotting_options/interactivity.ipynb
maximlt: Can you ask Andrew if he has an example where `hover_formatters` is useful?

doc/ref/plotting_options/interactivity.ipynb
```diff
@@ -230,7 +230,11 @@
     "(options-widget_location)=\n",
     "## `widget_location`\n",
     "\n",
-    "Control where widgets appear. Valid values are 'right' (default), 'left', 'top', or 'bottom'."
+    "Control where widgets appear. Valid values are `'right'` (default), `'left'`, `'bottom'`, `'right'`, `'top'`, `'top_left'`, `'top_right'`, `'bottom_left'`, `'bottom_right'`, `'left_top'`, `'left_bottom'`, `'right_top'`, `'right_bottom'`.\n",
```
maximlt, line 266: Found out that there are more options that the ones originally documented (maybe the source you used is incorrect?). Often a good way to find the options allowed for a Parameter is to simply run the code with a wrong value (`widget_location='foo'`), the error message often includes all the valid options.

doc/ref/plotting_options/interactivity.ipynb
```diff
@@ -30,11 +30,11 @@
     "(options-hover)=\n",
     "## `hover`\n",
     "\n",
-    "Enables or disables hover tooltips on the plot.\n",
+    "Enables or disables hover tooltips on the plot, also accepts `'hline'` and `'vline'` to change the hit-testing mode.\n",
```
maximlt, line 33: I remembered `hover` accepts these two values too and that they were pretty much undocumented.

Azaya89:
Thanks!

PR #1600, opened 2025-06-12
Fix handling CRS translation

ahuang11:
Motivated by https://discourse.holoviz.org/t/conus-albers-crs-works-in-hvplot-pandas-but-not-in-hvplot-xarray/8808/10

Main issue was we didn't translate `aea` into `AlbersEqualArea`, so I added a lot more to the mapping.

```python
import xarray as xr
import hvplot.pandas
import hvplot.xarray
import fsspec
import cartopy.crs as ccrs


fs = fsspec.filesystem('s3', anon=True, endpoint_url='https://usgs.osn.mghpcc.org')
ds = xr.open_dataset(fs.open('esip/rsignell/testing/1646955_2487285.tiff'), engine='rasterio').squeeze('band', drop=True)
conus_albers = ccrs.AlbersEqualArea(
    central_longitude=-96.0,
    central_latitude=23.0,
    standard_parallels=(29.5, 45.5)
)
ds['band_data'].hvplot(crs=conus_albers, rasterize=True)
```

### DIFF
--- a/hvplot/tests/testgeo.py
+++ b/hvplot/tests/testgeo.py
@@ -511,7 +511,6 @@ class TestGeoUtil(TestCase):
         Test that various PROJ.4 projection strings are correctly mapped to their
         expected Cartopy CRS classes by `proj_to_cartopy`.
         """
-        """Test various newly added projections"""
         from ..util import proj_to_cartopy
 
         test_cases = [


hvplot/tests/testgeo.py
```diff
@@ -480,3 +480,75 @@ def test_proj_to_cartopy_wkt_string(self):
...
+        from ..util import proj_to_cartopy
+
+        albers_wkt = 'PROJCS["Projection = Albers Conical Equal Area",GEOGCS["WGS 84",DATUM["WGS_1984",SPHEROID["WGS 84",6378137,298.257223563,AUTHORITY["EPSG","7030"]],AUTHORITY["EPSG","6326"]],PRIMEM["Greenwich",0],UNIT["degree",0.0174532925199433,AUTHORITY["EPSG","9122"]],AUTHORITY["EPSG","4326"]],PROJECTION["Albers_Conic_Equal_Area"],PARAMETER["latitude_of_center",23],PARAMETER["longitude_of_center",-96],PARAMETER["standard_parallel_1",29.5],PARAMETER["standard_parallel_2",45.5],PARAMETER["false_easting",0],PARAMETER["false_northing",0],UNIT["metre",1,AUTHORITY["EPSG","9001"]],AXIS["Easting",EAST],AXIS["Northing",NORTH]]'
+
+        crs = proj_to_cartopy(albers_wkt)
+
+        assert isinstance(crs, self.ccrs.AlbersEqualArea)
+        # Check that the projection parameters are correctly parsed
+        # Note: proj4_params might be available depending on cartopy version
+        if hasattr(crs, 'proj4_params'):
```
hoxbro: Let's check based on the cartopy version, so we don't silently let it be skipped if there's a rename.

hvplot/tests/testgeo.py
```diff
@@ -480,3 +480,75 @@ def test_proj_to_cartopy_wkt_string(self):
...
+        from ..util import proj_to_cartopy
+
+        proj4_string = '+proj=aea +lat_0=23 +lon_0=-96 +lat_1=29.5 +lat_2=45.5 +x_0=0 +y_0=0 +datum=WGS84 +units=m +no_defs'
+
+        crs = proj_to_cartopy(proj4_string)
+
+        assert isinstance(crs, self.ccrs.AlbersEqualArea)
+
+    def test_proj_to_cartopy_additional_projections(self):
+        """Test various newly added projections"""
```
hoxbro: This is a bad docstring.

hvplot/util.py
```diff
@@ -256,13 +270,37 @@ def proj_to_cartopy(proj):
...
+        kw_proj.pop('latitude_true_scale', None)
     else:
         kw_proj.pop('latitude_true_scale', None)
 
     try:
         return cl(globe=globe, **kw_proj)
     except TypeError:
-        del kw_proj['approx']
+        if 'approx' in kw_proj:
+            del kw_proj['approx']
```
hoxbro:
```suggestion
        kw_proj.pop('approx', None)
```

Nit

hvplot/util.py
```diff
@@ -256,13 +270,37 @@ def proj_to_cartopy(proj):
         if 'central_longitude' in kw_proj:
             kw_proj['pole_longitude'] = kw_proj['central_longitude'] - 180
             kw_proj.pop('central_longitude', None)
+    elif cl.__name__ in [
```
hoxbro: Is there a way to do these `__name__` lookups with inheritance or an attribute?

hoxbro (commented):
Thank you. Have left some comments.

Can you also apply these changes to GeoViews after this has been merged?

hvplot/util.py
```diff
@@ -143,6 +143,31 @@ def proj_to_cartopy(proj):
 
     import cartopy.crs as ccrs
 
+    PROJ_TO_CCRS = {
+        'longlat': ccrs.PlateCarree,
```
hoxbro: If any of these are renamed or removed, it will break the entire function, whereas previously it would only do so for the given case.

```suggestion
        'longlat': 'PlateCarree',
```

And then do `cl = getattr(ccrs, PROJ_TO_CCRS[v])` below

hvplot/tests/testgeo.py
ahuang11: I actually couldn't find a version for this so removed the logic.

ahuang11:
> Woohoo @ahuang11 !  This works for me!
> image
> 

Nice. Are the edges expected to be like that? Seems sketchy

rsignell:
Yes, this is expected behavior.   Although I agree it's sketchy.   Ideally it would be setting those "extrapolated" areas to NaN, but that's an issue for another day.  I imagine @philippjfr could remind was why it is this way....

ahuang11:
@philippjfr found the fix in geoviews (masked_extrapolated). He will make a PR soon.

rsignell:
Gorgeous!
image

maximlt (approved):
Thanks!!

PR #1615, opened 2025-07-17
Add violin with scatter

ahuang11:
Couldn't figure out how to do no fill alpha for matplotlib, also had to use the numpy way to jitter mpl.

image


Azaya89:
I think you can do
```python
edgecolors='black',
facecolors='white',
```
for the matplotlib plot and you get something like this:

image

PR #1631, opened 2025-07-31
Fix auto-triage bot.

Azaya89:
This PR updates the `auto_triage` workflow to check type or label in an issue before auto applying the triage label. See
https://github.com/Azaya89/hvplot/issues/9 for testing.

### DIFF
--- a/.github/workflows/auto_triage.yml
+++ b/.github/workflows/auto_triage.yml
@@ -16,7 +16,7 @@ jobs:
         run: |
           # Get current labels and issue type - if any exist, the issue has been looked at
           label_count=$(gh issue view "$NUMBER" --json labels --jq '.labels | length')
-          issue_type=$(gh issue view "$NUMBER" --json issueType --jq '.issueType')
+          issue_type=$(gh api repos/$GH_REPO/issues/$NUMBER --jq '.type')
           echo "Number of existing labels: $label_count"
           echo "Issue type: $issue_type"
 


.github/workflows/auto_triage.yml
```diff
@@ -11,7 +11,30 @@ jobs:
...
+      - name: Check if issue needs triage
+        id: check-labels
+        run: |
+          # Get current labels and issue type - if any exist, the issue has been looked at
+          label_count=$(gh issue view "$NUMBER" --json labels --jq '.labels | length')
+          issue_type=$(gh api repos/$GH_REPO/issues/$NUMBER --jq '.type')
+          echo "Number of existing labels: $label_count"
+          echo "Issue type: $issue_type"
+
+          if [ "$label_count" -eq 0 ] && [ "$issue_type" = "null" ]; then
```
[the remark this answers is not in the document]
Azaya89: The 'type' field exists. It can be verified via `gh api repos/$GH_REPO/issues/$NUMBER --jq '.type'` after setting the env variable `GH_REPO` and `NUMBER`.

Example:
```bash
export GH_REPO="holoviz/hvplot"
export NUMBER=1630

gh api repos/$GH_REPO/issues/$NUMBER --jq '.type'
```
Output:
```
{
  "color": "green",
  "created_at": "2025-04-08T16:53:13Z",
  "description": "Related to the documentation and examples",
  "id": 26328872,
  "is_enabled": true,
  "name": "Documentation",
  "node_id": "IT_kwDOAxSOD84Bkb8o",
  "updated_at": "2025-04-08T16:53:13Z"
}
```

.github/workflows/auto_triage.yml
```diff
@@ -11,7 +11,30 @@ jobs:
     permissions:
       issues: write
     steps:
-      - run: gh issue edit "$NUMBER" --add-label "$LABELS"
+      - name: Check if issue needs triage
+        id: check-labels
+        run: |
+          # Get current labels and issue type - if any exist, the issue has been looked at
+          label_count=$(gh issue view "$NUMBER" --json labels --jq '.labels | length')
```
maximlt, line 18: Don't issues have a default `TRIAGE` label when created? If so, that means the bot will always skip.

Azaya89: It is the same bot that adds the default label so with this fix, it will only add the label if there was no prior label/type attached to the issue.

maximlt: Not exactly. For us maintainers who have the ability to set a label on an issue, the issue template defaults to adding the TRIAGE label.

image

There's a way to manually remove it before creating the issue but I don't think it hurts improving the bot a bit (specially if we then copy / paste its implementation to other repos).
image

Azaya89: OK, I understand you now. It still works.

- If the automatic `TRIAGE` label is added by a maintainer, the bot will skip.
- If a maintainer removes the label but adds an issue type, the bot will also skip. That way, you won't need to manually remove the `TRIAGE` label if you've already added an issue type while creating the issue.

maximlt (approved):

PR #1658, opened 2025-08-11
Enhance docstrings of plotting methods

### LINKED ISSUE #1629: Enhance the plotting method docstrings.
The plotting method docstrings are a bit of a patchwork of different documentation bits and need some more love. I think an LLM set up correctly would do a pretty good job at improving and homogenizing them.

Azaya89:
resolves #1629


maximlt:
Before I review, can you please explain how you made these changes and what you focused on?

hvplot/plotting/core.py
```diff
@@ -319,51 +323,53 @@ def line(self, x=None, y=None, **kwds):
...
+            Color(s) for the step series. Can be:
 
-            A sequence of color strings referred to by name, RGB or RGBA code, which will be used
-            for each series recursively. For instance ['green','yellow'] each field's line will be
-            filled in green or yellow, alternatively. If there is only a single series to be
-            plotted, then only the first color from the color list will be used.
+            - A single color string (name, RGB, or RGBA code), e.g., 'red' or '#a98d19'
+            - A sequence of color strings used recursively for each series
+
+            For multiple series, colors are applied cyclically.
```
maximlt: Doesn't *used recursively* and *applied cyclically* mean the same thing?

hvplot/plotting/core.py
```diff
@@ -383,64 +388,64 @@ def step(self, x=None, y=None, where='mid', **kwds):
...
             Field name(s) to draw y-positions from. If not specified, all numerical
             fields are used.
-        marker : string, optional
-            The marker shape depends on the activated plotting backend:
+        marker : str, optional
+            The marker shape. Options depend on the plotting backend:
 
-            - Bokeh: Bokeh marker styles and a subset of Matplotlib styles, e.g.
-              ``'circle'`` (default), ``'dot'``, ``'cross'``, ``'x'``, ``'square'``
-              for Bokeh markers and ``'+'``, ``'x'``, ``'s'`` for Matplotlib-
```
maximlt, line 408: Was something wrong with the matplotlib-compatible markers?

hvplot/plotting/core.py
```diff
@@ -827,56 +851,55 @@ def bivariate(
 
     def bar(self, x=None, y=None, stacked=False, **kwds):
         """
-        A vertical bar plot
+        Create a vertical bar chart for categorical data comparison.
```
maximlt: This isn't correct. HoloViews has proper support for Bars plotted on continuous axes, see https://github.com/holoviz/holoviews/pull/6145. We also have an example in the docs. So the docstring needs to be updated and also the parameter definition for x.

Image

hvplot/plotting/core.py
```diff
@@ -888,9 +911,9 @@ def bar(self, x=None, y=None, stacked=False, **kwds):
 
     def barh(self, x=None, y=None, stacked=False, **kwds):
         """
-        A horizontal bar plot
+        Create a horizontal bar plot.
```
maximlt, line 913: I guess same as bar it's not limited to categorical data?

hvplot/plotting/core.py
```diff
@@ -1027,48 +1076,54 @@ def hist(
...
+        by : str or list of str, optional
+            Field name(s) to group data by before computing histograms.
+        bins : int, str, array-like, default 20
+            Bin specification. Can be:
+
+            - int: Number of equal-width bins
+            - str: Method for automatic bin selection ('auto', 'fd', 'scott', etc.)
+            - array-like: Explicit bin edges
+
+            See :func:`numpy.histogram_bin_edges` for automatic methods.
```
maximlt: What does "for automatic methods" mean?

hvplot/plotting/core.py
```diff
@@ -1027,48 +1076,54 @@ def hist(
...
+            Field name(s) to compute distribution from. Should contain continuous
+            data, not categorical. If not specified, all numerical fields are used.
+        by : str or list of str, optional
+            Field name(s) to group data by before computing histograms.
+        bins : int, str, array-like, default 20
+            Bin specification. Can be:
+
+            - int: Number of equal-width bins
+            - str: Method for automatic bin selection ('auto', 'fd', 'scott', etc.)
+            - array-like: Explicit bin edges
```
maximlt: Not sure it's so obvious that a list of tuples is supported.

hvplot/plotting/core.py
```diff
@@ -1274,74 +1334,82 @@ def points(self, x=None, y=None, **kwds):
...
         ----------
-        x : string
-            Field name to draw x-positions from
-        y : string
-            Field name to draw y-positions from
-        mag : string
-            Magnitude.
-        angle : string
-            Angle in radians.
+        x : str, required
```
maximlt: Why `required`?

hvplot/plotting/core.py
```diff
@@ -1735,44 +1821,48 @@ class hvPlot(hvPlotTabular):
 
     def image(self, x=None, y=None, z=None, colorbar=True, **kwds):
         """
-        Image plot
+        Create an image plot from gridded data.
 
-        You can use `image` to display for example geographic data with independent `latitude` and
-        `longitude` fields and a third dependent field.
+        You can use image plots to display for example geographic data with independent ``latitude`` and
+        ``longitude`` fields and a third dependent field.
```
maximlt: They could be confused with actual arguments, which they are not.

```suggestion
        You can use image plots to display for example geographic data with independent latitude and
        longitude fields and a third dependent field.
```

maximlt (changes requested):

hvplot/plotting/core.py
Azaya89: No. It's still listed below (lines 415) but not redundantly as before.

hvplot/plotting/core.py
Azaya89: I meant the automated bin size that can be calculated when the specified options found in https://numpy.org/devdocs/reference/generated/numpy.histogram_bin_edges.html are used (e.g, 'fd', 'doane' etc).

Maybe i'll rephrase it.

hvplot/plotting/core.py
Azaya89: From testing the reference examples, I see that they are required parameters.

hvplot/plotting/core.py
Azaya89: Actually, in this context yes.I'll rephrase.

hoxbro: I see it as recursively depends on the input and how nested it is. Cyclically is that it will get the next item in the sequence and when it ends it will start over. But maybe that is just how I see it.

Azaya89: Yes, that's the correct meanings but in the context of how color is used here, it is basically simple loop behaviour. So functionally it's really the same thing.

hvplot/plotting/core.py
maximlt: I don't see a `required` tag in the numpydoc style https://numpydoc.readthedocs.io/en/latest/format.html. A bare `x: str` should be enough to indicate `x` is required. Let me know what you think.

hvplot/plotting/core.py
maximlt: @Azaya89 I remember now I wrote it this way. I think that, when the bokeh plotting backend is active, we support the Bokeh markers AND a subset of the Maplotlib markers thanks to this mapping in holoviews https://github.com/holoviz/holoviews/blob/67201d13e226f04b614ede4131c321fed5353edb/holoviews/plotting/bokeh/styles.py#L88-L100. I don't think that mapping is documented anywhere in HoloViews, but I found out about it looking at the code, and so I documented it here but not very clearly I guess.

hvplot/plotting/core.py
```diff
@@ -1027,48 +1074,54 @@ def hist(
...
+            Field name(s) to compute distribution from. Should contain continuous
+            data, not categorical. If not specified, all numerical fields are used.
+        by : str or list of str, optional
+            Field name(s) to group data by before computing histograms.
+        bins : int, str, array-like, default 20
+            Bin specification. Can be:
+
+            - int: Number of equal-width bins
+            - str: Method for automatic bin selection ('auto', 'fd', 'scott', etc.)
+            - array-like: Explicit bin edges or a list of tuples
```
maximlt: Ugh ok I read the original docstring incorrectly somehow as "list of tuple" instead of "list OR tuple". "list of tuple" makes no sense for a histogram :D.

What we support is `int` (passed as `num_bins` to holoviews), and `np.array`, `list` and `tuple` that we pass as `bins` to holoviews (https://github.com/holoviz/holoviews/blob/67201d13e226f04b614ede4131c321fed5353edb/holoviews/operation/element.py#L847).

maximlt (commented):
Thanks for the changes. Added a few comments.

hvplot/plotting/core.py
Azaya89: Yeah, on second thoughts and looking at the docs again, you're right. It's implied unless it's optional then the `optional` tag is specified.

Azaya89:
> Thanks for the changes. Added a few comments.

I think I've addressed them all now.

maximlt (approved):

PR #1672, opened 2025-09-02
DOC: Scale networkx layout and fix graphviz init in doc build

MridulS:
I was about to point someone towards hvplot docs for networkx and saw that the notebook build is broken https://hvplot.holoviz.org/en/docs/dev/user_guide/NetworkX.html for pygraphviz (and rest of the notebook).

graphviz from conda-forge should run the post link `dot -c` but I'm not sure why it isn't doing it. (maybe pixi ignores it?)

I also scaled the layout for shell layout from networkx so it's looks prettier while comparing the two layouts.


MridulS:
> Nice to hear that! Do you have any feedback on the networkx plotting API implemented in hvPlot? I have to say I haven't used it much myself, if at all, so I can't really say how good/practical/robust it is.

Nothing major, just wanted to point someone who asked for bokeh + networkx :)

PR #1680, opened 2025-09-29
Add gridstyle

ahuang11:


hvplot/converter.py
```diff
@@ -475,8 +475,11 @@ class HoloViewsConverter:
     fontsize : number or dict or None, default=None
         Set title, label and legend text to the same fontsize. Finer control
         by using a dict: ``{'title': '15pt', 'ylabel': '5px', 'ticks': 20}``.
-    grid : bool or None, default=None
-        Whether to show a grid.
+    grid : bool, str, or None, default=None
+        Whether to show a grid. If True, shows grid on both axes.
+        If ``'x'`` or ``'y'``, shows grid only on the specified axis.
+        Suffix with ``'dashed'``, ``'dotted'``, ``'dotdash'``, or ``'dashdot'``
```
maximlt: Are all these options supported by Matplotlib too? I recently looked into what `line_dash` (bokeh) / `linestyle` (matplotlib) supported and documented it there https://hvplot.holoviz.org/en/docs/latest/ref/api/manual/hvplot.hvPlot.line.html#line-dash.

maximlt (commented):
Does it allow to style both the x and y axes at the same time? If not, how about using another approach (e.g. dict from axis to style, list of tuples)?

hvplot/converter.py
ahuang11: > Are all these options supported by Matplotlib too? I recently looked into what line_dash (bokeh) / linestyle (matplotlib) supported and documented it there

Thanks, I totally neglected matplotlib (and plotly); I'll see if there's a way to convert it; if not just the dash or dotted should be sufficient.

> Does it allow to style both the x and y axes at the same time? If not, how about using another approach (e.g. dict from axis to style, list of tuples)?

I personally think at that point, people should use `plot.opts(gridstyle={...})`, else typing the `{"line_dash": "dashed"}` would be tedious. What do you think?

maximlt: > Thanks, I totally neglected matplotlib (and plotly); I'll see if there's a way to convert it; if not just the dash or dotted should be sufficient.

You can ignore Plotly.

ahuang11: Interestingly, holoviews mpl does not support gridstyle so I guess I need to implement it in hv first...

maximlt: Ugh :/ Had a discussion recently on Discord where I said HoloViews isn't a plotting library. That's exactly what I meant :D

I just realized something. When it comes to plot styling options, hvPlot usually acts as a pass-through. One advantage of that is that the keys and their allowed values are common between hvPlot and HoloViews (so it's easier for users to go from one to the other). One disadvantage is that the keys and their values are backend-dependent. As a result, we haven't attempted to build a "unified styling API" in hvPlot, that would work for all the supported backends.

It looks like this is where we are heading in this PR with e.g. `grid="x-dashed"`. Can I have your thoughts on that?

ahuang11: > Interestingly, holoviews mpl does not support gridstyle
https://github.com/holoviz/holoviews/pull/6700

> It looks like this is where we are heading in this PR with e.g. grid="x-dashed". Can I have your thoughts on that?.

I actually just added a clause for dict support, so all of `grid='x'`, `'grid='x-dashed'`, `grid={'grid_line_color': 'red'}` works.

> hvPlot usually acts as a pass-through

To me, hvplot is not only a pass-through--(me being a PM for a second) it's a stress-free, direct plotting experience (at least for me), without having to deal with complex dict structures. Every time I have to use legend_opts or gridstyle, I have to pause and remember the valid key value pairs, which is why I like having a string like `x-dashed` or `x.dashed` or `xdashed`.

ahuang11: > As a result, we haven't attempted to build a "unified styling API" in hvPlot, that would work for all the supported backends.

Oh and I also don't fully understand the backend_compat, so I dumped directly in the function, but I imagine there could be a better place if you can help me understand the flow of converter.py, maybe in `_process_style`?

doc/ref/plotting_options/styling.ipynb
```diff
@@ -311,6 +311,54 @@
...
+    "plot2 = df.hvplot(group_label=\"Company\", grid='y-dashed', width=400, title=\"grid='y-dashed'\")\n",
+    "(plot1 + plot2).cols(1)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "59315b89",
+   "metadata": {},
+   "source": [
+    " A dictionary of grid style options may also be supplied, e.g. for bokeh, ``{'grid_line_color': 'red', 'grid_line_alpha': 0.5}``."
```
maximlt, line 344: Could you update one of the examples to make it a bit more advanced? I'm thinking that one with minor grid lines would be pretty nice.

doc/ref/plotting_options/styling.ipynb
```diff
@@ -311,6 +311,54 @@
...
+    "plot1 = df.hvplot(group_label=\"Company\", grid='x', width=400, title=\"grid='x'\")\n",
+    "plot2 = df.hvplot(group_label=\"Company\", grid='y-dashed', width=400, title=\"grid='y-dashed'\")\n",
+    "(plot1 + plot2).cols(1)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "59315b89",
+   "metadata": {},
+   "source": [
```
maximlt, line 343: Could you link to Bokeh's docs? So users have an easy way to check all the options they can set.

hvplot/converter.py
```diff
@@ -1059,7 +1064,28 @@ def __init__(
             plot_opts['logy'] = logy
 
         if grid is not None:
-            plot_opts['show_grid'] = grid
+            if isinstance(grid, str):
+                gridstyle = {}
+                axis = grid[0]
+                other_axis = 'x' if axis == 'y' else 'y'
+                if len(grid) > 1:
+                    line_dash = grid[1:].lstrip('-').lstrip('.').lstrip('_')
```
maximlt, line 1088: Why `lstrip('.').lstrip('_')`?

hvplot/converter.py
maximlt: Yeah I'm not sure where to do that to be honest. I think you're pretty much introducing a new pattern with a unified plot API for multiple backends. Probably not `_process_style` as `gridstyle` is a plot option in HoloViews. But having that in the `__init__.py` doesn't feel right either?

PR #1681, opened 2025-09-29
Automatically use y as default 'text' kwarg if only two cols

ahuang11:
If only two columns, use y (this would previously crash because of `text = [c for c in data.columns if c not in (x, y)][0]`):
```python
import hvplot.pandas
import pandas as pd

df = pd.DataFrame({
    'lat': [-34.58, -15.78, -33.45, 4.60, 10.48],
    'lon': [-58.66, -47.91, -70.66, -74.08, -66.86],
})

df.hvplot.points(x='lon', y='lat') * \
df.hvplot.labels(x='lon', y='lat', hover=False, padding=0.2)
```

image

### DIFF
--- a/hvplot/tests/testcharts.py
+++ b/hvplot/tests/testcharts.py
@@ -543,7 +543,8 @@ class TestChart1D(ComparisonTestCase):
             Dimension('Latitude'),
             Dimension('Latitude'),
         ]
-        assert list(plot.data['label']) == ['-34.58', '15.78', '-33.45']
+        assert plot.kdims == ['Longitude', 'Latitude']
+        assert plot.vdims == ['Latitude']
 
     def test_labels_by(self):
         plot = self.edge_df.hvplot.labels(


hvplot/tests/testcharts.py
```diff
@@ -535,6 +535,17 @@ def test_labels_format_float(self):
...
+    def test_labels_default_y(self):
+        edge_df = self.edge_df.copy().drop(columns=['Volume {m3}'])
+        plot = edge_df.hvplot.labels('Longitude', 'Latitude')
+        assert list(plot.dimensions()) == [
+            Dimension('Longitude'),
+            Dimension('Latitude'),
+            Dimension('Latitude'),
+        ]
+        assert plot.kdims == ['Longitude', 'Latitude']
+        assert plot.vdims == ['Latitude']
```
maximlt, line 547: Just to double check, is it ok for a holoviews element to have a dimension both in the kdims and vdims?

Azaya89: I think it can if you pass them in directly:
```python
import holoviews as hv
import pandas as pd

df = pd.DataFrame({
    'lat': [-34.58, -15.78, -33.45, 4.60, 10.48],
    'lon': [-58.66, -47.91, -70.66, -74.08, -66.86],
})

plot = hv.Points(df, ['lon', 'lat'], 'lon')
plot.dimensions()
```
`[Dimension('lon'), Dimension('lat'), Dimension('lon')]`

This is what is now achieved here via `next((c for c in data.columns if c not in (x, y)), y)`, no? @ahuang11

ahuang11: Almost! Very close:

`next((c for c in data.columns if c not in (x, y)), y)` should default to y, which is:

```python
import holoviews as hv
import pandas as pd

df = pd.DataFrame({
    'lat': [-34.58, -15.78, -33.45, 4.60, 10.48],
    'lon': [-58.66, -47.91, -70.66, -74.08, -66.86],
})

plot = hv.Points(df, ['lon', 'lat'], 'lat')  # lat not lon
plot.dimensions()
```

If df has an extra col, it'll use that:
```python
import holoviews as hv
import pandas as pd

df = pd.DataFrame({
    'lat': [-34.58, -15.78, -33.45, 4.60, 10.48],
    'lon': [-58.66, -47.91, -70.66, -74.08, -66.86],
    'col': ['a', 'b', 'c', 'd', 'e']
})

plot = hv.Points(df, ['lon', 'lat'], 'col')
plot.dimensions()
```

And yes to Maxime's question:
image

maximlt: Yes I saw the plot could be created but was wondering if it wouldn't mess up with other things in HoloViews like slicing data. Anyway, I don't think it's so important for this use case so I'll merge.

maximlt (approved):

PR #1682, opened 2025-10-08
Compatibility with the latest versions of DuckDB

maximlt:
The `pip` tests (scheduled to run weekly) started to fail with the last release of duckdb. Somehow I haven't been able to find their changelog. The changes made in this PR:
- `is_duckdb` returned False with the newest versions, as the module is not called `_duckdb` and not `duckdb`
- they deprecated (not removed yet) the `duckdb.typing` module. I assumed this was done in 1.4.0 (current is 1.4.1)

Additionally, one ibis test was failing as it uses duckdb as its internal in-memory backed, and they haven't yet released a version compatible with the latest duckdb. The issue I was facing is fixed in https://github.com/ibis-project/ibis/pull/11666 but not yet released. So the test is skipped, assuming the next release of ibis will be compatible.

### DIFF
--- a/doc/conftest.py
+++ b/doc/conftest.py
@@ -48,7 +48,7 @@ try:
 
     # 'Ibis <= 10.8.0 is incompatible with DuckDB >= 1.4')
     if parse(ibis.__version__) <= parse('10.8.0') and parse(duckdb.__version__) >= parse('1.4'):
-        collect_ignore_glob += ['doc/ref/data_libraries.ipynb']
+        collect_ignore_glob += ['ref/data_libraries.ipynb']
 except ImportError:
     pass
 


hoxbro:
> current is 1.4.1

It was done in 1.4.1. FWIW.

doc/conftest.py
```diff
@@ -42,6 +42,16 @@
         'user_guide/Integrations.ipynb',
     ]
 
+try:
+    import ibis
+    import duckdb
+
+    # 'Ibis <= 10.8.0 is incompatible with DuckDB >= 1.4')
+    if parse(ibis.__version__) <= parse('10.8.0') and parse(duckdb.__version__) >= parse('1.4'):
+        collect_ignore_glob += ['doc/ref/data_libraries.ipynb']
```
hoxbro:
```suggestion
        collect_ignore_glob += ['ref/data_libraries.ipynb']
```

PR #1700, opened 2025-12-17
Address warnings emitted by the unit tests suite

maximlt:
One code change is that we no longer try to parse a crs with `pyproj.Proj(init=crs, preserve_units=True)` as this is deprecated in pyproj.

### DIFF
--- a/hvplot/tests/testgeo.py
+++ b/hvplot/tests/testgeo.py
@@ -13,7 +13,6 @@ import pandas as pd
 import pytest
 
 from hvplot.util import proj_to_cartopy
-from packaging.version import Version
 
 pytestmark = pytest.mark.geo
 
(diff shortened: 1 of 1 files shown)


hvplot/converter.py
```diff
@@ -1625,7 +1625,11 @@ def _process_data(
                 # Broken, see https://github.com/holoviz/hvplot/issues/1364.
                 # Dask reset_index doesn't accept a level, so this would need to
                 # be adapted for Dask.
-                self.data = data.stack().reset_index(1).rename(columns={'level_1': group_label})
+                self.data = (
+                    data.stack(future_stack=True)
```
hoxbro, line 1629: I don't think this is backward compatible for all supported versions of pandas. It looks like this was added in Pandas 2.1, and we support down to 1.3.

maximlt: > we support down to 1.3

Hmm this is just a lower pin in the `pyproject.toml`. In practice, I have no clue what the real lower pin is, until testing with lower pins is enabled. But yes I'll add compatibility code

PR #1725, opened 2026-05-12
deprecate sample_data module

### LINKED ISSUE #1723: Deprecate the sample_data module
We currently have both [sample_data](https://github.com/holoviz/hvplot/blob/main/hvplot/sample_data.py) and [sampledata](https://github.com/holoviz/hvplot/blob/main/hvplot/sampledata.py) as modules within hvPlot for downloading some data. The former uses intake and iirc, we are trying to move away from using intake completely. Also, both names are very similar and can be confusing for both users and LLMs (at least I can confirm this case. See https://github.com/holoviz-dev/holoviz-skills/pull/7#issuecomment-4392518763).

I propose we deprecate `sample_data` since we already have the newer and more actively maintained `sampledata` module.

Azaya89:
Fixes #1723

## Description

This PR deprecates the `sample_data` module and recommends `sampledata` instead.

## AI Disclosure
- [x] This PR contains AI-generated content.
  - [x] I have tested all AI-generated content in my PR.
  - [x] I take responsibility for all AI-generated content in my PR.
 
Tools and Models: Copilot Claude Sonnet 4.6

## Checklist
- [x] Tests added and are passing


Azaya89:
@maximlt What do you think about updating the examples in the `User Guide` seeing that that is what is causing the tests to fail?

hvplot/sample_data.py
```diff
@@ -1,30 +1,21 @@
...
+    "future version. Use 'hvplot.sampledata' instead.",
+    FutureWarning,
+    stacklevel=_find_stack_level(),
+)
 
-# Add catalogue entries to namespace
-for _c in catalogue:
-    globals()[_c] = catalogue[_c]
+from .sampledata import *  # noqa: F401, F403, E402
+from .sampledata import __getattr__  # noqa: F401, E402
```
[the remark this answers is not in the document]
Azaya89: I already removed all such examples.

doc/user_guide/Plotting.ipynb
```diff
@@ -157,8 +147,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "flight_subset = flights[flights.carrier.isin(['OH', 'F9'])]\n",
-    "flight_subset.hvplot(x='distance', y='depdelay', by='carrier', kind='scatter', alpha=0.2, persist=True)"
+    "taxi_subset = taxi[taxi.passenger_count.isin([1, 2])].sample(frac=0.01).compute()\n",
```
maximlt: Will change to 0.001. With 0.01 we throw 100,000 data points, the plot suffers from overplotting, and that makes the HTML page heavier.

doc/index.md
```diff
@@ -263,10 +263,11 @@ align: center
...
+from hvplot.sampledata import stocks
 
-plot1 = df.hvplot(x='Year', y='Violent Crime rate', width=400)
-plot2 = df.hvplot(x='Year', y='Burglary rate', width=400)
+df = stocks('pandas')
+plot1 = df.hvplot(x='date', y='Apple', width=400)
+plot2 = df.hvplot(x='date', y='Google', width=400)
 plot1 + plot2
 ```
 ```{image} ./_static/home/layout.gif
```
maximlt: Did you generate the GIF to replace this?

scripts/download_data.py
```diff
@@ -5,30 +5,37 @@
...
-    retry(xr.tutorial.open_dataset, 'air_temperature')
-    retry(xr.tutorial.open_dataset, 'rasm', decode_times=False)
-    print('xarray data downloaded.')
+    download('xarray air_temperature', xr.tutorial.open_dataset, 'air_temperature')
+    download('xarray rasm', xr.tutorial.open_dataset, 'rasm', decode_times=False)
 except ModuleNotFoundError as e:
     print(f'ModuleNotFoundError when attempting to download xarray datasets : {e}')
+
+try:
+    from hvplot.sampledata import download as _hvs_download
```
maximlt, line 37: FYI scripts are not distributed so you don't have to bother with this kind of import.

maximlt (commented):
My main comment is that the original change to the `sample_data` module wasn't a deprecation. It was direct code removal. Removing a feature has to go through a well-defined process (see [HEP2](https://holoviz.org/about/heps/hep2.html)) with discussion, deprecation, and finally removal. I added the `sample_data` code back. In a future release, we will be able to remove the module and `datasets.yaml` (that was forgotten here) entirely.

My second comment is that I find some of the replaced plots not always meaningful. I feel like that when we use datasets based on real data (e.g. stocks, penguins), we need to make realistic plots. Otherwise, we risk losing a bit of credibility with our users. I've listed these plots below. Soon (hopefully) these user guides should go away, so I'm ok merging these changes as they are meant to be temporary. Let's keep in mind though for the future that we have to make sure we generate realistic plots.

## Plotting with Bokeh

That's not a very pretty plot.
Image

I don't think that it makes so much sense.
Image

Not the most common kind of bar plot.
Image

hexbin plots are better displayed imo with an equal aspect ratio.
Image

Can't really see much on this one.
Image

Same.
Image

## Plotting Extensions

The first plot isn't great.

Image

## Subplots

Lots of empty cells in this subplot, might look a bit off for visitors.
Image


I think there's a missing GIF for the landing page.

maximlt:
Just to clarify @Azaya89, the only change that I would like is a new GIF for the landing page. Everything else is good to be merged.

doc/index.md
Azaya89: Ooof!, oversight. Will fix now.

maximlt (approved):

PR #1731, opened 2026-06-18
Avoid duplicate parametrization of 'backend'

Azaya89:
## Description
This PR fixes the test failures caused by duplicate parametrization of 'backend'. See https://github.com/holoviz/hvplot/actions/runs/27755482801/job/82116308402

- [x] Tests added and are passing


Azaya89:
i wanted to try it with the first test to see if it would work. Looks like that was the fix. Will apply across the test file now...

maximlt (approved):
The issue is coming from a regression in pytest that is already fixed and should be released soon https://github.com/pytest-dev/pytest/issues/14591. Anyway this change LGTM.

PR #1654, opened 2025-08-05
Reorganize Widgets User guide

Azaya89:
fixes #1641

- [x] Create how-to section and notebook for widgets exploration
- [x] Move panel widgets part to reference section
- [x] Create Explanation section and notebook for `groupby` vs `by` vs `color`

PR #1741, opened 2026-07-14
Document generated widget options

### LINKED ISSUE #1108: Document widget_* kwargs
There doesn't seem to be any docs on it.

https://hvplot.holoviz.org/user_guide/Customization.html

Also mention it converts to panel obj (inspired by https://discourse.holoviz.org/t/secondary-y-axis-for-time-series/5716/2)

Kkkakania:
## Summary

- document the actual `right_top` default for generated widget placement
- add `widgets`, `widget_layout`, and `widget_type` to the public converter option reference
- add executable examples for a custom widget type, horizontal widget layout, and a multi-dimension scrubber

The new reference stays in the existing size and layout options page and does not overlap with the broader widget-guide reorganization in #1654.

Closes #1108.

## Validation

- `python -m pytest hvplot/tests/testpanel.py -q` (3 passed)
- `python -m pytest --nbval-lax -p no:python doc/ref/plotting_options/size_layout.ipynb -q` (14 passed)
- `pre-commit run --files hvplot/converter.py doc/ref/plotting_options/size_layout.ipynb` (all hooks passed)
- `jq empty doc/ref/plotting_options/size_layout.ipynb`
- `git diff --check`

AI disclosure: OpenAI Codex assisted with repository inspection, the documentation changes, and local validation under Kkkakania's direction.


maximlt:
We decided not to document these options in details as there is a chance they will be deprecated/modified. Enabling them returns a Panel object instead of a HoloViews object, which we want to fix, and in doing so the API may change. So we don't want users starting to leverage them.

Kkkakania:
Understood, thanks for the context. I agree that the reference should not encourage use of an API that may be deprecated or substantially changed, so I'm closing this PR.

